PR #93, opened 2016-07-21
Changed tradeCommits[](hash, block) to tradeCommits[](hash[], block[]…

tinybike:
…) on trades.se, and added an id parameter to commitTrade, zeroHash, checkHash


tinybike:
Revert the multiple trade commits stuff, merge the rest to master.

PR #101, opened 2017-01-31
Fixed markdown issues

AndrewBezold:
Fixed markdown issues in Readme


joeykrug:
Oh that should be 200 instead of 180

PR #106, opened 2017-04-18
Renames `reporting` to `reputation` and updates comment to reflect reality.

MicahZoltu:
This PR serves two purposes:
1. To help improve the readability of this contract.
2. To ensure that my understanding of this contract's behavior is correct.

If the readability improvement isn't desired, I would still find value in an answer to whether or not my understanding is correct that this field is a mapping of Ethereum address to reputation held by that address.


joeykrug:
Correct, merging + closing

PR #108, opened 2017-04-18
Adds refund call to a number methods.

MicahZoltu:
Anything that doesn't expect ether (which is everything other than deposit function) should refund any ether passed in.

Let me know if the whitespace cleanup bothers you.  My editor does it automatically but I can turn it off/fixup the PR if this sort of thing is undesirable in PRs.


joeykrug:
whitespace is fine

PR #114, opened 2017-04-19
Simplifies the mutex and changes naming to match convention found in other languages.

MicahZoltu:
Note: This PR is on top of #113, that should be merged first.  If #113 is rejected, then I'll update this PR to use the old function name.

This makes the mutex have two methods, `acquire` and `release`.  Calling `acquire` will either get you the mutex or it will throw and calling `release` will simply release the mutex.  This makes its usage much simpler in the common case as you only need to call `acquire`, you don't need to check the result or do a `get` first.

**NOTE**: There were a few places in the code where it was acquiring a mutex even if the caller was whitelisted.  I assumed that these were mistakes and replaced them with the pattern of "either you are whitelisted or you acquire, never both" but perhaps there is a legitimate reason that functions have the contract acquire a mutex even when whitelisted?  An example of this can be seen in the old code in `collectFees.se`

There may be value in a future improvement in creating a macro for `if (!whitelisted): MUTEX.acquire()` but I wanted to keep this PR as simple as possible.

### DIFF
--- a/src/data_api/mutex.se
+++ b/src/data_api/mutex.se
@@ -7,15 +7,14 @@ data controller
 def init():
     self.controller = 0x0
 
-def setMutex():
+def acquire():
+    if (self.mutex == 1):
+        throw()
     self.controller.checkWhitelist(msg.sender)
     self.mutex = 1
     return(1)
 
-def unsetMutex():
+def release():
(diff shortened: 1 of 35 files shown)


MicahZoltu:
Rebased and resolved merge conflicts.

Issue #111, opened 2017-04-19
Remove debug/test code from cash.se


MicahZoltu:
cash.se has quite a bit of code that shouldn't go into production.  This includes the seeding of the creator's account with a balance of 10,000,000 Cash and all of the code around allowing Augur to force transfers.

From what I can tell, the allowing of Augur to force transfers is to allow for faucets, which is a test-only feature.  This should not go into production as it increases code complexity and increases the attack surface area against the cash contract significantly, as now any Augur contract has the ability to "steal money" sitting in people's cash accounts.  For testing, Ethereum's testnet should be used and it is easy to either setup a private chain seeded with all the ETH you want, or one can mine for a day or two on testnet to get some ETH.

Augur contracts are not the right place for faucet code because of the increased risk and complexity it introduces for very little benefit.

Removing this code from the cash contract will greatly simplify the contract and remove the need to have it linked in to the whitelist system.  The contract will then be a standard token contract with a 1:1 ETH peg.

joeykrug:
Go ahead on this one

MicahZoltu:
I think this didn't auto-close because the fix hasn't been merged to default branch (master) yet.  Manually closing.

PR #122, opened 2017-04-20
Removes test code from cash contract.

MicahZoltu:
This means no more cash faucets, you'll need to use a testnet faucet instead.


MicahZoltu:
Fixes #111

PR #128, opened 2017-04-24
Miner readability improvements.

MicahZoltu:
Extracted out a function to find a file by name which allows us to get rid of a couple of loops (and therefore nesting).

Also made it so it raises an error if a contract was referenced but not actually included.

Finally removed a block that was in a `recompile == 0` check since this code is in an `if(recompile)` block meaning that check would always fail.

### DIFF
--- a/load_contracts.py
+++ b/load_contracts.py
@@ -413,37 +413,35 @@ class ContractLoader(object):
             print('Contract loading successful')
 
         if(recompile):
-            for file in serpent_files:
-                if os.path.basename(file) == controller:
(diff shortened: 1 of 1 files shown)


MicahZoltu:
Feel free to close if the changes don't interest you.  Just some minor things I did while I was reading the code to make it easier to grok.

MicahZoltu:
Recommend viewing with [?w=1](https://github.com/AugurProject/augur-core/pull/128/files?w=1) since most of the "changes" are whitespace.

PR #141, opened 2017-04-25
Refactors reporting.se.

MicahZoltu:
## NOTE
This PR _only_ includes changes to `reporting.se`.  Before merging, all callsites need to be updated to match the new function signatures and may need minor modifications to maintain the "no bad state allowed" paradigm mentioned below.  I wanted to get some initial feedback before moving forward as I don't want to burn time if these changes are not approved/agreed-on.  I recommend starting with the description of changes below before reviewing the code itself.

## Changes

The dominant change here is making it so `reporting.se` is always in a healthy state.  Previously, there were methods for manipulating the state of this contract that allowed the caller to put this contract into a bad state under the assumption that the caller would follow-up with another call that put the contract back into a good state.  Examples of this were functions like `addRep` and `subtractRep` which would all for more or less than 11,000,000 REP to be in existence.  Now any single method call on this contract will leave this contract in a reasonable and healthy state.  There is no expectation that the caller chains certain calls together to keep the system healthy.

Another change of note is that I have normalized the state in this contract.  REP now is either in `activeRep` or `dormantRep`.  Previously the state was a bit denormalized in that `reputation` contained both `dormant` and `active` rep while `dormant` contained only `dormant`.

Other changes of note:
* I have changed the nomenclature from `fxpamount` to `attorep` to make it clear to the reader that we are working with attorep (which equals `10^-18` REP).
* `reportedOnNonFinalRoundTwoEvent` renamed to `reporterLocks` for improved readability and also changed to an array of events, thus allowing a reporter to be locked by multiple events at a time.  I believe this is correct behavior, though I don't fully understand round 2 locking enough to be confident.
* Got rid of the `reporterIndex` concept, reputation is now mapped directly by reporter address.
* Added a number of comments related to bugs, attack vectors and questions.  These should all be resolved in the near future, but they can wait until a follow-up PR.  Discussion and/or more understanding/research is necessary to resolve most of them.
* Removed some duplicate code in `getReputation`
* REP that needs to be redistributed is now stored in `Reputation[branch].dormantRep[0x0000000000000000000000000000000000000000]` rather than branch ID.  This is mainly because branch ID is redundant and this is how it is stored in `init`, so keeps things consistent.
* Removes test faucet support.  We can add this back in temporarily if necessary, though it breaks the contract mentioned in the first paragraph (it puts REP into a state that it shouldn't be allowed to get into by creating REP out of thin air).

### DIFF
--- a/src/data_api/reporting.se
+++ b/src/data_api/reporting.se
@@ -27,199 +27,147 @@ extern branches: [addCurrency:[int256,int256,int256,int256]:int256, addMarketToB
 REPCONTRACT = self.controller.lookup('repContract')
 extern repContract: [allowance:[address,address]:uint256, approve:[address,uint256]:uint256, balanceOf:[address]:uint256, decimals:[]:uint256, getRidOfDustForLaunch:[]:int256, getSeeded:[]:uint256, name:[]:uint256, setSaleDistribution:[address[],uint256[]]:uint256, symbol:[]:uint256, totalSupply:[]:uint256, transfer:[address,uint256]:uint256, transferFrom:[address,address,uint256]:uint256]
 
+inset('../macros/float.sem')
 inset('../macros/refund.sem')
 
 # Storage of all data associated with reporters
-    # Reporting index is the branch
-    # Reputation index is the index of the reporter
-    # RepIDtoIndex returns a reporter's reputation[] index given their reputationID as the key
-    # total rep is all the rep in augur
-    # active rep is the fxpAmount which is actually active
-    # fork is the child branch of a parent which forked
-    # fxpDormant rep stores fxpDormant rep values
-    # reportedOnNonFinalRoundTwoEvent is the eventID if a person reported on a round 2 event before it was in the second round [i.e. the first reporting backstop], they cannot convert their rep to dormant or send rep until they've finished the resolution process for that round 2 event, set this to 0 once the event is final
-data Reporting[](reputation[](repValue, reporterID), numberReporters, repIDtoIndex[], totalRep, dormantRep[](repValue, reporterID), activeRep, reportedOnNonFinalRoundTwoEvent[])
+    # Reporting is a map of branch ID to branch details
+    # activeRep is a map of reporter address to attorep
+    # dormantRep is a map of reporter address to attorep. NOTE: address 0x0000000000000000000000000000000000000000 contains REP that needs to be redistributed
+    # totalActiveRep is all of the active attorep in augur
+    # totalDormantRep is all of the dormant attorep in augur (totalActiveRep + totalDormantRep is all REP in Augur)
+    # reporterLocks is a map of reporter address to reporting period that the reporter is locked for.  If the reporter is not locked (value = 0) then the reporter is free to deactivate their rep.  If the reporter is locked then the reporter is unable to deactivate their REP.
+data Reporting[](activeRep[], dormantRep[], totalActiveRep, totalDormantRep, reporterLocks[])
 
 data controller
 
 def init():
     self.controller = 0x0
     REPCONTRACT = self.controller.lookup('repContract')
-    self.Reporting[1010101].repIDtoIndex[1010101] = 0
     redistributionRep = REPCONTRACT.balanceOf(0x0000000000000000000000000000000000000000)
-    self.Reporting[1010101].reputation[0].repValue = redistributionRep
-    self.Reporting[1010101].reputation[0].reporterID = 1010101
-    self.Reporting[1010101].dormantRep[0].reporterID = 1010101
-    self.Reporting[1010101].numberReporters = 2
-    self.Reporting[1010101].repIDtoIndex[msg.sender] = 1
-    self.Reporting[1010101].reputation[1].repValue = 0
-    self.Reporting[1010101].reputation[1].reporterID = msg.sender
-    self.Reporting[1010101].dormantRep[1].reporterID = msg.sender
-    self.Reporting[1010101].repIDtoIndex[msg.sender] = 1
-    self.Reporting[1010101].activeRep += redistributionRep
-    self.Reporting[1010101].totalRep += redistributionRep
+    self.privateIncreaseDormantRep(1010101, 0x0000000000000000000000000000000000000000, redistributionRep)
 
-def getReportedOnNonFinalRoundTwoEvent(branch, reporter):
-    return(self.Reporting[branch].reportedOnNonFinalRoundTwoEvent[reporter])
+#######
+# FIXME: TODO: NOTE: DELETE ME!!!
+#######
+def giveAwayFreeRep(branch, reporterAddress, attorep):
+    self.privateIncreaseDormantRep(branch, reporterAddress, attorep)
+#######
+# FIXME: TODO: NOTE: DELETE ME!!!
+#######
 
-def setReportedOnNonFinalRoundTwoEvent(branch, event, reporter):
+def isReporterLocked(branch, reporterAddress):
+    return(self.Reporting[branch].reporterLocks[reporterAddress] != 0)
+
+def lockReporter(branch, reporterAddress, reportingPeriod):
     self.controller.checkWhitelist(msg.sender)
-    self.Reporting[branch].reportedOnNonFinalRoundTwoEvent[reporter] = event
+    self.Reporting[branch].reporterLocks[reporterAddress] = reportingPeriod
     return(1)
 
-# @return fxp
-def getActiveRep(branch):
-    return(self.Reporting[branch].activeRep)
-
-def adjustActiveRep(branch, fxpAmount):
+def unlockReporter(branch, reporterAddress):
     self.controller.checkWhitelist(msg.sender)
-    self.Reporting[branch].activeRep += fxpAmount
+    # TODO: figure out if we need to assert that they weren't already unlocked
+    self.Reporting[branch].reporterLocks[reporterAddress] = 0
     return(1)
 
-# @return fxp
-def getRepByIndex(branch, repIndex):
-    return(self.Reporting[branch].reputation[repIndex].repValue)
+def activateRepFor(branch, reporterAddress, attorepToActivate):
+    self.controller.checkWhitelist(msg.sender)
+    self.privateDecreaseDormantRep(branch, reporterAddress, attorepToActivate)
+    self.privateIncreaseActiveRep(branch, reporterAddress, attorepToActivate)
+    return(1)
 
-# @return fxp
-def getRepBalance(branch, address):
-    repIndex = self.Reporting[branch].repIDtoIndex[address]
-    if(!repIndex and self.Reporting[branch].reputation[repIndex].reporterID != address):
+# TODO: figure out what other constraints there are on rep deactivation and consider migrating those constraints to this contract
+def deactivateRepFor(branch, reporterAddress, attorepToDeactivate):
+    self.controller.checkWhitelist(msg.sender)
+    if (self.Reputation.reporterLocks[reporterAddress] != 0):
         throw()
-    return(self.Reporting[branch].reputation[repIndex].repValue)
+    self.privateDecreaseActiveRep(branch, reporterAddress, attorepToDeactivate)
+    self.privateIncreaseDormantRep(branch, reporterAddress, attorepToDeactivate)
+    return(1)
 
-def getDormantRepByIndex(branch, repIndex):
-    return(self.Reporting[branch].dormantRep[repIndex].repValue)
+def getActiveRepBalance(branch, reporterAddress):
+    return(self.Reporting[branch].activeRep[reporterAddress])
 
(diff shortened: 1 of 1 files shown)


MicahZoltu:
If any of the changes in here are undesirable, let me know and I can back them out.  I originally started with the intention of just making small changes, but as I was touching code I found more little things that I fixed along the way and ended up with this.

src/data_api/reporting.se
```diff
@@ -27,186 +27,161 @@ extern branches: [addCurrency:[int256,int256,int256,int256]:int256, addMarketToB
...
-data Reporting[](reputation[](repValue, reporterID), numberReporters, repIDtoIndex[], totalRep, dormantRep[](repValue, reporterID), activeRep, reportedOnNonFinalRoundTwoEvent[])
+    # Reporting is a map of branch ID to branch details
+    # activeRep is a map of reporter address to attorep
+    # dormantRep is a map of reporter address to attorep. NOTE: address 0x0000000000000000000000000000000000000000 contains REP that needs to be redistributed
+    # totalActiveRep is all of the active attorep in augur
+    # totalDormantRep is all of the dormant attorep in augur (totalActiveRep + totalDormantRep is all REP in Augur)
+    # numberReporters is the number of addresses that hold non-zero rep
+    # reporterLocks is a map of reporter address to an array of events that is causing them to be locked and the number of events causing them to be locked.  If there are no events then the reporter is free to deactivate their rep.  If there are any locking events the reporter is unable to deactivate their REP.
```
joeykrug: Being locked by one event is the same as being locked by n since you'll always be unlocked at the same time, so at least as it is now there's no need to store more

src/data_api/reporting.se
```diff
@@ -27,186 +27,161 @@ extern branches: [addCurrency:[int256,int256,int256,int256]:int256, addMarketToB
...
 
-# @return fxp
-def getActiveRep(branch):
-    return(self.Reporting[branch].activeRep)
+def unlockReporter(branch, reporterAddress, event):
+    self.controller.checkWhitelist(msg.sender)
+    if (self.Reporting[branch].reporterLocks[reporterAddress].events[event] == 0):
+        throw()
+    if (self.Reporting[branch].reporterLocks[reporterAddress].eventCount <= 0):
+        throw() # if we are in this state things are _really_ broken
```
joeykrug: Haha

src/data_api/reporting.se
```diff
@@ -27,186 +27,161 @@ extern branches: [addCurrency:[int256,int256,int256,int256]:int256, addMarketToB
...
+    return(1)
 
-def adjustActiveRep(branch, fxpAmount):
+def activateRep(branch, attorep):
+    return(self.activateRepFor(branch, attorep, msg.sender))
+
+def activateRepFor(branch, attorepToActivate, reporterAddress):
     self.controller.checkWhitelist(msg.sender)
-    self.Reporting[branch].activeRep += fxpAmount
+    # CONSIDER: should people who are locked be prevented from activating rep?
```
joeykrug: Nah there's no issue with activating rep during a lock, just deactivating

MicahZoltu:
Part of one of the changes to keep the contract in a healthy state was adding a function for `migrateReporterToNewBranch`.  I'm not actually sure what the correct business logic here is.  In this PR it _move_ the reporter's REP (active and dormant) over to the new branch, meaning it removes it from the old branch.  I'm not sure if this is the correct behavior or not.

If the expectation is that their REP exists on both branches, this creates a conundrum where the reporter could migrate to a new branch, then send their dormant REP on old branch away, then migrate again resulting in doubling their REP (repeat indefinitely).  It is possible I'm not fully understanding the logic here as well.

src/data_api/reporting.se
```diff
@@ -27,186 +27,161 @@ extern branches: [addCurrency:[int256,int256,int256,int256]:int256, addMarketToB
...
     self.controller.checkWhitelist(msg.sender)
-    reporterIndex = self.Reporting[branch].numberReporters
-    self.Reporting[branch].repIDtoIndex[sender] = reporterIndex
-    self.Reporting[branch].reputation[reporterIndex].repValue = fxpAmount
-    self.Reporting[branch].reputation[reporterIndex].reporterID = sender
-    self.Reporting[branch].dormantRep[reporterIndex].reporterID = sender
-    self.Reporting[branch].dormantRep[reporterIndex].repValue = fxpDormant
-    self.Reporting[branch].activeRep += fxpAmount + fxpRepToBonderOrBranch
-    self.Reporting[branch].totalRep += fxpAmount + fxpDormant + fxpRepToBonderOrBranch
+    # TODO: figure out if it is intended that both active and dormant rep are moved
```
joeykrug: Yes it is

src/data_api/reporting.se
```diff
@@ -27,186 +27,161 @@ extern branches: [addCurrency:[int256,int256,int256,int256]:int256, addMarketToB
...
+    # FIXME: calling this fully unlocks the sender as we aren't migrating event locks, is this OK?
```
joeykrug: As long as they aren't unlocked on the originalBranch, yes

src/data_api/reporting.se
```diff
@@ -27,186 +27,161 @@ extern branches: [addCurrency:[int256,int256,int256,int256]:int256, addMarketToB
...
     return(self.Reporting[branch].numberReporters)
 
-def repIDToIndex(branch, repID):
-    repIndex = self.Reporting[branch].repIDtoIndex[repID]
-    # means repID didn't matchup with 0 index reporterID
-    if(!repIndex and self.Reporting[branch].reputation[repIndex].reporterID != repID):
-        throw()
-    return(repIndex)
-
+# TODO: can this be renamed to "initializeBranch"?  Is that accurate?
```
joeykrug: Yes

src/data_api/reporting.se
```diff
@@ -27,186 +27,161 @@ extern branches: [addCurrency:[int256,int256,int256,int256]:int256, addMarketToB
...
+# TODO: figure out if this function should take in parent branch and migrate over REP at address 0x0000000000000000000000000000000000000000
```
joeykrug: Good catch, yes it should

joeykrug:
```Part of one of the changes to keep the contract in a healthy state was adding a function for migrateReporterToNewBranch. I'm not actually sure what the correct business logic here is. In this PR it move the reporter's REP (active and dormant) over to the new branch, meaning it removes it from the old branch. I'm not sure if this is the correct behavior or not.```

It should _not_ remove it from the old branch

```If the expectation is that their REP exists on both branches, this creates a conundrum where the reporter could migrate to a new branch, then send their dormant REP on old branch away, then migrate again resulting in doubling their REP (repeat indefinitely). It is possible I'm not fully understanding the logic here as well.```

I don't think this is possible / I've added checks to prevent this, but if you do find a way to actually do this, let me know

File: src/data_api/reporting.se [hunk 8]
```diff
@@ -27,186 +27,161 @@ extern branches: [addCurrency:[int256,int256,int256,int256]:int256, addMarketToB
...
+    if (self.Reporting[branch].reporterLocks[reporterAddress].events[event] == 0):
+        throw()
+    if (self.Reporting[branch].reporterLocks[reporterAddress].eventCount <= 0):
+        throw() # if we are in this state things are _really_ broken
+    self.Reporting[branch].reporterLocks[reporterAddress].events[event] = 0
+    self.Reporting[branch].reporterLocks[reporterAddress].eventCount -= 1
+    return(1)
 
-def adjustActiveRep(branch, fxpAmount):
+def activateRep(branch, attorep):
```
joeykrug: This function is basically useless, would remove

src/data_api/reporting.se
```diff
@@ -27,186 +27,161 @@ extern branches: [addCurrency:[int256,int256,int256,int256]:int256, addMarketToB
...
+    startingDormantRep = self.Reputation[branch].dormantRep[reporterAddress]
+    startingActiveRep = self.Reputation[branch].activeRep[reporterAddress]
+    self.Reputation[branch].dormantRep[reporterAddress] = safeSub(startingDormantRep, attorepToActivate)
+    self.Reputation[branch].activeRep[reporterAddress] = safeAdd(startingActiveRep, attorepToActivate)
     return(1)
 
-# @return fxp
-def getRepByIndex(branch, repIndex):
-    return(self.Reporting[branch].reputation[repIndex].repValue)
+def deactivateRep(branch, attorep):
```
joeykrug: ditto for this one re removing, any account that holds rep will never be calling deactivateRep and so the msg.sender will never be right

src/data_api/reporting.se
```diff
@@ -27,186 +27,161 @@ extern branches: [addCurrency:[int256,int256,int256,int256]:int256, addMarketToB
...
-    self.Reporting[branch].reputation[reporterIndex].reporterID = sender
-    self.Reporting[branch].dormantRep[reporterIndex].reporterID = sender
-    self.Reporting[branch].dormantRep[reporterIndex].repValue = fxpDormant
-    self.Reporting[branch].activeRep += fxpAmount + fxpRepToBonderOrBranch
-    self.Reporting[branch].totalRep += fxpAmount + fxpDormant + fxpRepToBonderOrBranch
+    # TODO: figure out if it is intended that both active and dormant rep are moved
+    # FIXME: calling this fully unlocks the sender as we aren't migrating event locks, is this OK?
+    originalActiveRep = self.Reporting[originalBranch].activeRep[reporterAddress]
+    originalDormantRep = self.Reporting[originalBranch].dormantRep[reporterAddress]
+    self.Reporting[originalBranch].activeRep[reporterAddress] = 0
```
joeykrug: user shouldn't lose rep on original branch

src/data_api/reporting.se
MicahZoltu: Hmm, perhaps I'm not understanding something.  My understanding was that `reporterAddress` throughout this contract generally refers to someone's account/wallet.  Are you saying that `reporterAddress` is not actually a wallet that a user controls but instead is some intermediate contract (where there is one per user account/wallet)?  If so, where can I find the details on that?  If not, why wouldn't a user be able to call `activateRep` or `deactivateRep`?

MicahZoltu: Or perhaps the argument is that activating/deactivating rep isn't an action that a user would every want to take, instead it is intended to be an automated process (in which case I'm probably not fully grokking active/dormant rep).  I thought a user would generally have REP dormant when they wanted to trade it or were unable to report (e.g., vacation) and then when they wanted to start receiving fees they would activate.

src/data_api/reporting.se
MicahZoltu: Can `reporterLock` function correctly if it is just `1` or `0`?  Or does it need to store at least one of the events that caused it to get locked?

src/data_api/reporting.se
joeykrug: None of the setters in data_api are meant to be called by users. They're meant to be places to store + access state but have almost no business logic, which is done in the functional contracts. You'll notice this function, nor does the function that it calls that you wrote actually check whether a user's rep can be activated or not, that's done in the functions outside of this file that convertRepToDormant or convertRepToActive.  So no user should be able to call activateRep.  With that established, no _contract_ of ours should call it either, b/c then msg.sender would be the contract, not the user of the rep it meant to activate.  activateRepFor is thus the function that'd actually be used

joeykrug: That's correct --- a user would want to activate or set dormant their rep, they just wouldn't call a bare metal function like this, they'd go through one of the public functions in functions/convertRepBlahBlah

MicahZoltu: Ah, I didn't realize that there were more constraints on activating REP besides the locking logic contained in this contract (now named `reporterLocks`.

src/data_api/reporting.se
joeykrug: It needs either an event [or alternatively, the reporting period in which the reporter got locked]

MicahZoltu:
I believe I have applied all of the changes discussed/pointed out here.  I'll move forward with applying changes to all of the call sites and comment on this PR once they are ready for final review.

src/data_api/reporting.se
```diff
@@ -27,202 +27,147 @@ extern branches: [addCurrency:[int256,int256,int256,int256]:int256, addMarketToB
...
-
-def getReportedOnNonFinalRoundTwoEvent(branch, reporter):
-    return(self.Reporting[branch].reportedOnNonFinalRoundTwoEvent[reporter])
-
-def setReportedOnNonFinalRoundTwoEvent(branch, event, reporter):
+    self.privateIncreaseDormantRep(1010101, 0x0000000000000000000000000000000000000000, redistributionRep)
+
+#######
+# FIXME: TODO: NOTE: DELETE ME!!!
+#######
```
MicahZoltu: I wish `<marquee>` HTML tags worked in source code...

joeykrug:
Gonna close b/c out of date with new reporting system

PR #150, opened 2017-04-30
Switches to new reputation scheme.

MicahZoltu:
See https://www.websequencediagrams.com/files/render?link=kUm7MBHLoO87M3m2dXzE for an overview of the flow.


joeykrug:
Also noticed - current pattern for stuff at top of file is extern -> inset -> data -> event but some files don't follow that

Will push a commit making everything compliant w/ that

joeykrug:
Types for all the tokens should be consistent across contract / function when they're the same part of erc20 as well as consistent with erc20 types, not all are atm

joeykrug:
When doing an internal function call instead of blah() do self.blah(), fixing these, also we need unique function names to get externs to work properly + compile [also fixing these], you can view all this stuff in the next commit I'm gonna make once it all compiles :)

src/extensions/reputationFaucet.se
```diff
@@ -21,6 +21,7 @@ data controller
 event fundedAccount(branch: indexed, sender: indexed, cashBalance, repBalance, timestamp)
 
 ### Reputation faucet - get ~free rep!
+### TODO Fix to work with new system
```
joeykrug: Needs fixing to work with new system

src/data_api/registrationToken.se
```diff
@@ -0,0 +1,113 @@
...
+def any():
+    refund()
+
+def register(branch, reportingWindow):
+    # do not allow for registration for reporting in the current window
+    if (reportingWindow <= self.controller.lookup('reporting').getCurrentReportingWindow()):
+        ~invalid()
+    self.controller.lookup('reputationToken').transferFrom(branch, msg.sender, self, bondAmount)
+    startingTokens = self.tokens[branch][reportingWindow][msg.sender]
+    self.tokens[branch][reportingWindow][msg.sender] = safeAdd(startingRegistrationTokens, 1)
```
joeykrug: is this supposed to be `startingTokens` here?

src/data_api/registrationToken.se
```diff
@@ -0,0 +1,113 @@
...
+
+def migrateRegistrationToken(parentBranch, childBranch, reportingWindow, tokensToMigrate):
+    branchesContract = self.controller.lookup('branches')
+    # validate that the source branch is actually forked/forking
+    parentForkingEvent = branchesContract.getForkingEvent(parentBranch)
+    if (parentForkingEvent == 0):
+        ~invalid()
+    # validate that destinationBranch is a child of sourceBranch
+    sourceBranchHashHighBits = sha3(sourceBranch) & 0xffffffffffffffff000000000000000000000000000000000000000000000000
+    destinationBranchParentHashHighBits = destinationBranch & 0xffffffffffffffff000000000000000000000000000000000000000000000000
```
joeykrug: This takes the first 64 bits, is that right or am I misunderstanding?

src/data_api/registrationToken.se
```diff
@@ -0,0 +1,113 @@
...
+
+def redeem(branch, reportingWindow):
+    startingTokens = self.tokens[branch][reportingWindow][msg.sender]
+    startingSupply = self.tokenSupply[branch][reportingWindow]
+    reportingContract = self.controller.lookup('reporting')
+    reportingContract.assertDoneReporting(branch, reportingWindow, msg.sender)
+    self.tokens[branch][reportingWindow][msg.sender] = safeSub(startingTokens, 1)
+    self.tokenSupply[branch][reportingWindow] = safeSub(startingSupply, 1)
+    reportingContract.clearUserReports(branch, reportingWindow, msg.sender)
+    self.controller.lookup('reputation').transfer(branch, msg.sender, self.bondAmount)
```
joeykrug: Just a reminder if bond is dynamic then it'll need to be something like self.bondAmount[reportingWindow]

src/data_api/registrationToken.se
```diff
@@ -0,0 +1,113 @@
...
+
+####
+#### ERC20-like implementation
+####
+
+def totalSupply(branch, reportingWindow):
+	return(self.tokenSupply[branch][reportingWindow]: uint256)
```
joeykrug: has an extra indent

src/data_api/registrationToken.se
```diff
@@ -0,0 +1,113 @@
...
+
+def registrationTokenBalance(branch, reportingWindow, address):
+	return(self.tokens[branch][reportingWindow][address]: uint256)
```
joeykrug: ditto re extra indent

joeykrug:
Another thing I noticed is this stuff isn't currently written to work very well with upgrading code, it should however --- the way the current contracts do that is they have data contracts which have getters and setters for their various data, then the business logic / functions contracts can be upgraded

src/data_api/reporting.se
```diff
@@ -1,220 +1,114 @@
...
+    self.reportingDisputePeriodSeconds = 3 * SECONDS_PER_DAY
+    self.reportsPerEvent = 7
+
+def any():
+    refund()
+
+# TODO: simplify this to allow people to report on any event, as long as it doesn't already have too many reporters
+# only to be called for events being reported on by a subset of reporters, up to caller to make sure the event is generally elligible for reporting right now
+def assertCanReportOnEvent(eventId, reporterAddress):
+    # CONSIDER: change this scheme to just disallowing more than `n` reports per event, but allow anyone to report on anything
```
joeykrug: I like this

src/data_api/reporting.se
```diff
@@ -1,220 +1,114 @@
...
+    self.userReports[branch][currentReportingWindow][reporterAddress].reportedEvents[startingEventCount] = eventId
+    self.userReports[branch][currentReportingWindow][reporterAddress].reportedEventsCount = safeAdd(startingEventCount, 1)
     return(1)
 
-def suicideFunds(to: address):
-    if(msg.sender != self.controller):
-        ~invalid()
-    suicide(to)
\ No newline at end of file
+def isInReportingPhase():
```
joeykrug: I like the way you did this, nice and simple

src/data_api/reporting.se
```diff
@@ -1,220 +1,114 @@
...
-    self.Reporting[branch].dormantRep[index].repValue += fxpValue
-    return(1)
-
-def subtractDormantRep(branch, index, fxpValue):
-    self.controller.assertIsWhitelisted(msg.sender)
-    self.Reporting[branch].dormantRep[index].repValue -= fxpValue
+# FIXME: if an event moves out of a reporting period (e.g., due to a fork) this may start to return an incorrect result
+def assertDoneReporting(branch, reportingWindow, reporterAddress):
+    eventsContract = self.controller.lookup('events')
+    totalAllReporterEvents = eventsContract.getNumberOfAllReporterEvents(branch, reportingWindow)
```
joeykrug: Are L45 and L46 needed?

joeykrug:
There are also circular / co-dependencies which should be avoided

src/data_api/events.se
```diff
@@ -1,346 +1,420 @@
...
+    # FIXME: need to deal with fees here as well, remaining_fees_in_period / remaining_correct_outcome_tokens_in_period
+    self.finalizeEvent(eventId)
+    finalOutcome = self.events[eventId].finalOutcome
+    if (outcome != finalOutcome):
         ~invalid()
-    self.controller = newController
+    startingTokenSupply = self.events[eventId].tokenSupply[branch][outcome]
+    startingAllOutcomesTokenSupply = self.events[eventId].allOutcomesTokenSupply[branch]
+    attotokens = self.events[eventId].tokens[branch][outcome][msg.sender]
+    # CONSIDER: can rounding errors leave us with a few attorep sitting around?  do we care?
```
joeykrug: Probably, probably not are my two cents

src/data_api/events.se
```diff
@@ -1,346 +1,420 @@
...
+    # if we aren't through the scheduled reporting window yet then the event can't be finalized
+    currentReportingWindow = self.controller.lookup('reporting').getCurrentReportingWindow()
+    scheduledReportingWindow = self.events[eventId].reportingWindow
+    nowIsBeforeOrDuringScheduledReportingWindow = currentReportingWindow <= scheduledReportingWindow
+    if (nowIsBeforeOrDuringScheduledReportingWindow):
+        return(0)
+    # disputes and following forks changes the reporting window, so if we make it to the end of a reporting window we know the event can be finalized
+    finalOutcome = self.events[eventId].tentativeOutcome
+    self.events[eventId].finalOutcome = finalOutcome
+    if (self.events[eventId].automatedReporterOutcome == finalOutcome && self.events[eventId].automatedReporterDisputeBondHolder != 0):
```
joeykrug: If it is `!=` to final outcome we should pay the bond holder 2x the bond amount. If you're disputing an outcome and the final outcome was the same you should lose the bond, but if you're disputing and the final outcome was _different_ then your dispute was needed / worthwhile, so you get 2x the bond back.

src/data_api/events.se
```diff
@@ -1,346 +1,420 @@
...
+        # TODO: pay bond holder 2x bond amount, pulling from incorrect reporting pool. NOTE: this needs to decrease available funds privateRedeemCorrectOutcomeTokens pulls from
+        return(1)
+    if (self.events[eventId].limitedReportersOutcome == finalOutcome && self.events[eventId].limitedReportersDisputeBondHolder != 0):
```
joeykrug: If it is `!=` to final outcome we should pay the bond holder 2x the bond amount. If you're disputing an outcome and the final outcome was the same you should lose the bond, but if you're disputing and the final outcome was _different_ then your dispute was needed / worthwhile, so you get 2x the bond back.

src/data_api/events.se
```diff
@@ -1,346 +1,420 @@
...
+        # TODO: pay bond holder 2x bond amount, pulling from incorrect reporting pool. NOTE: this needs to decrease available funds privateRedeemCorrectOutcomeTokens pulls from
+        return(1)
+    if (self.events[eventId].allReportersOutcome == finalOutcome && self.events[eventId].allReportersDisputeBondHolder != 0):
```
joeykrug: If it is `!=` to final outcome we should pay the bond holder 2x the bond amount. If you're disputing an outcome and the final outcome was the same you should lose the bond, but if you're disputing and the final outcome was _different_ then your dispute was needed / worthwhile, so you get 2x the bond back.

src/data_api/events.se
```diff
@@ -1,346 +1,420 @@
...
     return(1)
 
-def getExtraBondPoster(event):
-    return(self.Events[event].extraBondPoster)
-
-def setController(newController: address):
-    if(msg.sender != self.controller):
+def privateRedeemCorrectOutcomeTokens(eventId, branch, outcome, tokenHolderAddress):
+    privateOnly()
+    # FIXME: need to deal with fees here as well, remaining_fees_in_period / remaining_correct_outcome_tokens_in_period
```
joeykrug: I also think it's not considering failed bonds [ones where the final outcome was no different than the outcome when posted] in the rep redistribution here as well

src/data_api/events.se
```diff
@@ -1,346 +1,420 @@
...
-def getRejectedPeriod(event):
-    return(self.Events[event].rejectedPeriod)
-
-def setExtraBondPoster(event, poster):
-    self.controller.assertIsWhitelisted(msg.sender)
-    self.Events[event].extraBondPoster = poster
+# if this event is the cause of a fork, force tokens to be redeemed on the branch associated with the token's outcome
+def privateTryRedeemTokensOnForkingEvent(eventId, branch, outcome, tokenHolderAddress):
+    privateOnly()
+    # FIXME: need to redeem fees, but the current implementation just redeems tokens on outcome branch. CONSIDER: change this to only allow token redemption after event is finalized, then we can also distribute fees on correct outcome branch
```
joeykrug: Reminder to consider failed bonds [ones where the final outcome was no different than the outcome when posted] in the rep redistribution here as well --- this is in the bottom section of the diagram

joeykrug:
Will also need: 

- [ ] `getEventsWithSubmittedReport(branch, period, reporter)`
- [ ] getters for num reports per event for UI 
- [ ] `marketInfo.se` will need updating
- [ ] `marketsInfo.se` updating
- [ ] a new reputation faucet
- [ ] `eventBondResolution.se` made to comply with new system
- [ ]  get anything else in other contracts to comply with new simplified reporting system that doesn't yet

Also in reporting.se ```self.userReports[branch][currentReportingWindow][reporterAddress].reportedEventsCount = safeAdd(startingEventCount, 1)``` we shouldn't increment it if a user has already reported on a given event

joeykrug:
Currently events can be created on a branch that's already forked, I think our forking scheme prohibits that

Also should people be able to transfer rep on a branch that's already forked or only migrate

src/data_api/events.se
```diff
@@ -1,346 +1,420 @@
...
+    if (self.events[eventId].limitedReportersDisputeBondHolder != 0):
+        self.events[eventId].limitedReportersDisputeBondHolder = 0
+        # self.controller.lookup('reputationToken').transfer(self.events[eventId].limitedReportersDisputeBondHolder, self.limitedReportersDisputeBondAmount)
+    if (self.events[eventId].allReportersDisputeBondHolder != 0):
+        self.events[eventId].allReportersDisputeBondHolder = 0
+        # self.controller.lookup('reputationToken').transfer(self.events[eventId].allReportersDisputeBondHolder, self.allReportersDisputeBondAmount)
+    # follow the forking event to its branch
+    self.events[eventId].branch = self.events[forkingEventId].branch
+    # reset reporting state back to un-reported (except for automated)
+    self.events[eventId].reportingWindow = self.controller.lookup('reporting').getCurrentReportingWindow() + 1
```
joeykrug: This is correct if the event's reporting window was in the past / the event should expire asap, but if it expires in the future make it the same future window as before [so no update].

so if reportingWindow < currentWindow + 1, update it to currentWindow + 1

otherwise don't

joeykrug:
- event should have a validity bond, the way this works is:
                if(eventHasToBeReportedOnByReporters and finalOutcome != automatedOutcome):
                    send half to branch
                    send half to the challenger
                elif(finalOutcome == INDETERMINATE):
                    if(eventHasToBeReportedOnByReporters):
                        send to branch
                    # event is reported on by the automated resolver
                    else:
                        send to automated resolver
                else:
                    # automated resolver resolved the event and it was determinate
                    return bond to bond poster

joeykrug:
Events need `max, min, numOutcomes, resolutionSource:str, resolutionSourceLength, type` in their creation as things on develop that are still needed in the new system

src/data_api/events.se
```diff
@@ -1,346 +1,420 @@
...
+    self.events[eventId].endTime = endTime
+    self.events[eventId].description = description
+    self.events[eventId].automatedReporterAddress = automatedReporterAddress
+    earliestReportingTime = endTime
+    if (automatedReporterAddress != 0):
+        earliestReportingTime = endTime + 6 * SECONDS_PER_DAY
+    self.events[eventId].reportingWindow = self.controller.lookup('reporting').getReportingWindowForTimestamp(earliestReportingTime) + 1
+    return(eventId)
+
+def report(eventId, outcome, attorepStake):
```
joeykrug: Should also assert something like 

```
    max = EVENTS.getMaxValue(event)
    min = EVENTS.getMinValue(event)
    if(outcome < min):
        outcome = min
    elif(outcome > max):
        outcome = max
```

Or alternatively throw if the outcome is > max or < min

joeykrug:
To merge markets and events together into just one contract `events`, events also need to have ```tradingFee, tag1, tag2, tag3, gasSubsidyToClose, shareContracts[], shareWallets[], currency``` as part of events

The remaining machinery from markets ```volume, orderIDs[](id, nextID, prevID), lastOrder, totalOrders, prices[]``` I'm going to move to `orders.se` and commit it to develop

joeykrug:
For reference: https://github.com/AugurProject/augur-core/commit/c3da33ae3146b9a79b5765e5e685de327e1ede8e --- basic overview of how creation fee can only be paid in eth --- markets can be created denominated in other markets' shares [but not other currencies yet] and those ones pay no creation fee or complete set fees b/c they're already paid in the base market's currency [ETH for now].

src/data_api/events.se
```diff
@@ -1,360 +1,420 @@
...
+    if (!self.tryFinalizeEvent(eventId)):
+        ~invalid()
 
-def addReportersPaidSoFar(event):
-    self.controller.assertIsWhitelisted(msg.sender)
-    self.Events[event].reportersPaidSoFarForEvent += 1
+# will return 0 if event is not finalized by the time this method returns
+def tryFinalizeEvent(eventId):
+    # FIXME: handle finalizing forked events
+    # FIXME: when the event is finalized, we need to add its fees to the pool for the ending reporting window and we need to add the count of correct outcome tokens to the pool for the window.  These are necessary for fee collection
```
joeykrug:
```
    # send fees in market to branch
    marketCurrency = INFO.getCurrency(market)
    marketWallet = INFO.getWallet(market)
    marketBalance = marketCurrency.balanceOf(marketWallet)
    marketShareValue = safeFxpMul(MARKETS.getSharesPurchased(market, 1), MARKETS.getCumulativeScale(market))
    if(marketBalance > marketShareValue and MARKETS.getTradingFee(market) != 0):
        marketWallet.transfer(BRANCHES.getBranchWallet(branch, marketCurrency), marketBalance - marketShareValue)

```

Current code on develop for this to give you an idea.  Main meat is `balance Of Market` - `complete sets outstanding`*`range` gives you fees to distribute

joeykrug:
[A New Fee Structure For Securing Reporting for Markets.docx](https://github.com/AugurProject/augur-core/files/1020588/A.New.Fee.Structure.For.Securing.Reporting.for.Markets.docx)

joeykrug:
I'm handling the escape hatch stuff for trading, for reporting/markets if a critical bug is found there that involves rep itself or bond, there should also be an escape hatch that is usable `onlyInEmergency` [i.e. after the contracts are frozen] that:

- Allows a user to claim their registration bonds
- Allows a user to claim any other sort of bond back for rep [i.e. challenge bonds]
- Allows a user to claim their outcome tokens rep back for regular rep
- Allows a user to claim their market validity bond

This would all be in one function, i.e. escape(): then it'd try to do those 4 things.  Then users could use the regular rep erc-20 functions to burn/transfer their rep over to the new reporting contracts just like we're allowing on the very first v1 contract launch.

joeykrug:
Was skimming some of the latest code, noticed that many functions dont return anything in the event of success (1 is what we've been using across the project, it's useful when other contracts want to check return values) and there's some `return` with no value, which actually does nothing / doesn't return at all

MicahZoltu:
I tend to _avoid_ returning 1 for success of functions as a placeholder because it signals to the user that reading the return value is a reasonable strategy.  Most of the functions I author are succeed-or-throw functions, there is no way to fail and still return.  When I _do_ return it is very explicit and not only part of the function name, but these are the only functions that have return values which provides a secondary indicator that _YOU SHOULD LOOK AT THE RETURN VALUE!_.

As for the `return` with no value, that is definitely annoying.  There is one spot where I break the above rule because I need to early return and Serpent doesn't allow value-less returns (dumb).

joeykrug:
>  Most of the functions I author are succeed-or-throw functions, there is no way to fail and still return.

This isn't strictly true. If I call a function with y gas and I have some larger number x in the caller and I can't check the return value, I don't know whether it succeeded or failed, but a throw will not result in my transaction failing / ending.

The return with no value in the code is a bug (i saw one iirc but yeah it doesn't allow valueless returns so that'll need to be changed)

tinybike:
I think there's value in following the same conventions (such as always returning a value from functions) throughout the code.  We've always followed this particular convention closely; in fact, the middleware will assume the transaction has failed if it does not receive a return value.

joeykrug:
It's also annoying to simulate whether a call was successful, b/c with a throw it just won't return anything

src/data_api/events.se
```diff
@@ -1,346 +1,420 @@
...
+def any():
+    refund()
+
+def createEvent(branch, endTime, description, automatedReporterAddress):
+    # FIXME: assert the user sent enough fees and validity bond
+    # FIXME: assert not a duplicate event. CONSIDER: is this actually necessary?
+    # FIXME: add event to tally of total events in reporting window, also be sure to move this anytime the event changes reporting windows
+    eventId = self.nextEventId
+    self.nextEventId += 1
+    self.events[eventId].branch = branch
```
MicahZoltu: FIXME in the code to address this.

src/data_api/events.se
```diff
@@ -1,346 +1,420 @@
...
+        ~invalid()
+    if (msg.sender != self.events[eventId].automatedReporterAddress):
+        ~invalid()
+    if (self.events[eventId].automatedReporterOutcome != 0):
+        ~invalid()
+    if (block.timestamp < self.events[eventId].endTime):
+        ~invalid()
+    if (block.timestamp > self.getAutomatedReportDueTimestamp(eventId)):
+        ~invalid()
+    self.events[eventId].automatedReporterOutcome = outcome
```
MicahZoltu: @adrake33 is addressing this in his branch.

src/data_api/events.se
```diff
@@ -1,346 +1,420 @@
...
+    if (self.isEventFinalized(eventId)):
+        ~invalid()
+    if (self.events[eventId].automatedReporterOutcome == 0):
+        ~invalid()
+    if (self.events[eventId].automatedReporterDisputeBondHolder != 0):
+        ~invalid()
+    self.assertEventIsInAutomatedReportDisputePhase(eventId)
+    # self.controller.lookup('reputationToken').transferFrom(msg.sender, self, self.automatedReporterDisputeBondAmount)
+    self.events[eventId].automatedReporterDisputeBondHolder = msg.sender
+    self.events[eventId].automatedReporterOutcome = self.events[eventId].tentativeOutcome
```
MicahZoltu: This will change a bit in @adrake33's branch re automated reporting, but for limited/all reporting the tentative outcome is whatever has been reported at the time of the dispute.

src/data_api/events.se
MicahZoltu: All dispute bond related things addressed in @adrake33's PR.

src/data_api/events.se
```diff
@@ -1,346 +1,420 @@
...
+    reportingWindow = self.events[eventId].reportingWindow
+    numberOfRegistrationTokens = registrationTokenContract.registrationTokenBalance(branch, reportingWindow)
+    if (numberOfRegistrationTokens < 1):
+        ~invalid()
     return(1)
 
-def getEventBranch(event):
-    return(self.Events[event].branch)
+# TODO: implement ERC20-like interface for each of the 3 different bond tokens
+# CONSIDER: is it worth making them ERC20-like?  Perhaps just add a transfer/balance function instead?
```
MicahZoltu: They ended up as ERC20 tokens in @adrake33's changes mainly because they ended up as separate contracts to simplify code and help get some code out of market.se (which is too big) and adding the ERC20 interface was trivial at that point (just copy/paste).

MicahZoltu:
Things that still need to be done post-merge aside from FIXME and TODO comments:
* Reporting escape hatch.
* Mutexes.
* Extensions (need to work with the UI team to figure out exactly what they need).
* Potentially some controller stuff, though at the moment everything should be secured against attack without any need for whitelisting.
* Make all functions return 1 at the end rather than returning 0 (default behavior)

I went through all of the feedback you left previously @joeykrug and I think it is all no longer relevant, has been addressed, or is now captured in a TODO/FIXME or the above list.  I believe this branch is ready for a final review so we can merge into develop.  My argument at this juncture is that we should try to defer as much as possible to post-merge as most of the contract developers are now working off of this branch and the UI team is targeting this branch so having it unmerged is just causing us headaches.

PR #154, opened 2017-05-01
John

JohnDanz:
Fixed a typo in functions/roundTwoPenalize.se that was causing an error. added mutex tests to data_api_tests.py


tinybike:
May I suggest giving your PR a more informative title than "John"? :grin:

PR #166, opened 2017-05-11
Updates for data_api unit tests

JohnDanz:
Updates the data_api_tests.py file to include more unit tests for files in the data_api folder. 

Changes also include a fix for `consensusData.se` where `setRefund` wasn't setting the refund properly:
```
def setRefund(refunder, amount):
    self.controller.assertIsWhitelisted(msg.sender)
    self.refunds[refunder] = amt
    return(1)
```
became:
```
def setRefund(refunder, amount):
    self.controller.assertIsWhitelisted(msg.sender)
    self.refunds[refunder] = amount
    return(1)
```

as you can see, we were taking in the arguments of `refunder` and `amount` but we were attempting to set the `amt` value which doesn't exist. We now properly use `amount`.


joeykrug:
Good fix re refunder

PR #176, opened 2017-05-16
Adds stack.

MicahZoltu:
I have had this sitting in my repo for a while.  No applied usage of it yet, but I would like feedback on it and if it is seen as being useful we can leverage it (not sure where exactly yet).

If there is no desire to have it merged into augur-core, I'll probably leave this branch up for a bit until I find a better home.

### DIFF
--- a/src/extensions/stack.se
+++ b/src/extensions/stack.se
@@ -1,0 +1,28 @@
+data stacks[](items[], head)
+
+# should only be used with call = delegate
+def push(key, item):
+    index = self.stacks[key].head + 1
+    self.stacks[key].head = index
+    self.stacks[key].items[index] = item
+
+# should only be used with call = delegate
+def pop(key, item):
+    index = self.stacks[key].head
+    if(index == 0):
+        ~invalid()
+    self.stacks[key].head = index - 1
+    removedValue = self.stacks[key].items[index]
+    self.stacks[key].items[index] = 0
+    return(removedValue)
+
+# should only be used with call = delegate
+def peek(key):
+    index = self.stacks[key].head
+    if(index == 0):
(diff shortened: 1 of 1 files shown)


src/libraries/stack.se
```diff
@@ -0,0 +1,33 @@
...
+    return(removedValue)
+
+# should only be used with call = delegate
+def peek(key):
+    index = self.stacks[key].head
+    if (index == 0):
+        ~invalid()
+    return(self.stacks[key].nodes[index].item)
+
+def isEmpty():
```
joeykrug: This is missing a `key` param

joeykrug:
Also I think 

```
data stacks[](items[], head)

def push(key, item):
	self.stacks[key].head += 1
	index = self.stacks[key].head
	self.stacks[key].items[index] = item
	return(1)

def pop(key, item):
	index = self.stacks[key].head
	if(index == 0):
		~invalid()
	self.stacks[key].head = self.stacks[key].head - 1
	removedValue = self.stacks[key].items[index]
	self.stacks[key].items[index] = 0
	return(removedValue)

def peek(key):
	index = self.stacks[key].head
	if(index == 0):
		~invalid()
	return(self.stacks[key].items[index])

def isEmpty(key):
	return(self.stacks[key].head == 0)
```

Does the same thing at ~half or so the cost

MicahZoltu:
Ah, heh.  When I wrote this originally it was a doubly linked list that allowed for removing from the middle.  I simplified it over the course of several iterations and didn't realize I had reached the point where one could just use an array.  Will update to your version.

MicahZoltu:
Applied changes, ready for re-review.

PR #178, opened 2017-05-22
Tests for scalar markets

tinybike:
Added payout, make order, and take order tests for scalar markets.


src/functions/scalarPayouts.se
```diff
@@ -37,7 +38,7 @@ def payoutScalarMarket(sender: address, market: address, event: address):
     minValue = EVENTS.getMinValue(event)
     maxValue = EVENTS.getMaxValue(event)
     currency = INFO.getCurrency(market)
-    branch = EVENTS.getExpiration(MARKETS.getMarketEvent(market))
+    branch = EVENTS.getEventBranch(MARKETS.getMarketEvent(market))
```
joeykrug: Good catch

PR #183, opened 2017-06-07
Added new controller tests

charltonh:
Might still need some tweaks, such as perhaps the format or expected output of the testing.  Please leave any comments.

### DIFF
--- a/tests/controller_test.se
+++ b/tests/controller_test.se
@@ -1,0 +1,37 @@
+
+# Make a file controller_test.se which implements functions that use all
+# the functionality of controller.se
+# 
+# Write a controller_tests.py file which tests the controller by calling
+# various controller.se and controller_test.se functions --- current testing
+# contract examples available in the tests folder.
+
+# self.controller = 0x0 [this is the controller contract's address, you
+# have a few ways in getting it here, one is to get the address of the
+# controller contract once you've uploaded it in the pyethereum tester,
+# another is to just upload the contract here using create]. from binascii
+# import hexlify hexlify(c.address) in python and the Serpent wiki above
+# should let you figure out all the options here
+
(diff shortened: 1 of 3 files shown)


tests/controller_test.se
```diff
@@ -0,0 +1,37 @@
...
+# self.controller = 0x0 [this is the controller contract's address, you
+# have a few ways in getting it here, one is to get the address of the
+# controller contract once you've uploaded it in the pyethereum tester,
+# another is to just upload the contract here using create]. from binascii
+# import hexlify hexlify(c.address) in python and the Serpent wiki above
+# should let you figure out all the options here
+
+extern controller: [addToWhitelist:[int256]:int256, assertIsWhitelisted:[int256]:int256, assertOnlySpecifiedCaller:[int256,int256]:_, changeMode:[int256]:int256, emergencyStop:[]:int256, getMode:[]:int256, getOwner:[]:int256, lookup:[int256]:int256, onlyInEmergency:[]:_, release:[]:int256, removeFromWhitelist:[int256,int256]:int256, setValue:[int256,int256]:int256, stopInEmergency:[]:_, suicide:[int256,int256,int256]:int256, switchModeSoOnlyEmergencyStopsAndEscapeHatchesCanBeUsed:[]:_, transferOwnership:[int256,int256,int256,int256]:int256, updateController:[int256,int256]:int256]
+
+inset('../src/functions/controller.se')
```
joeykrug: This is wrong, you can't inset the controller into every file that uses it, this makes all the tests meaningless.

tests/controller_tests.py
```diff
@@ -0,0 +1,150 @@
...
+	try:
+		raise Exception(c.suicideFunds(t.a1, sender=t.k1))
+	except Exception as exc:
+		assert(isinstance(exc, ethereum.tester.TransactionFailed)), "suicideFunds should fail when attempted by non-controller address"
+	# Test with another non-controller address  (redundant?)
+	try:
+		raise Exception(c.suicideFunds(t.a2, sender=t.k2))
+	except Exception as exc:
+		assert(isinstance(exc, ethereum.tester.TransactionFailed)), "suicideFunds should fail when attempted by non-controller address"
+	# Test with controller address
```
joeykrug: t.k0 is *not* the controller --- it's the _owner_ of the controller.  suicideFunds should *never* be called directly, in fact it's actually impossible to do so unless the controller is calling it which is mistakenly set to the contract itself here due to the inset

joeykrug:
Just committed you readme updates, those were all good ++

PR #186, opened 2017-06-10
Added order insertion point finding to orders.se

tinybike:
Includes fixes for two bugs identified during audits -- thanks Sergio!


MicahZoltu:
Note: There are a few comments that GitHub is hiding because the code changed, recommend reading them to see if they are still relevant.

src/functions/binaryForkResolve.se
```diff
@@ -42,7 +42,7 @@ extern subcurrency: [allowance:[address,address]:uint256, approve:[address,uint2
 
 extern forkResolveContract: [resolveFork:[int256]:int256]
 
-inset('../macros/refund.sem')
+inset('../macros/assertZeroValue.sem')
```
tinybike: That's fine, I just did an all-file search-and-replace.  I'm guessing these reporting-related files have been deleted altogether in your branch, right?

src/data_api/orders.se
```diff
@@ -155,105 +155,79 @@ def getWorstBidOrderID(market: address, outcome: uint256):
...
-        # If price is higher than the best ask price:
-        #  - betterOrderID must be specified
-        #  - price must be greater than than or equal to betterOrderID's price
-        else:
-            if(betterOrderID == 0):
-                ~invalid()
-            if(fxpPrice < self.orders[betterOrderID].fxpPrice):
-                ~invalid()
-    return(1: uint256)
+        isWorstPrice = fxpPrice >= self.orders[worseOrderID].fxpPrice
```
tinybike: The difference is intentional in this case, since this function walks _up_ the order list.  i.e., if a new order has an equivalent price to an existing order, it's considered "worse" so that we respect price-time priority.

src/data_api/orders.se
```diff
@@ -155,105 +155,79 @@ def getWorstBidOrderID(market: address, outcome: uint256):
...
+        isWorsePrice = self.isWorsePrice(type, fxpPrice, self.orders[betterOrderID].worseOrderID)
+        if(isWorsePrice or fxpPrice == self.orders[self.orders[betterOrderID].worseOrderID].fxpPrice):
+            betterOrderID = self.orders[betterOrderID].worseOrderID
+    worseOrderID = self.orders[betterOrderID].worseOrderID
+    return([betterOrderID, worseOrderID]: arr)
+
+# @public
+def findBoundingOrders(type: uint256, fxpPrice, bestOrderID: address, worstOrderID: address, betterOrderID: address, worseOrderID: address):
+    if(bestOrderID == worstOrderID):
+        return([bestOrderID, bestOrderID]: arr)
```
tinybike: This is done in the next step -- the new best/worstOrderIDs are looked up, then compared to the current order.  It could be worth re-doing this comparison in `findBoundingOrders` to make its output easier to understand (if it's considered as a standalone function).

src/data_api/orders.se
```diff
@@ -155,164 +155,189 @@ def getWorstBidOrderID(market: address, outcome: uint256):
...
-# @private
-def updateBestBidOrder(orderID: address, market: address, fxpPrice, outcome: uint256, betterOrderID: uint256):
-    if(msg.sender != self):
-        ~invalid()
-    bestBidOrderID = self.bestOrder[market][outcome].bid
-    # If best bid is not set, this order is the new best bid
-    if(bestBidOrderID == 0):
-        self.bestOrder[market][outcome].bid = orderID
+# @public
+def isBetterPrice(type: uint256, fxpPrice, orderID: address):
```
tinybike: All unlabeled parameters are `int256` by default (which is why none of my code explicitly labels parameters as `int256`).  `fxpPrice` can be negative in scalar markets, which is why it is int256 rather than uint256.

src/data_api/orders.se
```diff
@@ -155,105 +155,79 @@ def getWorstBidOrderID(market: address, outcome: uint256):
...
-        #  - betterOrderID must be specified
-        #  - price must be less than or equal to betterOrderID's price
-        else:
-            if(betterOrderID == 0):
-                ~invalid()
-            if(fxpPrice > self.orders[betterOrderID].fxpPrice):
-                ~invalid()
-    return(1: uint256)
+# @public
+def isBetterPrice(type: uint256, fxpPrice, orderID: address):
```
tinybike: The convention I've been using is methods that are just getters (i.e., that don't modify any contract data) don't need a mutex and therefore also don't use the `public` prefix.

PR #205, opened 2017-06-27
Fix to validate that the share token is a share on the current market's branch

adrake33:
﻿# Pull Request Template
### Overview
Added check to validate that the share token is a share on the current market's branch

Added new functions/data members to branch.se, market.se, & shareToken.se for checking if the denomination token is from the current market's branch.  Removed code from completeSets.se

### Testing Instructions
Run test_market.py to test.

### Examples (if necessary)

### DIFF
--- a/src/reporting/branch.se
+++ b/src/reporting/branch.se
@@ -139,3 +139,14 @@ def isContainerForReportingToken(shadyReportingToken: address):
 def isParentOf(shadyChild: address):
     branchId = shadyChild.getParentPayoutDistributionHash()
     return(self.childBranches[branchId] == shadyChild)
+
+def isContainerForShareToken(denominationToken):
+    denominationTokenMarket = denominationToken.getMarket()
+    if (not denominationTokenMarket):
+        return(0)
+    denominationTokenBranch = denominationTokenMarket.getBranch()
+    if (not denominationTokenBranch):
+        return(0)
+    if (not denominationToken.isShareToken()):
+        return(0)
+    return(1)
--- a/src/reporting/branch.se
+++ b/src/reporting/branch.se
@@ -144,8 +144,7 @@ def isContainerForShareToken(denominationToken):
     denominationTokenMarket = denominationToken.getMarket()
     if (not denominationTokenMarket):
         return(0)
-    denominationTokenBranch = denominationTokenMarket.getBranch()
-    if (not denominationTokenBranch):
+    if (not denominationTokenMarket.getBranch()):
         return(0)
     if (not denominationToken.isShareToken()):
         return(0)
--- a/src/reporting/branch.se
+++ b/src/reporting/branch.se
@@ -140,12 +140,13 @@ def isParentOf(shadyChild: address):
     branchId = shadyChild.getParentPayoutDistributionHash()
     return(self.childBranches[branchId] == shadyChild)
 
-def isContainerForShareToken(shadyShareToken):
+def isContainerForShareToken(shadyShareToken: address):
     shadyMarket = shadyShareToken.getMarket()
     if not shadyMarket:
         return(0)
     if not self.isContainerForMarket(shadyMarket):
         return(0)
-    if not shadyMarket.isContainerForShareToken(shadyToken):
+    legitMarket = shadyMarket
+    if not legitMarket.isContainerForShareToken(shadyToken):
         return(0)
     return(1)
--- a/src/reporting/branch.se
+++ b/src/reporting/branch.se
@@ -140,12 +140,12 @@ def isParentOf(shadyChild: address):
     branchId = shadyChild.getParentPayoutDistributionHash()
     return(self.childBranches[branchId] == shadyChild)
 
-def isContainerForShareToken(denominationToken):
-    denominationTokenMarket = denominationToken.getMarket()
-    if (not denominationTokenMarket):
+def isContainerForShareToken(shadyShareToken):
+    shadyMarket = shadyShareToken.getMarket()
+    if not shadyMarket:
         return(0)
-    if (not denominationTokenMarket.getBranch()):
+    if not self.isContainerForMarket(shadyMarket):
         return(0)
-    if (not denominationToken.isShareToken()):
+    if not shadyMarket.isContainerForShareToken(shadyToken):
         return(0)
     return(1)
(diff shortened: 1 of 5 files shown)


tests/reporting/test_market.py
```diff
@@ -14,6 +14,14 @@ def test_market_creation():
     shadyReportingToken = fixture.upload('../src/reporting/reportingToken.se', 'shadyReportingToken')
     shadyReportingToken.initialize(market.address, [0,2])
 
+    shadyBranch = fixture.createBranch(0, 0)
+    regularMarket = fixture.createReasonableBinaryMarket(branch, cash)
+    shadyDenominationToken = fixture.applySignature('shareToken', regularMarket.getShareToken(0))
+    try:
+        raise Exception(fixture.createReasonableBinaryMarket(shadyBranch, shadyDenominationToken))
+    except Exception as exc:
+        assert(isinstance(exc, ethereum.tester.TransactionFailed)), "Shady share token should be failing due to being in a separate branch"
```
MicahZoltu: Instead of doing `try...except` you can do, which is notably simpler:
```
    fixture.createReasonableBinaryMarket(...)
```

If this file doesn't import `raises` you can check out trading/test_wcl.py for an example and the import statements.

src/reporting/branch.se
```diff
@@ -139,3 +139,13 @@ def isContainerForReportingToken(shadyReportingToken: address):
 def isParentOf(shadyChild: address):
     branchId = shadyChild.getParentPayoutDistributionHash()
     return(self.childBranches[branchId] == shadyChild)
+
+def isContainerForShareToken(denominationToken):
```
MicahZoltu: Parameter name should be `shadyShareToken`, since that is what we assume it will be in this context (shareToken).  The `shady` prefix is the way help make it clear to readers that the input is untrusted.

src/reporting/branch.se
```diff
@@ -139,3 +139,13 @@ def isContainerForReportingToken(shadyReportingToken: address):
...
+    denominationTokenMarket = denominationToken.getMarket()
+    if (not denominationTokenMarket):
+        return(0)
+    if (not denominationTokenMarket.getBranch()):
+        return(0)
+    if (not denominationToken.isShareToken()):
+        return(0)
```
MicahZoltu: This function needs to _trustlessly_ check to see if the denomination token is a shareToken.  At the moment, this code trusts what the `denominationToken` claims, though it could be malicious.

The way to do this is to ask the denomination token what "path" it claims to be at relative to the branch and then walk from the branch down that path and see if the provided path lands at `denominationToken`.

In this case, since we already have a function for `isContainerForMarket` you can just do something like:
```
shadyMarket = denominationToken.getMarket()
if not shadyMarket:
    return(0)
if not self.isContainerForMarket(shadyMarket):
    return(0)
if not market.isContainerForShareToken(shadyToken):
    return(0)
return(1)
```

You will need to add a function to `market` for `isContainerForShareToken(shadyShareToken)`.  This function would look something like:
```
outcome = shadyShareToken.getOutcome()
if self.getShareToken(outcome) == shadyShareToken:
    return(1)
else
    return(0)
```

src/reporting/market.se
```diff
@@ -88,14 +88,27 @@ def initialize(reportingWindow: address, endTime: int256, numOutcomes: int256, p
...
+        self.shareTokens[i].initialize(self, i)
         i += 1
     self.approveSpenders()
+
+    if self.shouldCollectReportingFees(denominationToken):
+        reportingWindow = market.getReportingWindow()
+        reportingFeeRate = MARKET_FEE_CALCULATOR.getReportingFeeInAttoethPerEth(reportingWindow)
+        reportingFee = safeFxpMul(safeFxpMul(reportingFeeRate, fxpAmount), cumulativeScale)
+    else:
+        require(self.controller.lookup('cash') == denominationToken)
```
MicahZoltu: note: this require should stay in market initialization.  Something like:
```
require(self.getBranch().isContainerForShareToken(denominationToken) or self.controller.lookup('cash') == denominationToken)

File: src/reporting/branch.se [hunk 5]
```diff
@@ -139,3 +139,13 @@ def isContainerForReportingToken(shadyReportingToken: address):
...
     branchId = shadyChild.getParentPayoutDistributionHash()
     return(self.childBranches[branchId] == shadyChild)
+
+def isContainerForShareToken(shadyShareToken):
+    shadyMarket = shadyShareToken.getMarket()
+    if not shadyMarket:
+        return(0)
+    if not self.isContainerForMarket(shadyMarket):
+        return(0)
+    if not shadyMarket.isContainerForShareToken(shadyToken):
```
MicahZoltu: For clarity, there may be value in assigning to a local variable named `legitMarket`.  This is what I believe is done in the others, and it helps the reader understand that "at this point in the code, we now trust this market".
```
if not self.isContainerForMarket(shadyMarket):
    return(0)
legitMarket = shadyMarket
if not legitMarket.isContainerForShareToken(shadyToken):
```

Perhaps a better name would be `trustedMarket`... I kind of wish I had used that in similar functions.

src/reporting/market.se
```diff
@@ -88,14 +88,27 @@ def initialize(reportingWindow: address, endTime: int256, numOutcomes: int256, p
     i = 0
     while(i < numOutcomes):
         self.shareTokens[i] = SHARE_TOKEN_FACTORY.createShareToken(self.controller)
+        self.shareTokens[i].initialize(self, i)
         i += 1
     self.approveSpenders()
+    require(self.getBranch().isContainerForShareToken(denominationToken) or self.controller.lookup('cash') == denominationToken)
```
MicahZoltu: I wonder if we should reverse the order here as an optimization?  In _most_ cases, the token will be `cash` and by having the comparison against cash be first we'll avoid the share token checks in those cases.

src/reporting/market.se
```diff
@@ -88,14 +88,27 @@ def initialize(reportingWindow: address, endTime: int256, numOutcomes: int256, p
...
+
     # TODO: log short description
     # TODO: log long description
     # TODO: log tags (up to 0-2)
     # TODO: log outcome labels (same number as numOutcomes)
     # TODO: log type (scalar, binary, categorical)
 
+def shouldCollectReportingFees(denominationToken):
```
MicahZoltu: This should not take in a `denominationToken`, instead it should use `self.denominationToken`.

src/reporting/branch.se
adrake33: Ok, I'll use legitMarket, since that's the convention used elsewhere.

src/reporting/market.se
adrake33: Sure, that sounds good.  I'll swap the order of these.

PR #209, opened 2017-07-03
Bond payout fixes

adrake33:
﻿# Pull Request Template
### Overview
These changes are to ensure that bond holders get paid before transferring REP to the winning  reporting token.

### Notes
Please be aware that these changes are not ready to be merged yet, as I am still writing the test cases.  However, I am creating this PR so I can get early feedback on what I've written so far. 

### Testing Instructions
TBD

### Examples (if necessary)


src/reporting/disputeBondToken.se
```diff
@@ -0,0 +1,90 @@
...
+    require(msg.sender == self.bondHolder)
+    require(self.market.isFinalized())
+    require(payoutDistributionHash == self.payoutDistributionHash)
+    if (self.bondRemainingToBePaidOut > 0):
+        amountToTransfer = min(amountToWithdraw, self.bondRemainingToBePaidOut)
+        self.market.getReputationToken().transfer(self.bondHolder, amountToTransfer)
+        self.bondRemainingToBePaidOut = safeSub(self.bondRemainingToBePaidOut, amountToTransfer)
+
+####
+#### Getters
```
MicahZoltu: Should have a getter for `market`, even if we don't need it in the contracts.  The UI may have the address of a bond token and need to figure out what market that bond token is for.

src/reporting/disputeBondToken.se
```diff
@@ -0,0 +1,90 @@
...
+    self.market = market
+    self.bondHolder = bondHolder
+    self.payoutDistributionHash = payoutDistributionHash
+    self.bondRemainingToBePaidOut = bondAmount * 2
+    self.market.getReputationToken().transferFrom(msg.sender, self, bondAmount)
+
+def withdraw(amountToWithdraw: int256, payoutDistributionHash: int256):
+    require(msg.sender == self.bondHolder)
+    require(self.market.isFinalized())
+    require(payoutDistributionHash == self.payoutDistributionHash)
```
MicahZoltu: This should not take in a `payoutDistributionHash`.  It should instead get it from `market.getFinalPayoutDistributionHash()`.  As currently written, we are trusting the caller to give us the correct final payout distribution hash, which they can easily lie about.

src/reporting/disputeBondToken.se
```diff
@@ -0,0 +1,90 @@
...
+    if (self.bondRemainingToBePaidOut > 0):
```
MicahZoltu: This condition _shouldn't_ be necessary.  If the caller gets this far, any REP in this contract is theirs.  All we need to do is make sure that we track how much they have withdrawn (which you already do a few lines down) so that other parts of the code don't send too much REP to this contract.

`amountToTransfer` can just be set to `reputationToken.balanceOf(self)`.

src/reporting/disputeBondToken.se
```diff
@@ -0,0 +1,90 @@
...
+    assertNoValue()
+
+def initialize(market: address, bondHolder: address, bondAmount: int256, payoutDistributionHash: int256):
+    require(not self.initialized)
+    self.initialized = 1
+    self.market = market
+    self.bondHolder = bondHolder
+    self.payoutDistributionHash = payoutDistributionHash
+    self.bondRemainingToBePaidOut = bondAmount * 2
+    self.market.getReputationToken().transferFrom(msg.sender, self, bondAmount)
```
MicahZoltu: I believe we will need to call `reputationToken.approve(market, 2**254)` in initialize so the market can withdraw REP from this contract in the case where the dispute was incorrect.

src/reporting/market.se
```diff
@@ -44,10 +45,13 @@ data tentativeWinningPayoutDistributionHash
 data finalPayoutDistributionHash
 data automatedReporterPayoutDistributionHash
 data automatedReporterDisputeBondHolder
+data automatedReporterDisputeBondToken
```
MicahZoltu: We can delete all of these other dispute related variables as well.  Anyone code that accesses them should be updated to get the data from the dispute bond token (though I don't know that market needs any of that anymore).

src/reporting/disputeBondToken.se
MicahZoltu: We'll probably also need a getter for `self.payoutDistributionHash`

src/reporting/market.se
```diff
@@ -140,16 +144,16 @@ def disputeAutomatedReport():
     # intentionally does not migrate the market as automated report markets won't actually migrate unless a dispute bond has been placed or the automated report doesn't occur
     require(not self.isFinalized())
     require(self.isInAutomatedDisputePhase())
-    self.getReputationToken().transferFrom(msg.sender, self, AUTOMATED_REPORTER_DISPUTE_BOND_AMOUNT)
-    self.automatedReporterDisputeBondHolder = msg.sender
+    if (not self.automatedReporterDisputeBondToken):
```
MicahZoltu: This if should instead be `require(not self.automatedReporterDisputeBondToken)`.  None of the code in this function should execute if there is already a disputer.

The same goes for the other two similar functions.

src/reporting/market.se
```diff
@@ -211,7 +215,6 @@ def tryFinalizeAutomatedReport():
         return(0)
     self.finalPayoutDistributionHash = self.automatedReporterPayoutDistributionHash
     self.finalizationTime = block.timestamp
-    self.privateRewardBondHolders()
     self.reportingWindow.updateMarketPhase()
```
MicahZoltu: While it is correct that we don't need to reward bond holders here, we _do_ need to transfer their REP to the winning reporting token _IF_ the dispute was incorrect.

Something along the lines of this:
```
if finalPayoutDistributionHash == bondToken.getPayoutDistributionHash():
    amount = reputationToken.balanceOf(bondToken)
    reputationToken.transferFrom(bondToken, self.getFinalWinningReportingToken(), amount)
```

Same for the other 3 similar functions.

src/reporting/market.se
```diff
@@ -580,39 +573,6 @@ def getAutomatedReportDisputeDueTimestamp():
...
-    # allReportersBondPayout = availableRep - automatedReporterBondPayout - limitedReportersBondPayout
-    # if (automatedReporterBondPayout):
-    #     self.getReputationToken().transfer(self.automatedReporterDisputeBondHolder, automatedReporterBondPayout)
-    # if (limitedReportersBondPayout):
-    #     self.getReputationToken().transfer(self.limitedReportersDisputeBondHolder, limitedReportersBondPayout)
-    # if (allReportersBondPayout):
-    #     self.getReputationToken().transfer(self.allReportersDisputeBondHolder, allReportersBondPayout)
+def receiveToken():
+    if (msg.sender == self.getReputationToken())
+        #treat it like a dispute bond receipt.
```
MicahZoltu: This won't compile as is.  Recommend deleting for now until we decide whether or not to move forward with the whole "receiveToken" pattern or not.

src/reporting/reportingToken.se
```diff
@@ -77,6 +77,19 @@ def migrateLosingTokens():
     require(self.market.getFinalWinningReportingToken() != self)
     self.supply = 0
     reputationSupply = self.getReputationToken().balanceOf(self)
+    reputationToken = self.market.getReputationToken()
+    disputeBondToken = self.market.getAutomatedReporterDisputeBondToken()
+    amount = min(disputeBondToken.getBondRemainingToBePaidOut(), reputationToken.balanceOf(self))
```
MicahZoltu: This needs to include REP held by the disputeBondToken but not yet paid out.
```
amountNeeded = disputeBondToken.getBondRemainingToBePaidOut() - reputationToken.balanceOf(disputeBondToken)
amount = min(amountNeeded, reputationToken.balanceOf(self))
```

Same for following few blocks.

src/reporting/branch.se
```diff
@@ -139,17 +148,15 @@ def isContainerForReportingToken(shadyReportingToken: address):
     legitMarket = shadyMarket
     return(legitMarket.isContainerForReportingToken(shadyReportingToken))
 
-def isParentOf(shadyChild: address):
-    branchId = shadyChild.getParentPayoutDistributionHash()
-    return(self.childBranches[branchId] == shadyChild)
```
adrake33: Hmm, I'm not sure why that's showing as having been deleted.  It currently exists in the bond-payment branch: https://github.com/AugurProject/augur-core/blob/bond-payout/src/reporting/branch.se

src/reporting/disputeBondToken.se
adrake33: Good catch.  I'll update it to get it from market.

src/reporting/disputeBondToken.se
adrake33: Ok, so I should just get rid of the amountToWithdraw parameter and change amountToTransfer:

amountToTransfer = self.market.getReputationToken().balanceOf(self)

Correct?

src/reporting/disputeBondToken.se
adrake33: Ok, so we'll need a separate getter for each type of bond in the market?

src/reporting/branch.se
MicahZoltu: Hmm, maybe merge in/rebase on latest?

src/reporting/disputeBondToken.se
adrake33: Never mind, I was thinking this was in market.se, not disputeBondToken.se.

src/reporting/branch.se
adrake33: Yeah, it may have had something to do with the rebase, but it seems correct now.

src/reporting/disputeBondToken.se
adrake33: So the approve() function just allows REP to potentially be transferred to a particular contract at some point in the future?

src/reporting/market.se
adrake33: So in tryFinalizeFork(), should I have all of the if statements used in tryFinalizeAutomatedReport(), tryFinalizeLimitedReporting(), and tryFinalizeAllReporting()?

src/reporting/disputeBondToken.se
MicahZoltu: Approve is a way of address A telling token T that address B is allowed to transfer token Ts on A's behalf.

In this context, the `disputeBondToken` is telling `reputationToken` that `market` is allowed to transfer `reputationToken`s on behalf of `disputeBondToken`.

src/reporting/disputeBondToken.se
MicahZoltu: That sounds correct, yes.

src/reporting/market.se
MicahZoltu: Oh, sorry.  This could should probably be extracted into a helper function that just looks at the final payout distribution, compares it to each of the dispute bonds, and sends REP to each dispute bonds and if they have a matching payout distribution then send their REP to the winning reporting token.

No matter how the event finalized, _all_ bonds should be evaluated against the _final_ outcome.

src/reporting/disputeBondToken.se
adrake33: Ok, and 2**254 is the exact amount that token B can transfer (i.e., it's not a maximum that's being set)?

src/reporting/market.se
adrake33: I'm not sure I understand this part.  Are you saying get rid of automatedReporterDisputeBondToken, limitedReportersDisputeBondToken and allReportersDisputeBondToken?  If so, where would I have the factory create them?

src/reporting/disputeBondToken.se
MicahZoltu: 2**254 is just about the largest number Ethereum can handle, and since we explicitly trust the market, its simplest to set this as high as possible.  Each time market transfers funds on behalf of disputeBond, the amount being transferred is deducted from that total and we _really_ don't want it to run out.

src/reporting/market.se
MicahZoltu: No, I'm saying keep `automatedReporterDisputeBondToken` and get rid of `automatedReporterPayoutDistributionHash` and `automatedReporterDisputeBondHolder`.  Repeat for the other bond tokens.  That data is now stored inside the bond token, so we only need the bond token address to have access to that data.

adrake33: Ah, ok.  I'll remove these.

joeykrug:
Lemme know when ready for review again and I'll give it a pass too

MicahZoltu:
Closing in favor of #247

PR #210, opened 2017-07-04
Modify FIXMEs to note why the bare subtraction operation is safe

nuevoalex:
The worst case is even further from overflow than I said before. The difference between `minDisplayPrice` and `maxDisplayPrice` is further restricted implicitly by `reportingWindow.se:createNewMarket` in the `payoutDenominator` requirement to be inclusively between 0 and 2**254.

Since the max range is so restricted already additional tests seem like overkill IMO.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,9 +62,6 @@ target/
 *~
 \#*\#
 
-#vscode
-.vscode/
-
 bower_components
 serpent/ui
 .gitstats
@@ -75,7 +72,6 @@ python
 Downloads
 src-extern
 notes
-.pylintrc
 
 local/*.py
 *build-*.json
--- a/.gitignore
+++ b/.gitignore
@@ -62,6 +62,9 @@ target/
 *~
 \#*\#
 
+#vscode
+.vscode/
+
 bower_components
 serpent/ui
 .gitstats
@@ -72,6 +75,7 @@ python
 Downloads
 src-extern
 notes
+.pylintrc
 
 local/*.py
 *build-*.json
--- a/src/trading/orders.se
+++ b/src/trading/orders.se
@@ -272,10 +272,10 @@ def fillOrder(orderId: address, orderType: int256, market: address, orderOutcome
     require(tx.gasprice <= self.orders[market][orderOutcome][orderType][orderId].gasPrice)
     fill = 0
     if orderType == BID:
-        # FIXME: we can't use safeSub here because we might be subtracting a negative, but we need to protect against overflows
+        # We can't use safeSub here because it disallows subtracting negative numbers. Worst case here is an operation of 2**254 - 1 which won't overflow
         fill = sharesFilled + safeFxpDiv(tokensFilled, self.orders[market][orderOutcome][orderType][orderId].fxpPrice - market.getMinDisplayPrice())
     if orderType == ASK:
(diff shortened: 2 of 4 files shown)


src/trading/orders.se
```diff
@@ -272,10 +272,10 @@ def fillOrder(orderId: address, orderType: int256, market: address, orderOutcome
     require(tx.gasprice <= self.orders[market][orderOutcome][orderType][orderId].gasPrice)
     fill = 0
     if orderType == BID:
-        # FIXME: we can't use safeSub here because we might be subtracting a negative, but we need to protect against overflows
+        # We can't use safeSub here because it disallows subtracting negative numbers. Worst case here is an operation of 2**254 - 1 which won't overflow
         fill = sharesFilled + safeFxpDiv(tokensFilled, self.orders[market][orderOutcome][orderType][orderId].fxpPrice - market.getMinDisplayPrice())
     if orderType == ASK:
-        # FIXME: we can't use safeSub here because we might be subtracting a negative, but we need to protect against overflows
+        # We can't use safeSub here because it disallows subtracting negative numbers. Worst case here is an operation of 2**254 - 1 which won't overflow
```
MicahZoltu: Consider mentioning where the constraints are coming from in the comment.  I believe fxpPrice is constrained at storage write time (`saveOrder`) to `market.minValue` and `market.maxValue` and mention that `market.minValue + market.maxValue` is constrained at market creation time to `2**254`.

Normally I'm fairly against comments in code, preferring to try and improve code readability via verbose variable and function names but in this particular case I think there is value to auditors to make the assumptions very clear as well as where those assumptions are sourced from.

Alternatively, we could just add some `require(...)` statements at the top of this function that validate the above assertions, which would probably be the more correct thing to do.

```
require(self.orders[market][orderOutcome][orderType][orderId].fxpPrice <= market.getMaxDisplayPrice())
require(self.orders[market][orderOutcome][orderType][orderId].fxpPrice >= market.getMinDisplayPrice())
require(market.getMaxDisplayPrice() + market.getMinDisplayPrice < 2**254)
```

The purpose of these requires would not be because we believe there is an attack vector, but to ensure that future changes to the other parts of the code won't introduce a an attack vector into this code.

.gitignore
```diff
@@ -62,6 +62,9 @@ target/
 *~
 \#*\#
 
+#vscode
+.vscode/
```
MicahZoltu: I believe `.vscode` is intended to be committed to source control.  I currently don't use it as I'm waiting for serpent support (someone is working on it), but once we do we will want this directory committed I think.

.gitignore
```diff
@@ -72,6 +75,7 @@ python
 Downloads
 src-extern
 notes
+.pylintrc
```
MicahZoltu: I'm not terribly familiar with this file, but it seems like something that should be in source control?

nuevoalex: It's a python lint configuration settings file. I can actually move this down so it's out of the project entirely and can be used / configured on an individual basis.

.gitignore
nuevoalex: makes sense! I'll remove this

src/trading/orders.se
nuevoalex: I'll modify the comments to be more explicit.

I also like the idea of adding those requires so I'll do that as well.

PR #213, opened 2017-07-06
Fixed FIXMEs in src/reporting/reportingToken.se.

janani-sridhar:
﻿# Pull Request Template
### Overview
Fixed the FIXME's in lines 59 and 84 of src/reporting/reportingToken.se.

Added require statements to ensure that the one calling the function is the reporter, not any other malicious individual.

### Notes
Is this too simple a fix? Are there other things that I should have done?

### Testing Instructions
N/A

### Examples (if necessary)
N/A

### DIFF
--- a/src/reporting/reportingToken.se
+++ b/src/reporting/reportingToken.se
@@ -56,7 +56,7 @@ def redeemDisavowedTokens(reporter: address):
 
 # NOTE: UI should warn users about calling this before first calling `migrateLosingTokens` on all losing tokens with non-dust contents
 def redeemForkedTokens(reporter: address):
-    # FIXME: this needs to be secured (msg.sender only?) since calling this before liquidation is an attack
+    require(msg.sender == reporter)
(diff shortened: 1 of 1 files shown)


joeykrug:
In that case would the rep go to the reporter or would it go to the approved spender that's calling? It'd also need to check that the rep being sent was <= approval amount. I think the msg.sender thing is sufficient

MicahZoltu:
Good question.  I assumed it would go to the reporter, since this isn't a transfer function, it is a migrate function which means REP wouldn't be changing hands here, just migrating forward.  The value of allowing approvers is to enhance support for smart wallets and such where the address holding the funds may not be the address that is regularly interacting with the UI.  Its a minor concern, and I agree that just `require(msg.sender == reporter)` is fine for now.

MicahZoltu:
@janani-sridhar Feel free to merge anytime after the PR has been approved (which this one has been).

PR #216, opened 2017-07-07
Sorted linked list develop

nuevoalex:
First commit of the sortedLinkedList implementation and tests.

I wrote it initially in a more generic sense but found no existing examples of a truly generic interface as the function signatures we provide in externs for other libraries end up just specifying int256 params and return values.

If there is a way to make this actually support generic value types (and a generic comparison provider interface) I would like to do that, but I'm ignorant of how (or if it's even possible in Serpent).


MicahZoltu:
In Serpent everything is int256.  Things like `address` are really just `int256` in disguise.  Because of this, you can make the signature for `compare` like `def compare(first: int256, second: int256)` and it will work if the item being compared is an address.

nuevoalex:
Perfect, that's how it's currently written :]

joeykrug:
You can also just do no :type thing and it'll be int256
		

	





On Thu, Jul 6, 2017 at 7:17 PM -0700, "Alex" <notifications@github.com> wrote:












Perfect, that's how it's currently written :]

src/libraries/sortedLinkedList.se
```diff
@@ -0,0 +1,164 @@
...
+    if (hasNext):
+        if (hasPrev):
+            self.collection[self.collection[item].next].prev = self.getPrev(item)
+        else:
+            self.tail = self.getNext(item)
+
+    self.collection[item].exists = 0
+
+    self.count -= 1
+    self.assertInvariants()
```
MicahZoltu: I'm a fan of this.

tests/test_sortedLinkedList.py
```diff
@@ -0,0 +1,136 @@
+#!/usr/bin/env python
+
+from ContractsFixture import ContractsFixture
+from ethereum.tester import TransactionFailed
+from pytest import raises
+
+def test_sortedLinkedList():
+    fixture = ContractsFixture()
```
MicahZoltu: This class has been deleted on develop in the most recent merge, you'll need to switch over to the dependency injection pattern.

tests/test_sortedLinkedList.py
```diff
@@ -0,0 +1,136 @@
+#!/usr/bin/env python
+
+from ContractsFixture import ContractsFixture
+from ethereum.tester import TransactionFailed
+from pytest import raises
+
+def test_sortedLinkedList():
```
MicahZoltu: Recommend breaking this up into a large number of smaller tests rather than one monolithic test.  Such a change would allow each test to be executed in isolation (useful for debugging), as well as make the test easier to grok since we each test will start with a clean slate.

In order to avoid the performance hit of this, you can create a test fixture that uploads the contract and then takes a snapshot, then restores the snapshot for each test:

```
@fixture(scope = 'session')
def mySnapshot(contractsFixture):
    contractsFixture.upload('../src/libraries/sortedLinkedList.se')
    contractsFixture.state.mine(1)
    return contractsFixture.state.snapshot()

@fixture
def myImprovedContractsFixture(contractsFixture, mySnapshot):
    contractsFixture.state.revert(mySnapshot)
    return contractsFixture

def test_something(myImprovedContractsFixture):
    linkedList = myImprovedContractsFixture.contracts['sortedLinkedList']
    ...
```

src/libraries/sortedLinkedList.se
```diff
@@ -0,0 +1,164 @@
...
+
+def initialize(comparor: address):
+    require(not self.initialized)
+    self.initialized = 1
+    self.comparor = comparor
+    return(1)
+
+def add(item: int256, hints: arr):
+    if (self.contains(item)):
+        return(0)
```
nuevoalex: I'll go with deletion and re-adding

src/libraries/sortedLinkedList.se
```diff
@@ -0,0 +1,164 @@
...
+        self.tail = item
+        self.collection[item].exists = 1
+        self.count += 1
+        return(1)
+
+    # By default we begin the insertion search at the head
+    nodeValue = self.head
+
+    # If hints are provided (and valid) we will use the first still present value to begin the search
+    numHints = len(hints)
```
nuevoalex: Will do

src/libraries/sortedLinkedList.se
```diff
@@ -0,0 +1,164 @@
...
+    hintIndex = 0
+    while(hintIndex < numHints and node == self.head):
+        hint = hints[hintIndex]
+        if (self.contains(hint) and self.comparor.compare(hint, item) == 1):
```
nuevoalex: I was going under the assumption that we always want the provided hint to be greater than the provided item. If we don't compare here then a bad hint could result in incorrect sorting (Unless we adjust the algorithm below to be a little more complex and traverse in either direction)

src/libraries/sortedLinkedList.se
```diff
@@ -0,0 +1,164 @@
...
+        self.count += 1
+        return(1)
+
+    # By default we begin the insertion search at the head
+    nodeValue = self.head
+
+    # If hints are provided (and valid) we will use the first still present value to begin the search
+    numHints = len(hints)
+    hintIndex = 0
+    while(hintIndex < numHints and node == self.head):
```
nuevoalex: Good catch, will fix!

src/libraries/sortedLinkedList.se
```diff
@@ -0,0 +1,164 @@
...
+        hint = hints[hintIndex]
+        if (self.contains(hint) and self.comparor.compare(hint, item) == 1):
+            nodeValue = hint
```
nuevoalex: Will fix

src/libraries/sortedLinkedList.se
```diff
@@ -0,0 +1,164 @@
...
+
+    # Do traversal to find the insertion point
+    while(self.hasPrev(nodeValue) and self.comparor.compare(nodeValue, item) == 1):
```
nuevoalex: That makes sense to me.

src/libraries/sortedLinkedList.se
```diff
@@ -0,0 +1,164 @@
...
+        return(0)
+
+    hasPrev = self.hasPrev(item)
+    hasNext = self.hasNext(item)
+
+    if (hasPrev):
+        if (hasNext):
+            self.collection[self.collection[item].prev].next = self.getNext(item)
+        else:
+            self.head = self.getPrev(item)
```
nuevoalex: I opted for not setting prev and next refs to 0 in favor of just relying on the exists value of the nodes.

As I understand it 0 is a perfectly legitimate value so setting the nodes next value to 0 might actually cause errors if that node exists (unless I'm incorrect about 0 being a legitimate value possibly?)

src/libraries/sortedLinkedList.se
```diff
@@ -0,0 +1,164 @@
...
+    if (self.count == 0):
+        ~invalid()
+
+def assertHasPrev(item: int256):
+    if (!self.hasPrev(item)):
+        ~invalid()
+
+def assertHasNext(item: int256):
+    if (!self.hasNext(item)):
+        ~invalid()
```
nuevoalex: Cool, I'll switch to those

tests/test_sortedLinkedList.py
nuevoalex: Looks good! Will do.

src/libraries/sortedLinkedList.se
```diff
@@ -0,0 +1,164 @@
...
+        hint = hints[hintIndex]
+        if (self.contains(hint) and self.comparor.compare(hint, item) == 1):
+            nodeValue = hint
+
+    # Do traversal to find the insertion point
+    while(self.hasPrev(nodeValue) and self.comparor.compare(nodeValue, item) == 1):
+        nodeValue = self.getPrev(nodeValue)
+    
+    # Actually insert the node
+    if (self.comparor.compare(nodeValue, item) == 1):
```
nuevoalex: I _think_ it works as is now but given the confusion resulting from how it's written (and the change in behavior with regard to equal members) I'll reorganize it to make it more clear.

src/libraries/sortedLinkedList.se
```diff
@@ -0,0 +1,164 @@
...
+        if (hasNext):
+            self.collection[self.collection[item].prev].next = self.getNext(item)
+        else:
+            self.head = self.getPrev(item)
+
+    if (hasNext):
+        if (hasPrev):
+            self.collection[self.collection[item].next].prev = self.getPrev(item)
+        else:
+            self.tail = self.getNext(item)
```
nuevoalex: Same as above

nuevoalex:
Review comments generally all addressed. If I should set next/prev refs to 0 as suggested I can add that in quickly.

src/libraries/sortedLinkedList.se
MicahZoltu: That is a reasonable argument about `0` maybe being valid.  On the other hand, you get a gas refund if you set something to `0` so the net cost of doing this operation could potentially be _much_ lower if you set things to `0`.

I will admit, I'm a bit conflicted... I do not like `0` being magic, but in Ethereum it _is_ magic due to the way internal state is stored.

src/libraries/sortedLinkedList.se
```diff
@@ -0,0 +1,156 @@
...
+        self.collection[item].exists = 1
+        self.count += 1
+        return(1)
+
+    # By default we begin the insertion search at the head
+    nodeValue = self.head
+
+    # If hints are provided (and valid) we will use the first still present value to begin the search
+    hintIndex = 0
+    while(hintIndex < len(hints) and nodeValue == self.head):
```
MicahZoltu: `nodeValue == self.head` is wrong, I'm not sure what your intent was but I'm pretty sure this wasn't it.  :)

nuevoalex: The intent is to break on the first valid hint. Is there a more preferable way?

MicahZoltu: Nevermind, `nodeValue == self.head` is basically filling in for the lack of a `break` statement in Serpent.

MicahZoltu: My preferred way is to extract this into a helper method and use `return` as the break:
```
def getBestHint(hints: arr, item):
    hintIndex = 0
    while hintIndex < len(hints):
        hint = hints[hintIndex]
        if self.isValidHint(hint, item):
            return hint
        hintIndex += 1
    return self.head
```
As seen in the above, I would also extract out `isValidHint` into a helper method because I don't like compound if statements:
```
def isValidHint(hint, index):
    if not self.contains(hint):
        return 0
    if not self.comparor.compare(hint, item) == 1:
        return 0
    return 1
```

All of the above is style of course, so entirely up to you.

nuevoalex: I like that better as it's much more readable. Will do here and in the future

src/libraries/sortedLinkedList.se
```diff
@@ -0,0 +1,164 @@
...
+        self.collection[item].exists = 1
+        self.count += 1
+        return(1)
+
+    # Use provided hints to get the starting node for traversal
+    nodeValue = self.getBestHint(hints, item)
+
+    # Do traversal to find the insertion point
+    while(self.hasPrev(nodeValue) and self.comparor.compare(nodeValue, item) >= 0):
+        nodeValue = self.getPrev(nodeValue)
```
MicahZoltu: Thinking out loud: I wonder if there is value in creating a function named `findNodeToInsertBefore` which would return the node that `item` should be inserted before and return `0` if `item` should be inserted at head?  I'm still trying to think of a way to make the state update below a bit cleaner.

If such a function existed and was called in place of this while loop (internally it would likely have such a while loop) then below I think we would just need to do something like:
```
# insert at head
if not nodeToInsertBefore:
    self.collection[item].next = 0
    self.collection[item].prev = self.head
    self.collection[self.getPrev(item)].next = item
    self.head = item
# insert not at head
else:
    self.collection[item].next = nodeToInsertBefore
    self.collection[item].prev = self.getPrev(nodeToInsertBefore)
    self.collection[self.getNext(item)].prev = item
    self.collection[self.getPrev(item)].next = item
```
This would remove the need for any comparor usage during the state update, and remove the nested if statements in the current code.

Another potential improvement on this would be to special case head insertion just like you special cased first item insertion:
```
if bestHint == 0: # depends on bestHint returning 0 when head insertion _is_ the best option
    # TODO: insert into head
    return(1)

# TODO: non-head, non-first insertion
```
In this case, we are basically taking out the edge cases early (head insertion and first item insertion) early on leaving the only remaining code path the one where we have a well defined "next" item and we just need to insert before that.  We could even go a step further and have a third early-exit for tail insertion (when `bestHint == self.tail`) which gives us 4 distinct and isolated scenarios.  With some extraction into functions we would end up with code that looks like:

```
if self.count == 0:
    self.insertFirstItem(item)
    return(1)
bestHint = self.getBestHint(hints, item)
if bestHint == 0:
    self.insertAtHead(item)
    return(1)
itemNext = self.findNodeToInsertBefore(item, hint)
if itemNext == self.tail:
    self.insertAtTail(item)
    return(1)
self.insertInMiddle(item, itemNext)
```

Note: The above is purely to help make the code easier to read, it shouldn't have any functional change.  The goal is to make it so the reader can read any one of the helper methods in isolation and make assertions as to whether it does what the function name says.  The reader can then read the above function and assert that "if each of the individual functions do what they say, then the overall algorithm is correct".

Note: If you decide to go down this path, any helper method that mutates state will need to be marked as privet.  There is a macro you can include that has an `assertPrivate()` or something in it.

Note 2: The above is all just style, so I leave it up to you whether to make any changes, or apply a subset of the changes suggested.

src/libraries/sortedLinkedList.se
```diff
@@ -0,0 +1,164 @@
...
+
+def getNext(item: int256):
+    require(self.contains(item))
+    require(self.hasNext(item))
+    return(self.collection[item].next)
+
+def getPrev(item: int256):
+    require(self.contains(item))
+    require(self.hasPrev(item))
+    return(self.collection[item].prev)
```
MicahZoltu: Now that we are special casing `0`, I wonder if there is value in having functions like `def tryGetPrev(item)` that return either 0 or item's prev?  If we decide those would be useful, we would need to decide if `0` should be returned if the item doesn't exist, or only if the item does exist but is at the tail.

tests/test_sortedLinkedList.py
```diff
@@ -0,0 +1,220 @@
+#!/usr/bin/env python
```
MicahZoltu: Consider moving this to `tests/libraries`.

tests/test_sortedLinkedList.py
```diff
@@ -0,0 +1,220 @@
...
+    doOrderValidation(sortedLinkedList)
+
+    assert sortedLinkedList.remove(45)
+    assert sortedLinkedList.getHead() == 43
+    assert sortedLinkedList.hasNext(43) == 0
+
+    # Validate order
+    doOrderValidation(sortedLinkedList)
+
+def test_sortedLinkedListHints(sortedLinkedListContractsFixture):
```
MicahZoltu: Recommend at least one test with multiple hints, ideally one where some hint from the middle is the one picked.

src/libraries/sortedLinkedList.se
nuevoalex: Yea I think those would be useful (no harm if not). I'm opting for returning 0 if the item is the tail and raising if the item doesn't exist since it that is an input error imo.

src/libraries/sortedLinkedList.se
nuevoalex: This seems good so I'll do these suggestions.

I'll try to make my serpent code follow this sort of pattern more in the future too.

nuevoalex:
Made the suggested adjustments. I think it's much easier to reason now for sure.

PR #218, opened 2017-07-07
Test update

janani-sridhar:
﻿# Pull Request Template
### Overview
Modify tests to align with new pyethereum testing syntax.

- `from ethereum import tester` -> `from ethereum.tools import tester as tester`
- `state = tester.state()` -> `state = tester.Chain()`
- `state.abi_contract(<code>)` -> `state.contract(<code>, language="serpent")`

### Notes
- Left all variable names the same so as to avoid confusion.
- As far as I know, Ethereum itself is having some issues, so this won't be useful for some time. 
- The function `contract` returns an ABIContract if the language name is passed in (in this case, `language="serpent"`).

### Examples (if necessary)


MicahZoltu:
Running `pytest` from the `tests` directory should pass.  According to travis, it isn't at the moment.

tests/conftest.py
```diff
@@ -1,8 +1,8 @@
 from binascii import hexlify
 from datetime import timedelta
-from ethereum import tester
+from ethereum.tools import tester as tester
```
MicahZoltu: the `as tester` suffix here is unnecessary and redundant I believe.  Same comment for all other places that have this import.

CONTRIBUTING.md
```diff
@@ -5,10 +5,10 @@ Thank you for taking the time to contribute to the Augur Project! Before you sta
 Make sure to be using Python 2.7.xx, Serpent 2.0.2, and Ethereum 1.6.1. It is recommended to clone from the master branch for both [Pyethereum](https://github.com/ethereum/pyethereum) and [Serpent](https://github.com/ethereum/serpent). Pyethereum may install Ethereum 1.4.0 -- uninstall the latter if it happens. Check your version of Ethereum by running this on terminal: `pip show ethereum | grep Version`
 Check your version of python by running this: `python --version`
  
-### Pull Requests and Issues
+###Pull Requests and Issues
 When making [pull requests](https://help.github.com/articles/creating-a-pull-request/), follow the pull request template in the root directory. When [raising issues](https://help.github.com/articles/creating-an-issue/), follow the appropriate issue template in the root directory. Make sure commit messages are informative and concise.
  
-### Communication
+###Communication
```
MicahZoltu: Why did you remove the space here?  Some markdown viewers will fail if the space is missing.

tests/conftest.py
```diff
@@ -98,7 +98,7 @@ def getAllDependencies(filePath, knownDependencies):
     ####
 
     def __init__(self):
-        self.state = tester.state()
+        self.state = tester.Chain()
```
MicahZoltu: We should rename the `state` variable to `chain`, otherwise I suspect we'll end up with quite a bit of confusion.

tests/delegation_sandbox.py
```diff
@@ -81,14 +81,14 @@ def setup(controller, key):
     file.write(delegateCode)
 
 try:
-    state = tester.state()
+    state = tester.Chain()
     state.block.number += 2000000
 
-    controller = state.abi_contract(controllerCode)
-    library = state.abi_contract(libraryCode)
+    controller = state.contract(controllerCode, language="serpent")
```
MicahZoltu: Hmm, does `language` no longer default to `serpent`?  I'm not against making this explicit, just curious.

tests/conftest.py
```diff
@@ -1,8 +1,8 @@
 from binascii import hexlify
 from datetime import timedelta
-from ethereum import tester
+from ethereum.tools import tester as tester
 from ethereum.abi import ContractTranslator
-from ethereum.tester import ABIContract
+from tester import ABIContract
```
MicahZoltu: This should use the full path rather than depending on a previous import:
```
from ethereum.tools.tester import ABIContract
```

The reason for this is so that it is easy for someone reading the code to quickly find provenance.  When it just says `from tester import ABIContract` the reader has to resolve the `tester` reference (which is not always a required step) before they can establish provenance.

Same comment for all other `from tester import ...` lines.

CONTRIBUTING.md
janani-sridhar: I don't exactly know how that happened, but it did... is there any way to get rid of this part?

tests/delegation_sandbox.py
janani-sridhar: `language` defaults to EVM in `tester.py`.

CONTRIBUTING.md
MicahZoltu: The easiest solution is to just open up the CONTRIBUTING.md file for your branch locally and add the `space` back in.  There are more complicated techniques involving git that I can work with you to apply if you prefer though.

janani-sridhar:
Made the following changes:
- `from ethereum import tester` -> `from ethereum.tools import tester`
- `from ethereum.tester import TransactionFailed` -> `from ethereum.tools.tester import TransactionFailed`
- Renamed all `state` variables to `chain`.

MicahZoltu:
Update `ethereum` to the latest version on PIP in `requirements.in` and then run `pip-compile --output-file requirements.txt requirements.in` to update `requirements.txt`.  This may be what is necessary to get Travis passing.

tests/conftest.py
```diff
@@ -131,19 +131,19 @@ def upload(self, relativeFilePath, lookupKey = None):
...
     def applySignature(self, signatureName, address):
         assert address
         translator = ContractTranslator(NewContractsFixture.signatures[signatureName])
-        contract = ABIContract(self.state, translator, address)
+        contract = ABIContract(self.chain, translator, address)
         return contract
 
     def resetSnapshot(self):
-        self.state.revert(self.snapshot)
+        self.chain.revert(self.snapshot)
```
MicahZoltu: If the tests don't pass, its possible due to this.  I believe reverting state may have changed in the latest version of pyethereum, I vaguely remember seeing something about it.

joeykrug:
Ethereum in requirements wasn't 2.0.4, fixed

```=========================================================================== ERRORS ============================================================================
_____________________________________________________ ERROR collecting libraries/test_sortedLinkedList.py _____________________________________________________
ImportError while importing test module '/home/ubuntu/workspace/tests/libraries/test_sortedLinkedList.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
libraries/test_sortedLinkedList.py:3: in <module>
    from ethereum import tester
E   ImportError: cannot import name tester
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
=================================================================== 1 error in 0.35 seconds ===================================================================
``` is where it fails for me, needs to be `from ethereum.tools import tester as t`

**Comment by joeykrug** [comment]
tester.gas_limit is no longer a thing, it's now tester.STARTGAS

File: tests/test_helpers.py [hunk 7]
```diff
@@ -2,20 +2,25 @@
...
-    with raises(TransactionFailed):
-        assertNoValue.assertNoValue(value=500*10**18, sender=tester.k2)
-    gasSpent = block.gas_used - startingGasUsed
-    balanceAfter = block.get_balance(tester.a2)
+    chain = assertNoValue[0]   
+
+    balanceBefore = chain.head_state.get_balance(tester.a2)
+    startingGasUsed = chain.head_state.gas_used 
+
+    (assertNoValue[1].assertNoValue)(sender=tester.k2)
```
joeykrug: looks to be a typo, will fix

janani-sridhar:
Modified AssertNoValue to return a tuple so that I could access the chain without creating a new one, that's why there's the indexing

joeykrug:
https://github.com/AugurProject/augur-core/pull/218/commits/0121731e9d9bbeb75f396f9853f7c9ca4776cac6 is a fix for issue I was referring to (unless I read it wrong)

Also prob. better practice to just have it be a separate param explicitly than a tuple like that

janani-sridhar:
Will keep in mind, thanks for the advice. Good night!

PR #221, opened 2017-07-12
Added return(1) statements to functions missing return statements.

janani-sridhar:
﻿# Pull Request Template
### Overview
Added return statements to functions that don't return anything.

### Notes
Accidentally cloned from `test_update` branch, didn't realize till 3/4 of the way through. 
Didn't add return statements to `init` functions, added them to `initialize` functions though.


src/reporting/reportingWindow.se
```diff
@@ -84,14 +88,15 @@ def updateMarketPhase():
     if (market.isDoneWithAllReporters()):
         self.allReporterMarkets.remove(market)
         self.limitedReporterMarkets.remove(market)
-        return
```
joeykrug: Good catch on these ones

PR #223, opened 2017-07-14
added remove to iterable map

joeykrug:
﻿# Pull Request Template
### Examples (if necessary)

### DIFF
--- a/src/libraries/iterableMap.se
+++ b/src/libraries/iterableMap.se
@@ -29,10 +29,6 @@ def addOrUpdate(key, value):
     else:
         self.update(key, value)
 
-# 5-10-0, 11-2-1, 3-4-2
-
-# [5, 11, 3]
-
 def remove(key):
     require(self.contains(key))
     keyRemovedOffset = self.itemsMap[key].offset
--- a/src/libraries/iterableMap.se
+++ b/src/libraries/iterableMap.se
@@ -31,12 +31,16 @@ def addOrUpdate(key, value):
 
 def remove(key):
     require(self.contains(key))
-    # move last item in collection to the newly opened slot
-    self.itemsArray[self.itemsMap[key].offset] = self.getByOffset(self.numberOfItems)
+    keyRemovedOffset = self.itemsMap[key].offset
+    self.itemsArray[keyRemovedOffset] = 0
     self.itemsMap[key].hasValue = 0
     self.itemsMap[key].value = 0
+    if(self.numberOfItems > 1):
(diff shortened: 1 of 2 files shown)


src/libraries/iterableMap.se
```diff
@@ -49,3 +57,6 @@ def contains(key):
 
 def count():
     return self.numberOfItems
+
+def getOffsetByKey(key):
+    return self.itemsMap[key].offset
```
MicahZoltu: Recommend not exposing this.  It is an implementation detail and shouldn't be relied on by users.

src/libraries/iterableMap.se
```diff
@@ -28,7 +29,14 @@ def addOrUpdate(key, value):
     else:
         self.update(key, value)
 
-# TODO: add support for item removal, note that this requires moving the last item in the collection to the newly opened slot so we don't end up with a sparse collection
+def remove(key):
+    require(self.contains(key))
+    # move last item in collection to the newly opened slot
+    self.itemsArray[self.itemsMap[key].offset] = self.getByOffset(self.numberOfItems)
```
MicahZoltu: What happens if item being removed is the last item?  What happens if item being removed is the only item?  I haven't thought through these, so maybe this code works with them already, but wanted to make sure someone thinks about it.

src/libraries/iterableMap.se
joeykrug: Fixing in next commit

joeykrug:
Yep, there were some I found when testing, I think they're fixed in latest commit

src/libraries/iterableMap.se
```diff
@@ -28,7 +29,20 @@ def addOrUpdate(key, value):
...
 
-# TODO: add support for item removal, note that this requires moving the last item in the collection to the newly opened slot so we don't end up with a sparse collection
+def remove(key):
+    require(self.contains(key))
+    keyRemovedOffset = self.itemsMap[key].offset
+    self.itemsArray[keyRemovedOffset] = 0
+    self.itemsMap[key].hasValue = 0
+    self.itemsMap[key].value = 0
+    self.itemsMap[key].offset = 0
+    if(self.numberOfItems > 1 and keyRemovedOffset != self.numberOfItems):
```
MicahZoltu: Won't the second half of this always be true?  Number of items will be one greater than the offset of the last item, meaning in will never equal offset of removed item.

joeykrug: Yeah good point, fixing

PR #228, opened 2017-07-18
change to Travis trusty image

beaugunderson:
they're starting to deprecate the `precise` image tomorrow; this PR also adds caching which cuts about a minute (20%) off of every build


joeykrug:
Thanks!

PR #229, opened 2017-07-18
Left-pad longToHexString

tinybike:
### Overview

Left-pads the output of the `longToHexString` method (in tests/utils.py) with zeros, to 40 characters by default since all this method appears to be used for at the moment is formatting addresses.  Optional 2nd parameter lets you pick whatever padding you want though.


joeykrug:
nvm doing now

PR #232, opened 2017-07-18
Cleans up conftest a bit.

MicahZoltu:
* Makes all imports be very specific.
* Removes `true,false = True,False (unused)
* Moves a couple global state mutations into `ContractsFixture.__init__` to make sure they are executed at an expected point in time (not arbitrary).

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,22 +3,14 @@ from datetime import timedelta
 from ethereum.tools import tester
 from ethereum.abi import ContractTranslator
 from ethereum.tools.tester import ABIContract
-from ethereum import utils as u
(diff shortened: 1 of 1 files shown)


nuevoalex:
LGTM.

Out of curiosity why do you prefer importing specific python functions rather than importing the base module and doing fully qualified calls? I've seen preferences for both so again this is just curiosity.

MicahZoltu:
For consistency.  I'm not against either style (import module and then access things via namespace) or import individual items from the module.  What I'm generally not a fan of is mixing the two styles.  I believe `from ... import ...` was the more common pattern which is what caused me to originally go with it, though I could be wrong.

PR #237, opened 2017-07-21
Initial stab at Solidity contracts.

MicahZoltu:
### Overview
This changes Controller and Cash contracts over to Solidity.  The goal was to both get the infrastructure setup to compile/run Solidity and get an idea of where the sharp edges are at (externs) so we can make a better estimate as to how long migration will take.

### Notes
You will need to install `solc` (Solidity compiler) on your host (unless you use Docker) before anything will run.  I _believe_ it needs to be in the path to work though the py-solc docs suggest that you can also provide the path to solc via environment variable if you prefer.

TODO: Cash contract isn't done being migrated, need to add MUTEX support and emergency stop support.  Both should be easy at this point.

### Testing Instructions
Currently 3 tests fail, I have yet to figure out why but I wanted to get eyes on these changes sooner rather than later, and also get some additional help with debugging the failing tests.


MicahZoltu:
We either need to get Solidity installed onto Travis CI, or change our testing strategy over to using Docker run the CI.  My preference is Docker, because I hate having to tweak CI anytime something changes (and I'm already keeping the Dockerfile up to date).

beaugunderson:
I've had good luck using docker within Travis; I can handle switching over

src/Controller.sol
```diff
@@ -0,0 +1,220 @@
...
+		_;
+	}
+
+	modifier onlyInGoodTimes
+	{
+		controller.stopInEmergency();
+		_;
+	}
+
+	modifier onlyInBadTimes
```
adrake33: Maybe we could call this onlyWhenControllerIsStopped?

src/Controller.sol
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	modifier ownerOnly
+	{
+		require(msg.sender == owner);
+		_;
+	}
+
+	modifier onlyInBadTimes
+	{
```
adrake33: And perhaps we could call this onlyWhenStopped?

MicahZoltu: The original names were:
`stopInEmergency` and `onlyInEmergency`.  I always found that `stopInEmergency` caused me to pause when reading and have to think for a moment, which is when I came up with `onlyInGoodTimes` as a replacement that was (hopefully) easier to read without breaking pace.  However, I wanted the inverse modifier to be named in a similar way, rather than having a seemingly unrelated modifier name, which is when I came up with `onlyInBadTimes`.

I'm not convinced that `onlyInGoodTimes` and `onlyInBadTimes` are great names, though I would like to come up with a naming convention that is easy to read in stride and where the two modifiers are clearly paired.

Perhaps someone knows an antonym for emergency?

I also considered `onlyWhenStarted`, but the system defaults to `started` state which makes it a bit confusing.  I also considered `onlyWhenRunning`, but that begs the question of "when what is running".

Another option that OpenZeppelin uses is `Pausable` with `onlyWhenPaused` and `onlyWhenUnpaused`.  Once again though, this doesn't imply a _system wide_ outage, it instead implies that a single contract is paused for some reason.  In this case, the state of emergency is system wide and I was hoping to come up with a good name that encapsulates that somehow (hence the usage of "times").

nuevoalex: I think `onlyInGoodTimes` and `onlyInBadTimes` is fine actually. Easy to understand.

src/Controller.sol
```diff
@@ -0,0 +1,205 @@
...
+		if (whitelist[owner])
+		{
+			whitelist[owner] = false;
+			whitelist[_newOwner] = true;
+		}
+		owner = _newOwner;
+		return true;
+	}
+
+	function switchModeSoOnlyEmergencyStopsAndEscapeHatchesCanBeUsed() public devModeOwnerOnly returns(bool)
```
janani-sridhar: `switchModeSoOnlyEmergencyStopsAndEscapeHatchesCanBeUsed` is descriptive, but quite a long name. Do you have any alternatives?

src/libraries/token/StandardToken.sol
```diff
@@ -0,0 +1,57 @@
...
+	mapping (address => mapping (address => uint256)) allowed;
+
+	/**
+	* @dev Transfer tokens from one address to another
+	* @param _from address The address which you want to send tokens from
+	* @param _to address The address which you want to transfer to
+	* @param _value uint256 the amout of tokens to be transfered
+	*/
+	function transferFrom(address _from, address _to, uint256 _value) returns (bool) {
+		var _allowance = allowed[_from][msg.sender];
```
nuevoalex: What are general opinions on var? I personally prefer being explicit in typed languages.

adrake33: Same here.

src/libraries/token/StandardToken.sol
```diff
@@ -0,0 +1,57 @@
...
+		return true;
+	}
+
+	/**
+	* @dev Function to check the amount of tokens that an owner allowed to a spender.
+	* @param _owner address The address which owns the funds.
+	* @param _spender address The address which will spend the funds.
+	* @return A uint256 specifing the amount of tokens still avaible for the spender.
+	*/
+	function allowance(address _owner, address _spender) constant returns (uint256 remaining) {
```
nuevoalex: Any reason to name the output parameter here? I don't think it does any harm but it makes me look for a variable that doesn't exist.

tests/conftest.py
```diff
@@ -182,15 +225,16 @@ def whitelistTradingContracts(self):
             self.controller.addToWhitelist(self.contracts[name].address)
 
     def initializeAllContracts(self):
-        for contractName in CONTRACTS:
+        contractsToInitialize = ['mutex','Cash','orders','completeSets','makeOrder','takeBidOrder','takeAskOrder','takeOrder','cancelOrder','trade','claimProceeds','tradingEscapeHatch','ordersFetcher']
```
nuevoalex: I'm assuming the argument for changing these to be defined here instead of at the top of the file is that they're only used once so it's arguably more readable/discover-able?

tests/utils.py
```diff
@@ -10,23 +10,14 @@ def unfix(n):
...
 def longToHexString(value, leftPad=40):
-    hexStr = hex(value)[2:-1]
-    while len(hexStr) < leftPad:
-        hexStr = '0' + hexStr
-    if(hexStr[0:2] == '0x'):
-        return hexStr
-    else:
-        return '0x' + hexStr
+    # convert the value to a hex string, strip off the `0x`, strip off any trailing `L`, pad with zeros, prefix with `0x`
+    return '0x' + hex(value)[2:].rstrip('L').zfill(leftPad)
```
nuevoalex: so pythonic 🐍

src/Controller.sol
MicahZoltu: This function will only ever be called once in the life of Augur, so its super-verbose-name doesn't bother me much (so I haven't put much thought into it).  However, if someone can offer a suggestion for something more pithy I am definitely open to it.  :)

src/libraries/token/StandardToken.sol
MicahZoltu: I'm a fan of `var` personally, but if as a team we would prefer to be explicit (which so far it sounds like we are leaning that way) I will get on board with that.

src/libraries/token/StandardToken.sol
MicahZoltu: Hmm, I didn't even realize that you could name output parameters (this code is cribbed from OpenZeppelin).  I actually like it because it allows me to describe (in code) what is being returned.  In general, when I can describe something in code instead of a comment I find that valuable.  I also suspect that this name _may_ make it into exported interface definitions, which could be valuable.

That being said, I don't feel particularly strongly since, as you hinted at, the variable will never be used so it would be easy to become out of date (just like comments).

tests/conftest.py
MicahZoltu: Yeah, that is the _exact_ argument for the move.  :)  I'm generally a fan of defining things scoped as tightly as possible.  In this case, since they are used in only one place they could be inlined (which I actually started with) but I found that having the variable name improved readability by associating a name with the collection, so I hoisted it up one scope level.

tests/utils.py
MicahZoltu: Heh, I wouldn't be against breaking this out into multiple lines.  I was iterating on it a bunch trying to fix a number of bugs and by the time I was done this is what I had and breaking it out into multiple lines would require me to come up with variable names for each step... and it was late and my brain hurt.  :)

If people feel strongly (or at all) about breaking this up I can do that.  Open to help on the intermediate variable naming front.  ;)

nuevoalex: I think this is much better actually. "Pythonic" is meant to be a commendation.

src/libraries/token/StandardToken.sol
nuevoalex: Yea, interestingly you don't even have to use the return statement if you format functions this way too:

http://solidity.readthedocs.io/en/develop/control-structures.html#output-parameters

So in this case you _could_ just have `remaining = allowed[_owner][_spender];`

MicahZoltu: Oh, very interesting... if we _did_ decide to follow that pattern I would strongly recommend we come up with a naming convention that makes it _painfully obvious_ which variables are output parameters.  I'm not outright against the pattern, but I would want to have some strict guidelines around it and (even better) a syntax highlighter that understand it and can highlight return values in bright green or magenta.  I worry that something like `remaining = allowed[...][...]` is not obvious enough that it is being returned.

joeykrug: Always use explicit returns, include them in the function header too is best imo

PR #239, opened 2017-07-21
Fix to avoid reuploading decentralized controller in tests.

nuevoalex:
Fix to avoid reuploading decentralized controller in tests. Retreiving the contract from the contractFixture for every individual test would fail as the chain is reset (I believe this was the cause of the error at least). I made it instead only upload once using a session based fixture.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -106,13 +106,13 @@ class ContractsFixture:
                 }
             },
             'settings': {
-                'remappings': [ 'ROOT=%s' % resolveRelativePath("../../augur-core/src") ]
+                'remappings': [ 'ROOT=%s' % resolveRelativePath("../src") ]
             },
             'outputSelection': {
                 '*': [ 'metadata', 'evm.bytecode', 'evm.sourceMap' ]
             }
         }
-        return compile_standard(compilerParameter, allow_paths=resolveRelativePath("../../augur-core/src"))['contracts'][filename][contractName]
+        return compile_standard(compilerParameter, allow_paths=resolveRelativePath("../src"))['contracts'][filename][contractName]
 
     @staticmethod
     def getAllDependencies(filePath, knownDependencies):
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -105,13 +105,13 @@ def compileSolidity(relativeFilePath):
                 }
             },
             'settings': {
-                'remappings': [ 'ROOT=/augur-core/src' ]
+                'remappings': [ 'ROOT=%' % resolveRelativePath("/augur-core/src") ]
```
MicahZoltu: The path provided here should not have a leading `/` I think.

MicahZoltu: I'm also quite surprised this actually works.  Shouldn't it be `resolveRelativePath('../src')`?

nuevoalex: I'll look at this and change possibly. I see why this seems odd but it does actually work though.

tests/conftest.py
```diff
@@ -105,13 +106,13 @@ def compileSolidity(relativeFilePath):
...
             'settings': {
-                'remappings': [ 'ROOT=/augur-core/src' ]
+                'remappings': [ 'ROOT=%s' % resolveRelativePath("../../augur-core/src") ]
             },
             'outputSelection': {
                 '*': [ 'metadata', 'evm.bytecode', 'evm.sourceMap' ]
             }
         }
-        return compile_standard(compilerParameter, allow_paths="/augur-core/src")['contracts'][filename][contractName]
+        return compile_standard(compilerParameter, allow_paths=resolveRelativePath("../../augur-core/src"))['contracts'][filename][contractName]
```
MicahZoltu: Remove `../augur-core`.  That will _only_ work if someone clones `augur-core` GitHub project into a folder named `augur-core`, which shouldn't be assumed.

tests/conftest.py
```diff
@@ -145,6 +146,7 @@ def __init__(self):
         config_metropolis['BLOCK_GAS_LIMIT'] = 2**60
         self.chain = tester.Chain(env=Env(config=config_metropolis))
         self.contracts = {}
+        self.originalContracts = {}
```
MicahZoltu: I think this line can be deleted.

PR #240, opened 2017-07-21
Switch travis to docker; use slim image

beaugunderson:
Some tests succeed but most currently fail with:

```
E               {"contracts":{},"errors":[{"component":"general","formattedMessage":"Cannot import url (\"/app/src/Controller.sol\"): File outside of allowed directories.","message":"Cannot import url (\"/app/src/Controller.sol\"): File outside of allowed directories.","severity":"error","type":"IOError"}],"sources":{}}
```

Does the current image have the same issue?

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,4 +2,4 @@
 **/*.pyc
 **/.cache/
 **/compilation_cache/
-src/macros/externs.sem
+.git/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,16 @@
 dist: trusty
-sudo: false
+sudo: required
 
-cache:
-  - apt
-  - pip
+language: bash
 
-language: python
+services:
+  - docker
 
 notifications:
   email: false
 
-addons:
-  apt:
-    packages:
-      - libblas-dev
-      - liblapack-dev
-      - gfortran
+before_install: ./.travis/ci-setup.sh
+script: ./.travis/ci.sh
 
-install:
-  - pip install -U pip setuptools
-  - pip install -r requirements.txt
-
-script:
-  - travis_wait 60 pytest -vv tests
+git:
+  depth: 5
--- a/Dockerfile-test
+++ b/Dockerfile-test
@@ -1,42 +1,23 @@
-FROM python:2.7.13-alpine3.6
+FROM python:2.7.13
 
-# install a nice prompt
-RUN apk --no-cache --update add bash \
-	&& rm -rf /var/cache/apk/*
-
-# install Solidity compiler (solc)
-RUN apk --no-cache --update add build-base cmake boost-dev git ca-certificates wget openssl \
-	&& update-ca-certificates \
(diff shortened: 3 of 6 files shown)


MicahZoltu:
Rebase on the recent PR merge, should resolve that.

beaugunderson:
@MicahZoltu only one failure now; I think this is ready for a merge:

https://travis-ci.org/AugurProject/augur-core/builds/256187053

.travis.yml
```diff
@@ -1,25 +1,23 @@
 dist: trusty
-sudo: false
+sudo: required
 
-cache:
-  - apt
-  - pip
+# cache:
+#   directories:
+#     - $HOME/docker
```
MicahZoltu: Recommend deleting rather than commenting out.

Dockerfile-test
```diff
@@ -1,42 +1,24 @@
...
-	&& pip install -r requirements.txt \
-	&& apk del gcc g++ openssl-dev \
-	&& rm -rf /var/cache/apk/*
 
-# install runtime dependencies
-RUN apk --no-cache --update add libstdc++ libssl1.0 \
-	&& rm -rf /var/cache/apk/*
+RUN cd /app && \
+  pip install --upgrade pip setuptools && \
+  pip install -r requirements.txt
```
MicahZoltu: I prefer tabs over spaces (I thought I had an .editorconfig to enforce this, but apparently not).

Dockerfile-test
```diff
@@ -1,42 +1,24 @@
...
-	&& wget -qO- https://github.com/ethereum/solidity/releases/download/v0.4.13/solidity_0.4.13.tar.gz | tar -xvz -C /solidity --strip 1 \
-	&& cd /solidity && cmake -DCMAKE_BUILD_TYPE=Release -DTESTS=0 -DSTATIC_LINKING=1 \
-	&& cd /solidity && make solc && install -s  solc/solc /usr/bin \
-	&& cd / && rm -rf solidity \
-	&& apk del sed build-base git make cmake gcc g++ musl-dev curl-dev boost-dev \
-	&& rm -rf /var/cache/apk/*
+FROM python:2.7.13
+
+RUN wget --quiet --output-document /usr/local/bin/solc \
+  https://github.com/ethereum/solidity/releases/download/v0.4.13/solc-static-linux && \
```
MicahZoltu: I prefer `&&` at the start of the line so it is clear to the read that this is a continuation, without them having to read to the end of the line to find out.  Helps quickly identify bugs where an `&&` is missing which can change behavior (but not fail) in some cases.

MicahZoltu: Also recommend keeping each command on one line, not split across lines.  In this case, this line is part of the previous line so there should not be a line break here.

.dockerignore
```diff
@@ -2,4 +2,4 @@
 **/*.pyc
 **/.cache/
 **/compilation_cache/
-src/macros/externs.sem
```
MicahZoltu: Ah, totally forgot to remove this when I removed the `generateExterns` call.  My bad.

beaugunderson:
@MicahZoltu 👍 made your suggested changes

PR #245, opened 2017-07-24
Migrates ReputationToken to Solidity.

MicahZoltu:
This is an important step because it shows a contract working with the delegation stuff.  It also brings to light an issue that is going to be uncomfortable during the migration, which is the need for interface definitions for any not-yet-migrated dependencies.


joeykrug:
Build failed

joeykrug:
> which is the need for interface definitions for any not-yet-migrated dependencies.

I dont follow

MicahZoltu:
See `ReputationToken.sol`, we have to define an interface for `Branch` at the top because Branch isn't migrated yet.

MicahZoltu:
Ugh.  The style checker doesn't support multi-line boolean conditionals.  :'(

MicahZoltu:
Fixing for now, but we should consider disabling that check: https://travis-ci.org/AugurProject/augur-core/builds/257026528#L581

MicahZoltu:
https://travis-ci.org/AugurProject/augur-core/builds/257026528#L583 is also a stupid rule (leads to mistakes).  By putting the operator on the left, it makes it clear to the reader at a glance that the line is a continuation.  When you put it on the line at the right, it is easy to not notice that both lines are part of the same condition because reading non-aligned right side is much harder than aligned left side.

src/reporting/ReputationToken.sol
```diff
@@ -0,0 +1,88 @@
...
+
+    string public name = "Reputation";
+    string public symbol = "REP";
+    uint256 public decimals = 18;
+    Branch public branch;
+    ReputationToken public topMigrationDestination;
+
+    function initialize(Branch _branch) public returns (bool) {
+        require(_branch != address(0));
+        require(branch == address(0));
```
joeykrug: These _branch vs branch seem kinda confusing, other than that looks good

MicahZoltu: This is fixed in my latest PR.  I also found it confusing.

PR #247, opened 2017-07-26
Bond payout 2

adrake33:
﻿# Pull Request Template
### Overview

These changes are to ensure that bond holders get paid before transferring REP to the winning reporting token.

### Notes

The changes I made to market.se have pushed it over the 4.7 gas limit.  (It now requires 5.1 gas to upload to the blockchain.)  We will need to find a way to reduce this.

### Testing Instructions

Run pytest -s reporting/test_bond_payouts.py

### Examples (if necessary)


joeykrug:
This looks like it needs merging of develop in, I think that's probably part of the gas issue is it's using old tester

joeykrug:
https://github.com/AugurProject/augur-core/blob/develop/tests/conftest.py#L181 for an example re gas amount for a tx / latest on chain tx gas limit on mainnet

adrake33:
Yes, this will need to be merged with develop.  I think I originally created bond-payout-2 based off of develop because bond-payout has having merge issues with develop.  I'll work with Micah on merging this with develop.

src/factories/disputeBondTokenFactory.se
```diff
@@ -0,0 +1,17 @@
+# Copyright (C) 2015 Forecast Foundation OU, full GPL notice in LICENSE
+
+extern delegateExtern: [setup:[int256,int256]:_]
+extern disputeBondToken: [initialize:[address,address,uint256,int256]:bool]
+
+# workaround for https://github.com/ethereum/serpent/issues/119
+data lastDisputeBondToken
+
+def createDisputeBondToken(controller: address, market: address, bondHolder: address, bondAmount: int256, payoutDistributionHash: int256):
```
nuevoalex: bondAmount I think should be uint256. Not really an attack vector just for consistency.

adrake33: Good catch!  Will update.

src/reporting/market.se
```diff
@@ -168,9 +166,9 @@ def disputeAllReporters():
     assertNoValue()
     self.migrateThroughAllForks()
     require(self.isInAllDisputePhase())
-    self.getReputationToken().trustedTransfer(msg.sender, self, ALL_REPORTERS_DISPUTE_BOND_AMOUNT)
-    self.allReportersDisputeBondHolder = msg.sender
-    self.allReportersPayoutDistributionHash = self.tentativeWinningPayoutDistributionHash
+    require(self.limitedReportersDisputeBondToken)
```
nuevoalex: Why is this require needed?

src/reporting/market.se
```diff
@@ -146,18 +144,18 @@ def disputeAutomatedReport():
...
     return 1
 
 def disputeLimitedReporters():
     assertNoValue()
     self.migrateThroughAllForks()
-    require(self.isInLimitedDisutePhase())
-    self.getReputationToken().trustedTransfer(msg.sender, self, LIMITED_REPORTERS_DISPUTE_BOND_AMOUNT)
-    self.limitedReportersDisputeBondHolder = msg.sender
-    self.limitedReportersPayoutDistributionHash = self.tentativeWinningPayoutDistributionHash
+    require(self.isInLimitedDisputePhase())
```
nuevoalex: For this and `disputeAllReporters` shouldn't we also require that there is not already a dispute bond for this phase placed as we do for the automated dispute?

src/reporting/market.se
```diff
@@ -359,6 +351,24 @@ def validatePayoutNumerators(payoutNumerators: arr):
     require(sum == self.payoutDenominator)
     return 1
 
+#CONSIDER: Do these 2 functions need any other require()s?
+def transferBondAmountToDisputeBondToken(bondHolder: address, disputeBondToken: address, bondAmount: int256):
+    reputationToken = self.getReputationToken()
```
nuevoalex: I think it might be reasonable to `assertPrivate` and require that the bondHolder is actually the bondHolder of the disputeBondToken

src/reporting/market.se
adrake33: It's redundant since this function is also requiring self.isInAllDisputePhase(), which is checking to see if self.limitedReportersDisputeBondToken is set.  I'll remove it.

src/reporting/market.se
```diff
@@ -568,66 +599,14 @@ def needsMigration():
         return(0)
     if (self.automatedReporterAddress and block.timestamp < self.getAutomatedReportDueTimestamp()):
         return(0)
-    if (self.automatedReporterPayoutDistributionHash and block.timestamp < self.getAutomatedReportDisputeDueTimestamp()):
-        return(0)
-    if (self.automatedReporterPayoutDistributionHash and not self.automatedReporterDisputeBondHolder):
+    if (self.automatedReporterDisputeBondToken and block.timestamp < self.getAutomatedReportDisputeDueTimestamp()):
```
MicahZoltu: I don't believe this change is correct.  We should _not_ migrate if there is an automated report (not dispute) and the dispute time window hasn't passed yet, which is what this check was originally doing.  There also needs to be a check like the one that was deleted:
```
if (self.automatedReportReceived and block.timestamp < self.getAutomatedReportDisputeDueTimestamp()):
    return 0
if (self.automatedReportReceived and not self.automatedReporterDisputeBondToken):
    return 0
```
The second will make it so that if there is an automated report and no dispute _after_ the dispute window then the market should be migrated.

src/reporting/reportingToken.se
```diff
@@ -81,8 +80,26 @@ def migrateLosingTokens():
...
+    reputationToken = self.getReputationToken()
+    automatedReporterDisputeBondToken = self.market.getAutomatedReporterDisputeBondToken()
+    limitedReportersDisputeBondToken = self.market.getLimitedReportersDisputeBondToken()
+    allReportersDisputeBondToken = self.market.getAllReportersDisputeBondToken()
+    if (automatedReporterDisputeBondToken and automatedReporterDisputeBondToken.getDisputedPayoutDistributionHash() != self.market.getFinalPayoutDistributionHash()):
+        amountNeeded = automatedReporterDisputeBondToken.getBondRemainingToBePaidOut() - reputationToken.balanceOf(automatedReporterDisputeBondToken)
+        amountToTransfer = min(amountNeeded, reputationToken.balanceOf(self))
+        if (amountToTransfer > 0):
+            reputationToken.transfer(automatedReporterDisputeBondToken, amountToTransfer)
+            log(type=Transfer, self, automatedReporterDisputeBondToken, amountToTransfer)
```
MicahZoltu: We aren't transferring reporting tokens here, so this log is unnecessary and should be avoided.  There may be value in logging _some_ event when this happens, but it wouldn't be the `Transfer` event.  For now, I recommend just deleting this log.

Same feedback for other log line below.

src/reporting/reportingToken.se
```diff
@@ -81,8 +80,26 @@ def migrateLosingTokens():
...
+    if (limitedReportersDisputeBondToken and limitedReportersDisputeBondToken.getDisputedPayoutDistributionHash() != self.market.getFinalPayoutDistributionHash()):
+        amountNeeded = limitedReportersDisputeBondToken.getBondRemainingToBePaidOut() - reputationToken.balanceOf(limitedReportersDisputeBondToken)
+        amountToTransfer = min(amountNeeded, reputationToken.balanceOf(self))
+        if (amountToTransfer > 0):
+            reputationToken.transfer(limitedReportersDisputeBondToken, amountToTransfer)
+            log(type=Transfer, self, limitedReportersDisputeBondToken, amountToTransfer)
+    if (reputationToken.balanceOf(self) > 0):
+        reputationToken.transfer(self.market.getFinalWinningReportingToken(), reputationToken.balanceOf(self))
```
MicahZoltu: Consider extracting out an `amountToTransfer` variable like in the above two blocks so you don't have to call `balanceOf` twice back-to-back.

src/reporting/reportingToken.se
```diff
@@ -98,24 +115,39 @@ def redeemWinningTokens(reporter: address):
...
+    assertNoValue()
+    require(self.market.isFinalized())
+    desiredBondPot = 0
+    if (self.market.getAutomatedReporterDisputeBondToken() and self.market.getAutomatedReporterDisputeBondToken().getDisputedPayoutDistributionHash() != self.market.getFinalPayoutDistributionHash()):
+        desiredBondPot += AUTOMATED_REPORTER_DISPUTE_BOND_AMOUNT * 2
+    if (self.market.getLimitedReportersDisputeBondToken() and self.market.getLimitedReportersDisputeBondToken().getDisputedPayoutDistributionHash() != self.market.getFinalPayoutDistributionHash()):
+        desiredBondPot += LIMITED_REPORTERS_DISPUTE_BOND_AMOUNT * 2
+    if (self.market.getAllReportersDisputeBondToken() and self.market.getAllReportersDisputeBondToken().getDisputedPayoutDistributionHash() != self.market.getFinalPayoutDistributionHash()):
+        desiredBondPot += ALL_REPORTERS_DISPUTE_BOND_AMOUNT * 2
+    return(desiredBondPot)
```
MicahZoltu: It feels like this method should be on `market` rather than on the reporting token.  I don't see any variables being accessed from this contract, they are all off of `market`.  Also, aren't those constants only defined over in `market.se` meaning this currently doesn't do the right thing?

src/reporting/reportingToken.se
```diff
@@ -98,24 +115,39 @@ def redeemWinningTokens(reporter: address):
...
     self.tokens[reporter] = 0
-    self.getReputationToken().transfer(reporter, reporterReputationShare)
-    log(type=Transfer, reporter, 0, attotokens)
+    if (reporterReputationShare > 0):
+        self.getReputationToken().transfer(reporter, reporterReputationShare)
+        log(type=Transfer, reporter, 0, attotokens)
     return 1
 
 # FIXME: bond holders have already been paid at this point (they are paid as soon as market finalization occurs)
 # CONSIDER: track how much bond holders are owed, then as REP is extracted from losing tokens it initially goes into
```
MicahZoltu: I believe these comments can be removed now.

src/reporting/reportingToken.se
```diff
@@ -98,24 +115,39 @@ def redeemWinningTokens(reporter: address):
...
 def extractBondHolderPayments():
```
MicahZoltu: Do we still need this function?  I believe now the mechanic is to extract REP from losing tokens during the losing token migration process, rather than at `redeemWinningToken` time so attempting to extract them again here is wasteful since it will never do anything.

This would explain why the `getDesiredDisputeRewardPot` function is not causing test failures despite the undefined constants.

tests/reporting/test_dispute_bond_token.py
```diff
@@ -0,0 +1,710 @@
...
+    # ----- End scalar market test cases -----
+])
+def test_dispute_bond_tokens(marketType, automatedReporterAccountNum, automatedReporterOutcome, automatedReporterDisputerAccountNum, automatedReporterDisputeStakes, limitedReportersDisputerAccountNum, limitedReportersDisputeStakes, allReportersDisputerAccountNum, allReportersDisputeStakes, expectedAccountBalances, contractsFixture):
```
MicahZoltu: While I really like the thoroughness of this parameterized test, I find it incredibly difficult to read and isolate errors (debug).  I don't think its necessary to change it now, but in the future consider looking for opportunities to write smaller/simpler tests.  Also, there may be value in breaking out more functions (I'm not actually sure on this) or perhaps having a few parameterized tests that are more similar to each other rather than a single parameterized test tha tests everything.

src/reporting/disputeBondToken.se
```diff
@@ -0,0 +1,125 @@
...
+inset('../macros/require.sem')
+
+data initialized
+data market
+data bondHolder
+data disputedPayoutDistributionHash
+data bondRemainingToBePaidOut
+data tokens[2**160]
+data spenders[2**160][2**160]
+data supply
```
adrake33: Does that mean we should get rid of the transferFrom() & allowance() functions as well, since they both reference spenders?

MicahZoltu: Yeah, I believe there is another comment related to this buried in this review that says about that.

tests/reporting/test_dispute_bond_token.py
adrake33: Agreed, this really ought to be broken up into smaller functions, but I was trying to wrap this up as quickly as possible since it was blocking the solidity migration.

src/reporting/reportingToken.se
adrake33: Yeah, I'll need to look this over again.  I've marked this as a TODO item.

MicahZoltu:
I'll get this rebased.

PR #248, opened 2017-07-28
Replaces legacy reputation contract with new contract.

MicahZoltu:
Retained old serpent legacyRep contract because a number of tests depend on it and I don't want to sink time into fixing them yet.

I put the code in `src/` so other contracts can reference it if necessary (e.g., for legacy rep migration process) but I have excluded it from being uploaded with everything else in the test suite.

Contracts should be byte-for-byte compatible with what is uploaded to main network (though I cleaned them up to match our team style and pass linter checks).

### DIFF
--- a/.soliumrc.json
+++ b/.soliumrc.json
@@ -8,7 +8,7 @@
     "lbrace": true,
     "mixedcase": true,
     "camelcase": true,
-    "uppercase": true,
+    "uppercase": false,
     "no-with": true,
     "no-empty-blocks": true,
     "no-unused-vars": true,
--- a/src/legacy_reputation/BasicToken.sol
+++ b/src/legacy_reputation/BasicToken.sol
@@ -1,0 +1,36 @@
+pragma solidity ^0.4.11;
+
+import 'ROOT/legacy_reputation/ERC20Basic.sol';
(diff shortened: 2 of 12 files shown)


joeykrug:
Looks like linter is getting ran on it causing build to fail

MicahZoltu:
@beaugunderson Do you happen to know if the linter supports inline ignores?  Not only do we need to ignore these errors here, but we'll need to ignore them for all of our ERC20 contracts as well.  The interface defines these as lowercase, but the linter wants all constants uppercase.

MicahZoltu:
Apparently it doesn't support it.  :'(  https://github.com/duaraghav8/Solium/issues/1

Should we disable this particular check then?

joeykrug:
Yeah let's disable this check for now

beaugunderson:
@MicahZoltu seems straightforward to add inline ignores to Solium; I'll work up a PR

MicahZoltu:
Check disabled and CI green.  Recommend we not wait for Solium to update to merge this.  :)  (though, huge thanks @beaugunderson!)

beaugunderson:
yeah definitely don't wait for that, assume we'll want to use it elsewhere though :)

PR #249, opened 2017-07-31
Map stack solidity

janani-sridhar:
﻿# Pull Request Template
### Overview
Attempted to rewrite iterableMap.se and stack.se in solidity.

### Notes
- Stacks don't seem to be used anywhere in the code so far, so I made the return types for `pop()` and `peek()` to be `uint`. 
- IterableMaps are only used in src/trading/topics.se and key/value pairs seem to be of type `int256`, so I made return types in IterableMap.sol to be of type `uint256`.


src/libraries/IterableMap.sol
```diff
@@ -0,0 +1,104 @@
...
+        uint256 offset;
+    }
+
+    address owner;
+    bool initialized;
+    uint256[] itemsArray;
+    Item[] itemsMap;
+    uint numberOfItems;
+
+    function any() {
```
adrake33: Does Solidity have any() functions that work like Serpent's any() functions (i.e., they are executed before all functions except init())?  If not, this will need to be done differently.

src/libraries/IterableMap.sol
```diff
@@ -0,0 +1,104 @@
...
+            this.itemsArray[keyRemovedOffset] = tailItemKey;
+            this.itemsMap[tailItemKey].offset = this.numberOfItems - 2;
+        }
+   
+        this.numberOfItems -= 1;
+        return (uint(1));
+    }
+
+    function getByKeyOrZero(uint256 key) returns (uint256) {
+        return this.itemsMap[key].value;
```
adrake33: What is our style convention for return statements?  I've noticed that some of them have parentheses around the returned value and some do not.  We should probably decide on a standard for this.

src/libraries/IterableMap.sol
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    function initialize(address owner) returns (uint) {
+         require(!this.initialized);
+         this.initialized = true;
+         this.owner = owner;
+         return (uint(1));
+    }
+
+    function add(uint256 key, uint256 value) returns (uint) {
```
adrake33: I think we can probably change these 'uint256' to just 'uint', since uint defaults to uint256?

src/libraries/IterableMap.sol
```diff
@@ -0,0 +1,104 @@
...
+   
+    function getByOffset(uint256 offset) returns (uint256) {
+        require(0 <= offset && offset < this.numberOfItems);
+        return this.itemsArray[offset];
+    }
+
+    function contains(uint256 key) returns (uint) {
+        if (this.itemsMap[key].hasValue) {
+            return (uint(1));
+	}
```
adrake33: This bracket should be moved to the right.

src/libraries/IterableMap.sol
```diff
@@ -0,0 +1,104 @@
...
+         require(!this.initialized);
+         this.initialized = true;
+         this.owner = owner;
+         return (uint(1));
+    }
+
+    function add(uint256 key, uint256 value) returns (uint) {
+         require(!this.contains(key));
+         this.itemsArray[this.numberOfItems] = key;
+	     this.itemsMap[key].hasValue = true;
```
adrake33: This line should be moved over a space.

src/libraries/IterableMap.sol
```diff
@@ -0,0 +1,104 @@
...
+         this.itemsMap[key].value = value;
+         return (uint(1));
+    }
+
+    function addOrUpdate(uint256 key, uint256 value) returns (uint) {
+        if (!this.contains(key)) {
+            this.add(key, value);
+        } 
+        
+        else {
```
adrake33: According to the Solidity style guide, this should be on the same line as the '}' above.

src/libraries/IterableMap.sol
```diff
@@ -0,0 +1,104 @@
...
+    function remove(uint256 key) returns (uint) {
+	    require(this.contains(key));
+        uint256 keyRemovedOffset = this.itemsMap[key].offset;
+        this.itemsArray[keyRemovedOffset] = 0;
+        this.itemsMap[key].hasValue = false;
+        this.itemsMap[key].value = 0;
+        this.itemsMap[key].offset = 0;
+        
+        if (this.numberOfItems > 1 && keyRemovedOffset != (this.numberOfItems - 1)) {
+        /* move tail item in collection to the newly opened slot from the key we just removed if not last or only item being removed */
```
adrake33: This comment should be moved over to the right.

src/libraries/Stack.sol
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    function push(var item) returns (uint) {
+         uint index = this.head + 1;
+         this.head = index;
+         this.collection[index] = item;
+         return (uint(1));
+    }    
+
+    function pop() returns (uint) {
```
adrake33: Should this be returning a uint?  Isn't removedValue set to a pointer in this case?  In push(), this.collection[index] looks like it's getting set to a pointer.

src/libraries/IterableMap.sol
joeykrug: Parentheses please

src/libraries/IterableMap.sol
MicahZoltu: There is no `any` function in Solidity, `function () { ... }` is equivalent.  However, See other comment on how this should probably be a modifier instead.

src/libraries/IterableMap.sol
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    function add(uint256 key, uint256 value) returns (uint) {
+         require(!this.contains(key));
+         this.itemsArray[this.numberOfItems] = key;
+	     this.itemsMap[key].hasValue = true;
+         this.itemsMap[key].value = value;
+         this.itemsMap[key].offset = this.numberOfItems;
+         this.numberOfItems += 1;
+         return (uint(1));
```
MicahZoltu: This function should return a `bool` (same to all of the other functions that just `return 1`).

src/libraries/IterableMap.sol
MicahZoltu: Recommend running Solium, it will call out things like this where the style should be `} else {`.  If you are using VSCode, @beaugunderson wrote an extension for it that will put red curlies in your code when you do something the linter doesn't like.

src/libraries/IterableMap.sol
```diff
@@ -0,0 +1,104 @@
+pragma solidity ^0.4.13;
+
+import 'ROOT/Controller.sol';
+
+contract IterableMap {
```
MicahZoltu: We should probably name this contract (and file) `IterableMapUint256` or something.  I suspect we will want/need other iterable maps in the future, and having them named such that we can easily spin them up with some simple copy/paste would be nice.

src/libraries/IterableMap.sol
```diff
@@ -0,0 +1,104 @@
+pragma solidity ^0.4.13;
+
+import 'ROOT/Controller.sol';
```
MicahZoltu: This import appears to be unused/unnecessary.  Recommend removing it.

src/libraries/IterableMap.sol
MicahZoltu: This is a delegated contract, which means it _must_ derive from `DelegationTarget` as its first derivation class:
```
contract IterableMap is DelegationTarget { ... }
```

src/libraries/IterableMap.sol
```diff
@@ -0,0 +1,104 @@
...
+    struct Item {
+        bool hasItem;
+        uint256 value;
+        uint256 offset;
+    }
+
+    address owner;
+    bool initialized;
+    uint256[] itemsArray;
+    Item[] itemsMap;
```
MicahZoltu: This should be a mapping rather than an array:
```
mapping(uint256 => Item) itemsMap
```

src/libraries/IterableMap.sol
```diff
@@ -0,0 +1,104 @@
...
+    uint numberOfItems;
```
MicahZoltu: All variables should have explicit visibility modifiers on them (public, private, internal, external).  For this contract, all of these could be public.

src/libraries/IterableMap.sol
```diff
@@ -0,0 +1,104 @@
...
+
+    function any() {
+         if (this.initialized) {
+             require(msg.sender == this.owner || msg.sender == this);
+         }
+    }
+
+    function initialize(address owner) returns (uint) {
```
MicahZoltu: Same comment as above with visibility modifiers, all functions should explicitly have visibility modifiers.  Any function that doesn't write to storage should be marked as `constant`, the rest should probably be `public onlyOwner` (see comment above re: modifiers).

src/libraries/IterableMap.sol
MicahZoltu: Solidity has proper arrays, which means we should no longer need `numberOfItems` as you can just do `itemsArray.length`.  Anywhere where we are adding an item and incrementing the length we can instead do:
```
itemsArray.push(newItem);
```
Anywhere where we are removing an item from the top of the array and shortening we can do:
```
delete itemsArray[itemsArray.length];
itemsArray.length -= 1;
```

src/libraries/Stack.sol
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    function initialize(address owner) returns (uint) {
+         require(!this.initialized);
+         this.initialized = true;
+         this.owner = owner;
+         return (uint(1));
+    }
+
+    function push(var item) returns (uint) {
```
MicahZoltu: Does this actually work (var parameter)?  I would be quite surprised if it did...

src/libraries/Stack.sol
```diff
@@ -0,0 +1,50 @@
+pragma solidity ^0.4.13;
+
+import 'ROOT/Controller.sol';
+
+contract Stack {
+    
+    var[] collection;
```
MicahZoltu: I don't think `var` works here... I think you will need to pick a type.

src/libraries/Stack.sol
```diff
@@ -0,0 +1,50 @@
+pragma solidity ^0.4.13;
+
+import 'ROOT/Controller.sol';
+
+contract Stack {
```
MicahZoltu: Apply all of the comments from IterableMap to this contract as well.

src/libraries/IterableMap.sol
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    function getByKey(uint256 key) returns (uint256) {
+        require(this.itemsMap[key].hasValue);
+        return (this.itemsMap[key].value);
+    }
+   
+    function getByOffset(uint256 offset) returns (uint256) {
+        require(0 <= offset && offset < this.numberOfItems);
+        return this.itemsArray[offset];
```
MicahZoltu: Don't use `this.` for accessing local state variables.  Instead just reference them directly:
```
return itemsArray[offset];
```

src/libraries/IterableMap.sol
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    function contains(uint256 key) returns (uint) {
+        if (this.itemsMap[key].hasValue) {
+            return (uint(1));
+	}
+        return (uint(0));
+    }
```
MicahZoltu: I believe this function can be simplified to:
```
return itemsMap[key].hasValue;
```

src/libraries/IterableMap.sol
```diff
@@ -0,0 +1,104 @@
...
+        else {
+            this.update(key, value);
+        }
+
+        return (uint(1));
+    }
+
+    function remove(uint256 key) returns (uint) {
+	    require(this.contains(key));
+        uint256 keyRemovedOffset = this.itemsMap[key].offset;
```
MicahZoltu: All local variables and function parameter names should be prefixed with `_` so we can tell them apart from state variables.
e.g.: `_key` and `_keyRemovedOffset`.

src/libraries/IterableMap.sol
MicahZoltu: The same goes for functions within this contract, call them without a `this.` prefix.

When you access a variable via `this` it compiles down to an external call rather than an internal call which is needlessly more expensive, and also means you can't call `private` and `internal` members (I believe).

src/libraries/StackUint256.sol
```diff
@@ -0,0 +1,43 @@
...
+
+import "ROOT/libraries/DelegationTarget.sol";
+import "ROOT/legacy_reputation/Ownable.sol";
+
+contract StackUint256 is DelegationTarget, Ownable {
+    
+    uint256[] public collection;
+    uint256 public head;
+    address public owner;
+    bool public initialized;
```
adrake33: I think we decided to make all member variables private by default, so the visibility of these will probably need to be private, and we may need getter functions for these as well.

MicahZoltu: Thinking about applying this policy to every contract, including the libraries, is quickly making me regret going along with the chosen policy... (though I recognize that the other options each had their own set of problems).

adrake33: Yeah, this also made me realize I'll need to do the same thing on Orders.sol.  There's no perfect solution, unfortunately.

src/libraries/IterableMap.sol
joeykrug: It does not, it has modifiers though which can be used to achieve similar effects (and modifiers can even take function parameters too and use them if needed)

src/libraries/IterableMap.sol
joeykrug: It does, but lets be as specific as possible

src/libraries/ComparorUint256.sol
```diff
@@ -0,0 +1,16 @@
+pragma solidity ^0.4.13;
+import "ROOT/libraries/DelegationTarget.sol";
+
+
+contract ComparorUint256 is DelegationTarget {
```
MicahZoltu: Since we no longer have generic implementations of the collections, I think we can get rid of this contract all together and just inline the checks.  The code will be much easier to read I suspect.

src/libraries/Delegator.sol
```diff
@@ -0,0 +1,52 @@
+pragma solidity ^0.4.13;
+
+import 'ROOT/Controller.sol';
+import 'ROOT/libraries/DelegationTarget.sol';
+
+
+contract Delegator is DelegationTarget {
```
MicahZoltu: Hmm, this shouldn't be showing up in this PR.  I suspect you might have some merge/rebase issues in here.  Let me know if you want some help resolving them.

tests/libraries/test_sortedLinkedList.py
```diff
@@ -19,21 +19,21 @@ def sortedLinkedListContractsFixture(sessionFixture, sortedLinkedListSnapshot):
...
-    assert int256Comparor.compare(1, 1) == 0
-    assert int256Comparor.compare(0, 1) == -1
-    assert int256Comparor.compare(1, 0) == 1
+#def test_helperComparor(sortedLinkedListContractsFixture):
+#
+#    int256Comparor = sortedLinkedListContractsFixture.contracts['int256Comparor']
+#
+#    assert int256Comparor.compare(1, 1) == 0
+#    assert int256Comparor.compare(0, 1) == -1
+#    assert int256Comparor.compare(1, 0) == 1
```
MicahZoltu: These should be uncommonted or deleted (deleted if you remove Comparor contract).

src/libraries/StackUint256.sol
```diff
@@ -0,0 +1,56 @@
...
+import "ROOT/libraries/DelegationTarget.sol";
+import "ROOT/legacy_reputation/Ownable.sol";
+import "ROOT/libraries/Delegator.sol";
+import "ROOT/Controller.sol";
+
+
+contract StackUint256 is DelegationTarget, Ownable {
+    
+    uint256[] private collection;
+    uint256 private head;
```
MicahZoltu: We no longer need `head` in Solidity.  We can use `collection.length` in its place.  For pop, you will need to do something like:
```
function pop() {
	require(collection.length > 0);
	uint256 poppedItem = collection[collection.length - 1];
	delete collection[collection.length - 1];
	collection.length -= 1;
	return poppedItem;
```

For `push` we can just remove the `head += 1` line.

For `peek` we would just use `return collection[collection.length - 1];`

src/libraries/StackUint256.sol
```diff
@@ -0,0 +1,56 @@
...
+    address private owner;
+    bool private initialized;
+
+    function initialize(address _owner) public onlyOwner returns (bool) {
+        require(!initialized);
+        initialized = true;
+        owner = _owner;
+        return (true);
```
MicahZoltu: Style: return statements shouldn't wrap the return expression in parenthesis.

src/libraries/StackUint256.sol
MicahZoltu: The same change can likely be applied to many of the other collections.  Anywhere where we were tracking the "most recently inserted item".  In some cases (like linked lists) though, `head` actually means head of a linked list and is different from the collection length.

MicahZoltu: Actually, I just glanced through and it doesn't look like any of the other collections have a counter.  So just this one should change.

MicahZoltu:
Closing in favor of https://github.com/AugurProject/augur-core/pull/298.  While cleaning up the source code in preparation for finalizing the Solidity migration I realized that we don't actually use most of the collections anymore.  In light of this, I deleted all but the one remaining collection and inlined the migration into a struct+extensions rather than following the previous separate contract pattern.  We may end up going back on this, and luckily Git history will retain all of the code so no need to worry about losing access to it.

Sorry for sending you down a path that ended up not being in line with where we landed @janani-sridhar!

PR #250, opened 2017-07-31
Convert delegator.se and factories to Solidity

nuevoalex:
I ended up needing to write some assembly that roughly matches what we had in `delegator.se` to support returning data from the delegatecall. That code is the most in need of review since it is inherently dangerous being assembly and all.


src/generateExterns.py
```diff
@@ -22,7 +22,7 @@ def getAllFilesInDirectories(sourceDirectories):
     return(allFilePaths)
 
 externLines = []
-for filePath in (getAllFilesInDirectories(['./factories/', './libraries/', './reporting/', './trading/', './extensions/']) + ['./mutex.se']):
```
nuevoalex: This file no longer exists so I removed it here.

src/extensions/marketCreation.se
```diff
@@ -2,6 +2,7 @@
 
 inset('../macros/externs.sem')
 inset('../macros/durations.sem')
+inset('../macros/require.sem')
```
MicahZoltu: Unnecessary addition.

src/factories/BranchFactory.sol
```diff
@@ -0,0 +1,20 @@
...
+
+
+// FIXME: remove once this can be imported as a solidty contract
+contract Branch {
+    function initialize(address parentBranch, int256 parentPayoutDistributionHash);
+}
+
+
+contract BranchFactory {
+
```
MicahZoltu: Style: I'm not personally a fan of empty lines after contract definition.  Though, if we _do_ go with this pattern I recommend a matching empty line before the contract closing `}` to satisfy my OCD.  :D

src/factories/BranchFactory.sol
```diff
@@ -0,0 +1,20 @@
...
+    function createBranch(address controller, address parentBranch, int256 parentPayoutDistributionHash) returns (Branch) {
```
MicahZoltu: Style: Locals and parameters should all start with `_`.

MicahZoltu: `_parenteBranch` should be of type `Branch` and `_controller` should be of type `Controller`.

src/factories/BranchFactory.sol
```diff
@@ -0,0 +1,20 @@
...
+        Delegator del = new Delegator(controller, "branch");
```
MicahZoltu: Recommend the more verbose `_delegator` variable name rather than `_del`.

src/factories/DisputeBondTokenFactory.sol
```diff
@@ -0,0 +1,20 @@
+pragma solidity ^0.4.11;
```
MicahZoltu: We should pick a compiler version and use it consistently throughout the project.  I suspect we'll need to bump to `0.4.14` to pick up a bugfix I'm waiting on for circular imports.  I have been using `0.4.13` in the meantime.

MicahZoltu: At least within this PR there should be consistency.  ;)

src/reporting/ReputationToken.sol
```diff
@@ -20,16 +20,21 @@ contract Branch {
...
     Branch public branch;
     ReputationToken public topMigrationDestination;
 
     function initialize(Branch _branch) public returns (bool) {
         require(_branch != address(0));
         require(branch == address(0));
         branch = _branch;
+        name = "Reputation";
+        symbol = "REP";
+        decimals = 18;
```
MicahZoltu: Ah, good catch.  This makes me sad, but I get it.

tests/conftest.py
```diff
@@ -178,8 +178,11 @@ def upload(self, relativeFilePath, lookupKey = None, signatureKey = None):
         if signatureKey not in ContractsFixture.signatures:
             ContractsFixture.signatures[signatureKey] = ContractsFixture.generateSignature(resolvedPath)
         signature = ContractsFixture.signatures[signatureKey]
+        ct = ContractTranslator(signature)
```
MicahZoltu: Recommend more verbose variable name `contractTranslator`.

tests/reporting/test_market.py
```diff
@@ -19,7 +20,7 @@ def test_market_creation(contractsFixture):
     with raises(TransactionFailed, message="Shady share token should be failing since it is in a separate branch"):
         contractsFixture.createReasonableBinaryMarket(shadyBranch, shadyDenominationToken)
 
-    assert(market.getBranch() == branch.address)
+    assert(longToHexString(market.getBranch()) == branch.address)
```
MicahZoltu: Hmm, I wonder why this change was necessary....

src/libraries/Delegator.sol
```diff
@@ -0,0 +1,38 @@
...
+contract Delegator is DelegationTarget {
+
+    function Delegator(address controllerAddress, bytes32 keyVal) {
+        controller = Controller(controllerAddress);
+        controllerLookupName = keyVal;
+    }
+
+    // We currently only support a single non-array return value
+    // If we truly need to return more advanced sets of data we could follow
+    // the pattern demonstrated here: https://gist.github.com/Arachnid/4ca9da48d51e23e5cfe0f0e14dd6318f
```
MicahZoltu: I'm not personally a fan of the pattern unless the process of seeding the mapping is fully automated by our build/deploy pipeline (which currently is non-existent).

src/libraries/Delegator.sol
```diff
@@ -0,0 +1,38 @@
...
+        // Do nothing if we haven't properly set up the delegator to delegate calls
+        if (controllerLookupName == 0) {
+            return;
+        }
+
+        // Get the delegation target contract
+        address target = controller.lookup(controllerLookupName);
+
+        assembly {
+            let o_code := mload(0x40) //Memory end
```
MicahZoltu: Can you add a more detailed comment about the black magic of 0x40 above this?

MicahZoltu: We should apply the same code style in inline assembly as the rest, which means `_` prefix parameters and locals.

src/libraries/Delegator.sol
```diff
@@ -0,0 +1,38 @@
...
+            // Copy method signature and parameters of this call into memory
+            calldatacopy(o_code, 0x0, calldatasize)
+            // Call the actual method via delegation and store the result at offset o_code assuming size of 32
+            let retval := delegatecall(sub(gas, 10000), target, o_code, calldatasize, o_code, 32)
+            // 0 == it threw, so we do so as well by jumping to bad destination (02)
+            jumpi(0x02, iszero(retval))
+            // Return the data in mem[o_code..32]
+            return(o_code, 32)
+        }
```
MicahZoltu: I'm curious if you tried the no-assembly mechanism for forwarding and it didn't work or if you just did a direct copy?  I would _prefer_ the no-assembly version if that actually works for us.  In particular, I really don't like the `sub(gas, 10000)` bit as it isn't actually the right amount of gas and may change in Metropolis (which would break our contracts).

src/libraries/Delegator.sol
nuevoalex: Totally agree. I thought the idea was neat although the implementation was too potentially fraught with accidents unless we automated it.

src/libraries/Delegator.sol
nuevoalex: I could not find a way to get return values from the no assembly method. The method posted here: https://ethereum.stackexchange.com/questions/2404/upgradeable-smart-contracts doesn't handle return values at all.

src/libraries/Delegator.sol
nuevoalex: Will do!

tests/reporting/test_market.py
nuevoalex: I didn't go too deep exploring exactly why this is occurring, but I suspect it has to do with the externs.sem entry for market being `getBranch:[]:int256`. I wanted to change externs.sem as little as possible though. This may change again when market is a Solidity file too.

src/libraries/Delegator.sol
```diff
@@ -0,0 +1,38 @@
+pragma solidity ^0.4.13;
+
+import 'ROOT/Controller.sol';
+import 'ROOT/libraries/DelegationTarget.sol';
+
+
+contract Delegator is DelegationTarget {
```
MicahZoltu: Why is the delegator a delegation target?

src/libraries/Delegator.sol
```diff
@@ -0,0 +1,38 @@
...
+        // Do nothing if we haven't properly set up the delegator to delegate calls
+        if (controllerLookupName == 0) {
+            return;
+        }
+
+        // Get the delegation target contract
+        address _target = controller.lookup(controllerLookupName);
+
+        assembly {
+            let _o_code := mload(0x40) //0x40 is the address where the next free memory slot is stored in Solidity
```
MicahZoltu: Coding style is camelCase for variable names, not snake case.  Also, recommend a better variable name than `o_code`.  I can't figure out what this variable is from the name.

MicahZoltu: Where did you get 0x40 from?  Presumably we need space for the `controller.lookup(...)` results, but 0x40 offset seems arbitrarily high.  We may want to move the controller lookup into the assembly so we have more control over memory address space, rather than trying to do it separately.

As for variable name, perhaps `_calldataMemoryOffset`?

src/libraries/Delegator.sol
```diff
@@ -0,0 +1,38 @@
...
+            // Copy method signature and parameters of this call into memory
+            calldatacopy(_o_code, 0x0, calldatasize)
+            // Call the actual method via delegation and store the result at offset o_code assuming size of 32
+            let _retval := delegatecall(sub(gas, 10000), _target, _o_code, calldatasize, _o_code, 32)
+            // 0 == it threw, so we do so as well by jumping to bad destination (02)
+            jumpi(0x02, iszero(_retval))
```
MicahZoltu: We should use a switch statement here and revert rather than jump:
```
switch _retval
case 0 {
    revert(0,0)
}
default {
    return(_o_code, 32)
}
```

MicahZoltu:
While you work on the PR for splitting apart the indention Solium check, I recommend just disabling the check in the project in this PR so we can get it merged.  Once you have the Solium check split out we can turn the check (or the split one) back on.

src/libraries/Delegator.sol
nuevoalex: I found 0x40 while looking at how Solidity assembly is generally formatted both in the docs and in many examples of doing something like this. The second answer here has a slightly more detailed explanation of why this is so prevalent: https://ethereum.stackexchange.com/questions/9603/understanding-mload-assembly-function

Use in docs: http://solidity.readthedocs.io/en/develop/assembly.html

Definitely a better name!

MicahZoltu: Ah, good find on that SO article.  So if I understand this code, we are reading whatever is at 0x40, which should be a memory offset, and then writing to the offset returned by that.  Then for the return value we overwrite into the same location, because we know we don't need the space anymore and that address was a good choice before the function call so it seems reasonable that it is a good choice after the function call as well.  I wonder if we should be updating the pointer at `0x40` to be `_calldataMemoryOffset + calldatasize`?  If so, then we should probably store the return value at `_calldataMemoryOffset + calldatasize` and bump the number at 0x40 to `_calldataMemoryOffset + calldatasize + 32`.  Alternatively, we could re-use the space but "allocate" extra to ourselves by setting the value at 0x40 to `_calldataMemoryOffset + max(calldatasize, 32)`.

nuevoalex: I understand the reasoning for doing something like that although I'm not sure when it would actually be necessary to prevent unexpected behavior and I'd prefer the assembly portion have as little code as possible.

MicahZoltu: I also don't understand the ramifications of our decisions which is why I am inclined to "follow the rules" (bump the free pointer appropriately) rather than hope that us using some space but not bumping the free pointer doesn't cause problems.  Alternatively, we could go learn more about how this free pointer is used and build something that is more correct.  :)

MicahZoltu: http://solidity.readthedocs.io/en/develop/assembly.html#standalone-assembly

Check out the example a little down the page where they have `function $allocate`.  You can see here how they update the pointer at 0x40 after reading the current value.

src/libraries/Delegator.sol
```diff
@@ -0,0 +1,52 @@
...
+            // Call the actual method via delegation and store the result at offset _calldataMemoryOffset assuming size of 32
+            let _retval := delegatecall(gas, _target, _calldataMemoryOffset, calldatasize, _calldataMemoryOffset, 32)
+            // 0 == it threw, so we revert
+            switch _retval
+            case 0 {
+                revert(0,0)
+            } default {
+                // If the call succeeded return the data in mem[_calldataMemoryOffset..32]
+                return(_calldataMemoryOffset, 32)
+            }
```
MicahZoltu: Nit: There is a bunch of trailing space on this line.

src/libraries/Delegator.sol
```diff
@@ -0,0 +1,52 @@
...
+                _size := calldatasize
+            } default {
+                _size := 32
+            }
+            mstore(0x40, add(_calldataMemoryOffset, _size))
+            // Copy method signature and parameters of this call into memory
+            calldatacopy(_calldataMemoryOffset, 0x0, calldatasize)
+            // Call the actual method via delegation and store the result at offset _calldataMemoryOffset assuming size of 32
+            let _retval := delegatecall(gas, _target, _calldataMemoryOffset, calldatasize, _calldataMemoryOffset, 32)
+            // 0 == it threw, so we revert
```
MicahZoltu: Consider moving this comment into the case statement so it is more clear what it is related to

PR #251, opened 2017-08-02
Specify our forked Solium for npm and update configuration for Solium.

nuevoalex:


Dockerfile-test
```diff
@@ -2,7 +2,7 @@ FROM python:2.7.13
 
 RUN curl -sL https://deb.nodesource.com/setup_8.x | bash - \
 	&& apt-get install -y nodejs
-RUN npm install -g solium
+RUN npm install -g AugurProject/solium
```
MicahZoltu: Hmm, this will update the Dockerfile to use our fork of Solium, but it won't update things like VSCode extension.  @beaugunderson, if VSCode has a node project open (package.json in root), will it use the version of Solium specified there or will it use some other version?  If some other version, where is it pulling from?

beaugunderson: @MicahZoltu VSCode uses the one from its `package.json`; which is `solium` right now; happy to change it though... not sure how to make it configurable since it's packaged with the extension

PR #253, opened 2017-08-03
Type-Based SafeMath

adrake33:
﻿# Pull Request Template
### Overview

Creating separate libraries for SafeMath based on type.

### Examples (if necessary)


MicahZoltu:
Note: Since this is a pretty critical operation, I would like to ensure that a couple people spend a few minutes thinking about all of the possible things that can go wrong with signed math functions (overflows, underflows, wrong signs, etc.) before we merge this.  If no one else has time (@tinybike @joeykrug @nuevoalex) I'll do a more in depth review tonight (I really need to get my branch debugged so we can merge it as it has a number of quality of life things for everyone else).

joeykrug:
Before I review, has it been checked against these comments on safemath + against Zeppelin safe math repo?

[CoinspectAugurAudit2017v1.0.pdf](https://github.com/AugurProject/augur-core/files/1198675/CoinspectAugurAudit2017v1.0.pdf)

adrake33:
@joeykrug No, I haven't verified it against those.  I'll do that and let you know once it's done.

joeykrug:
```macro sSafeToAdd($a, $b):
(($a >= 0) && ($b <= INT_MAX - $a)) || (($a < 0) && ($b >= INT_MIN - $a))```

Signed safe add

and

```
macro sSafeToSubtract($a, $b):
(($a >= 0) && ($b >= $a -INT_MAX)) || (($a < 0) && ($b <= $a-INT_MIN ))```

Signed safe subtract


From coinspect audit

MicahZoltu:
@joeykrug I suspect that in our signed-safe-math we want to allow the result to be negative.  If the result shouldn't be negative, we should we working with unsigned integers in the first place.  This means that some of the checks in the first (like `$a >= 0`) don't apply.

MicahZoltu:
Alternatively, we could just punt on getting this right and move to push display price to the UI.  😁

/me pushes his secret agenda.

src/library_test_contracts/SafeMathUint256Tester.sol
```diff
@@ -0,0 +1,33 @@
...
+
+import 'ROOT/libraries/math/SafeMathUint256.sol';
+
+
+contract SafeMathUint256Tester {
+    function initialize() public constant {
+    }
+
+    function mul(uint256 _a, uint256 _b) public constant returns (uint256) {
+        return SafeMathUint256.mul(_a, _b);
```
MicahZoltu: The syntax for these can be `_a.mul(_b)` by putting `using SafeMathUint256 for uint256` at the top of the contract.  Recommend using this style as this is how we will be using the library, in case there are any oddities with that usage pattern.

src/library_test_contracts/SafeMathUint256Tester.sol
```diff
@@ -0,0 +1,33 @@
+pragma solidity ^0.4.13;
+
+import 'ROOT/libraries/math/SafeMathUint256.sol';
+
+
+contract SafeMathUint256Tester {
+    function initialize() public constant {
+    }
```
MicahZoltu: Remove unused function.

MicahZoltu: Same for other contract).

src/library_test_contracts/SafeMathUint256Tester.sol
adrake33: Fixed.

src/library_test_contracts/SafeMathUint256Tester.sol
adrake33: Fixed.

PR #254, opened 2017-08-03
Migrates ReportingToken to Solidity.

MicahZoltu:
`Initializable` is a helper base contract that deals with initialized variable and modifiers for methods.

`Typed` is an interface that all of our contracts should derive from.  Just return the name of the contract as a string (which will auto-convert to bytes32).

I also made `Set` only have the restricted access to the mutable members.  This greatly helps with debugging and I think we should migrate to this pattern generally when we move the the libraries to Solidity cc @janani-sridhar (I can help with the merge conflicts when the time comes).

Until the Solidity compiler bug around cyclical imports is fixed, I have created `reporting/Interfaces.sol` to put all of the reporting interfaces.  This will eventually go away, but buys us time while we wait for the solc bugfix.

This PR will probably cause merge conflicts and failing tests for all of the other outstanding PRs, and I apologize for that.  :/  Ping me in #dev if anyone would like assistance dealing with the conflicts.


src/libraries/Initializable.sol
```diff
@@ -0,0 +1,21 @@
+pragma solidity ^0.4.13;
+
+
+contract Initializable {
+    bool public initialized = false;
```
adrake33: Should this be a private member with a corresponding getter?  (Sorry, I know you said you're regretting that decision. :))

src/reporting/ReportingToken.sol
```diff
@@ -0,0 +1,150 @@
...
+import 'ROOT/reporting/Interfaces.sol';
+
+
+contract ReportingToken is DelegationTarget, Typed, Initializable, VariableSupplyToken, IReportingToken {
+    using SafeMath for uint256;
+
+    IMarket public market;
+    int256[] public payoutNumerators;
+
+    function initialize(IMarket _market, int256[] _payoutNumerators) public beforeInitialized returns(bool) {
```
adrake33: Should have a space after 'returns' here.

src/libraries/token/VariableSupplyToken.sol
```diff
@@ -0,0 +1,23 @@
...
+import 'ROOT/libraries/token/StandardToken.sol';
+
+
+contract VariableSupplyToken is StandardToken {
+    using SafeMath for uint256;
+
+    event Mint(address indexed target, uint256 value);
+    event Burn(address indexed target, uint256 value);
+
+    function mint(address _target, uint256 _amount) internal returns (bool) {
```
adrake33: This function (and the next one) are not returning boolean values as specified in their function declarations.

Also, can you please add a TODO or FIXME to the test_dispute_bond_token.py file so we remember to verify that these new functions work properly?

src/libraries/Initializable.sol
MicahZoltu: Fixed.

src/reporting/ReportingToken.sol
MicahZoltu: Fixed.

src/libraries/token/VariableSupplyToken.sol
MicahZoltu: Fixed the returns.  I do agree that we should get some test coverage for this contract, though it is unclear to me how that is related to `test_dispute_bond_token.py`?  This is intended to be used as a base class for some of the internal tokens, but it shouldn't (and doesn't so far) change any external behavior of those tokens.  Its just to simplify the code in a number of places.

MicahZoltu:
Pushed changes per feedback with the exception of the TODO/FIXME, which I commented on (see comment reply for discussion).

src/libraries/token/VariableSupplyToken.sol
adrake33: Okay, I guess I misunderstood what this is for.  You said a while back that when users call disputeBondToken.withdrawToBranch(), we should be minting coins in order to achieve 2x target payout for bond holders during a fork, but this is just intended for ReportingTokens?

MicahZoltu: Ah, that is a separate thing unrelated to this.  This is just a primitive for tokens that need the ability to mint/burn.  The thing you are thinking about is a strategy that would (if/when implemented) leverage this primitive.

PR #255, opened 2017-08-04
Solidity extensions

nuevoalex:
One odd note is that OrderBook had to be modified to take a fixed size array from orderFetcher as Solidity lacks support for contracts receiving a dynamic array from another contract. I set it to 100 initially but I suspect this may provoke discussion on how to handle this sort of case.


src/extensions/MarketCreation.sol
```diff
@@ -0,0 +1,34 @@
...
+// FIXME: Remove this stub once the contract is implemented in Solidity
+contract ReportingWindow {
+    function createNewMarket(int256 _endTime, int256 _numOutcomes, int256 _payoutDenominator, int256 _feePerEthInWei, address _denominationToken, address _sender, int256 _minDisplayPrice, int256 _maxDisplayPrice, address _automatedReporterAddress, int256 _topic) payable returns (Market);
+}
+
+
+// FIXME: Remove this stub once the contract is implemented in Solidity
+contract Branch {
+    function getReportingWindowByMarketEndTime(int256 _endTime, int256 _hasAutomatedReporter) returns (ReportingWindow);
+}
```
MicahZoltu: Recommend importing `reporting/Interfaces.sol` for these (you'll need to rebase/merge develop).

src/extensions/MarketCreation.sol
```diff
@@ -0,0 +1,34 @@
...
+
+contract MarketCreation {
+    function createScalarMarket(Branch _branch, int256 _endTime, int256 _feePerEthInWei, address _denominationToken, int256 _minDisplayPrice, int256 _maxDisplayPrice, address _automatedReporterAddress, int256 _topic) payable public returns (Market _market) {
+        ReportingWindow _reportingWindow = _branch.getReportingWindowByMarketEndTime(_endTime, _automatedReporterAddress != 0 ? 1 : 0);
+        _market = _reportingWindow.createNewMarket.value(msg.value)(_endTime, 2, _maxDisplayPrice - _minDisplayPrice, _feePerEthInWei, _denominationToken, msg.sender, _minDisplayPrice, _maxDisplayPrice, _automatedReporterAddress, _topic);
+        return _market;
+    }
+
+    function createCategoricalMarket(Branch _branch, int256 _endTime, int256 _numOutcomes, int256 _feePerEthInWei, address _denominationToken, address _automatedReporterAddress, int256 _topic) payable public returns (Market _market) {
+        ReportingWindow _reportingWindow = _branch.getReportingWindowByMarketEndTime(_endTime, _automatedReporterAddress != 0 ? 1 : 0);
```
MicahZoltu: Style: I find parenthesis are helpful in ternary statements so one doesn't need to mentally deal with order of operations:
```
(_automatedReporterAddress != 0) ? 1 : 0
```
The idea here is to make it clear to the reader what makes up the conditional and isolate it from the result of the conditional.

Do others have opinions on this?

src/extensions/MarketCreation.sol
```diff
@@ -0,0 +1,34 @@
...
+        _market = _reportingWindow.createNewMarket.value(msg.value)(_endTime, _numOutcomes, _numOutcomes, _feePerEthInWei, _denominationToken, msg.sender, 0, 10 ** 18, _automatedReporterAddress, _topic);
```
MicahZoltu: Style: While I recognize it is mildly inconsistent, I like my `**` operator to be smashed up against the operands:
```
10**18
```
It is inconsistent because it is the only math operator I do this for.  I don't know why I prefer it, and I'll back down if anyone else has an even mildly strong opinion because I know it is inconsistent.

src/extensions/MarketCreation.sol
```diff
@@ -0,0 +1,34 @@
...
+        return _market;
```
MicahZoltu: I _thought_ we decided not to use named return values?  If I am correct in that, then this could be simplified to just `return _reportingWindow.createNewMarket(...)`

MicahZoltu: Same comment above.

src/extensions/MarketCreation.sol
```diff
@@ -0,0 +1,34 @@
...
+
+// FIXME: Remove this stub once the contract is implemented in Solidity
+contract Branch {
+    function getReportingWindowByMarketEndTime(int256 _endTime, int256 _hasAutomatedReporter) returns (ReportingWindow);
+}
+
+
+contract MarketCreation {
+    function createScalarMarket(Branch _branch, int256 _endTime, int256 _feePerEthInWei, address _denominationToken, int256 _minDisplayPrice, int256 _maxDisplayPrice, address _automatedReporterAddress, int256 _topic) payable public returns (Market _market) {
+        ReportingWindow _reportingWindow = _branch.getReportingWindowByMarketEndTime(_endTime, _automatedReporterAddress != 0 ? 1 : 0);
```
MicahZoltu: Have we already had the `var` discussion?  If so tell me what our conclusion was.  I'm personally a fan of `var`, _especially_ when the type is obvious (like this).  I believe the argument against `var` is the fact that it chooses the most constrained type it can, which means `uint8` if you do `var i = 0`.  Knowing this puts me a bit on the fence with Solidity, so I am curious to hear if others have strong opinions.  Also, because editors right now lack strong inference support, unlike Kotlin/C# it isn't trivial to figure out what the inferred type of something is so when it is non-obvious from context, having the explicit type name can be valuable.

src/extensions/OrderBook.sol
```diff
@@ -0,0 +1,42 @@
+pragma solidity ^0.4.13;
+
+// NOTE: We're using address[100] to get back orderIDs from the orderFetcher as Solidity currently lacks
+// the ability to receive dynamic arrays from external contract calls. The list gets shrunk to the desired size
+// before returning
```
MicahZoltu: Style: don't include hard newlines.  If you don't like reading long lines, enable soft-wrap in your editor.

MicahZoltu: I believe in this case we _can_ return a dynamic array because this function is not intended to be called by other contracts.  It is intended to be called by a client which (according to Solidity documentation) can receive a dynamic array result.  I believe all of the cases where we have this problem are in the same situation (called by client) so we can sidestep this whole issue.

src/extensions/OrderBook.sol
```diff
@@ -0,0 +1,42 @@
...
+
+    function getOrderBook(uint256 _type, address _marketID, uint256 _outcomeID, address _startingOrderID, uint256 _numOrdersToLoad) constant public returns (uint256[] _orderBook) {
+        OrdersFetcher _ordersFetcher = OrdersFetcher(controller.lookup("ordersFetcher"));
+        address[100] memory _orders = _ordersFetcher.getOrderIDs(_type, _marketID, _outcomeID, _startingOrderID, _numOrdersToLoad);
+        _orderBook = new uint256[](8 * _numOrdersToLoad);
+        uint8 _numAvailableOrders = 0;
+        uint8 _i = 0;
+        while (_i < _numOrdersToLoad) {
+            // FIXME: Once ordersFetcher.se is converted to solidity make the return value here an 8-tuple and unpack it
+            uint256[8] memory _orderInfo = _ordersFetcher.getOrder(_orders[_i], _type, _marketID, _outcomeID);
```
MicahZoltu: Curious: Is it possible for `_orderInfo` here to be a struct?  Is a struct in Solidity the same as an array in terms of in-memory representation?

src/extensions/OrderBook.sol
```diff
@@ -0,0 +1,42 @@
...
+                uint8 _j = 0;
+                while (_j < 8) {
+                    _orderBook[8 * _numAvailableOrders + _j] = _orderInfo[_j];
+                    _j++;
+                }
+                _numAvailableOrders++;
+            }
+            _i++;
+        }
+        return(_orderBook.slice(0, 8 * _numAvailableOrders));
```
MicahZoltu: Since this is only called by a client (not a contract) can we return an array of structs?  If so I think we should do that as it is far more readable IMO.

src/extensions/MarketCreation.sol
nuevoalex: I could get behind this. Will change here.

src/extensions/MarketCreation.sol
nuevoalex: I do have a mildly strong preferences for the spaces. It makes it more readable to me. It is a _mild_ preference.

src/extensions/ReputationFaucet.sol
```diff
@@ -0,0 +1,24 @@
...
+import 'ROOT/Controller.sol';
+
+
+// FIXME this can be removed once the contract is implemented in Solidity
+contract Branch {
+    function stub () {}
+}
+
+
+contract ReputationFaucet is Controlled {
```
MicahZoltu: I think we might be able to just delete this contract now?  It doesn't actually do anything (and hasn't since reporting branch was merged in).

src/extensions/MarketCreation.sol
nuevoalex: I'm only doing it here because it actually reduces the variable count in the Solidity compiler and we are at/approaching the limit in these functions. I had to go through a lot of variable destruction to make these work in Solidity :/.

src/libraries/arrays/AddressArrays.sol
```diff
@@ -0,0 +1,29 @@
...
+
+library AddressArrays {
+    /*
+     * @dev Returns a slice of the array
+     * @param self The array to make a slice from.
+     * @param start The index to start at.
+     * @param size The target number of elements to return in the slice
+     * @return A newly allocated slice containing the requested elements
+     */
+    function slice(address[] _self, uint256 _start, uint256 _size) internal returns (address[]) {
```
MicahZoltu: Ooh, good use of extension methods.  I like it.  Recommend making sure we have adequate test coverage, as this is one of those areas of code that people will just assume works correctly.

src/extensions/MarketCreation.sol
nuevoalex: > because editors right now lack strong inference support, unlike Kotlin/C# it isn't trivial to figure out what the inferred type of something is so when it is non-obvious from context

This is actually why I would prefer not using var. As a reader of code this makes var on obstruction to immediate understanding.

I get the argument with regards to picking constrained types though so I'm not totally opposed. We should definitely come up with a rule on usage (or avoidance).

src/libraries/arrays/AddressArrays.sol
```diff
@@ -0,0 +1,29 @@
...
+        uint256 _maxSize = _self.length - _start;
+        if (_maxSize < _targetSize) {
+            _targetSize = _maxSize;
+        }
+        uint256 _index = 0;
+        address[] memory _retval = new address[](_targetSize);
+        while (_index < _targetSize) {
+            _retval[_index] = _self[_start + _index];
+            _index++;
+        }
```
MicahZoltu: Recommend using a for loop rather than a while loop here.

src/reporting/ReputationToken.sol
```diff
@@ -36,7 +43,7 @@ contract ReputationToken is DelegationTarget, StandardToken {
         symbol = "REP";
         decimals = 18;
         // FIXME: DELETE THIS BEFORE LAUNCH
-        var _reputationFaucet = controller.lookup("reputationFaucet");
+        var _reputationFaucet = controller.lookup("ReputationFaucet");
```
MicahZoltu: Heh, so apparently we do use the `ReputationFaucet`, even though it doesn't actually faucet reputation.  :P

src/extensions/MarketCreation.sol
nuevoalex: Tested it and we cant get rid of this here though, so I'll simplify it as suggested!

src/trading/completeSets.se
```diff
@@ -105,7 +105,7 @@ def sellCompleteSets(sender: address, market: address, fxpAmount: int256):
     if reportingFee:
         require(denominationToken.transferFrom(market, reportingWindow, reportingFee))
     require(denominationToken.transferFrom(market, sender, sellerPayout))
-
+
```
MicahZoltu: Style: no spaces on empty lines.  I _believe_ if you use an editor that supports .editorconfig this will happen for you with the `trim_trailing_whitespace` setting.

tests/trading/test_completeSets.py
```diff
@@ -36,12 +36,12 @@ def test_publicBuyCompleteSets(contractsFixture):
         {
             "_event_type": "CompleteSets",
             "sender": bytesToHexString(tester.a1),
-	    "reportingFee": 0L,
+	        "reportingFee": 0L,
```
MicahZoltu: Mixed tabs and spaces.  Consider using a `.editorconfig` extension to address this problem as you type.

tests/trading/test_completeSets.py
```diff
@@ -68,7 +68,7 @@ def test_publicBuyCompleteSets_failure(contractsFixture):
 
     # buyCompleteSets exceptions
     with raises(TransactionFailed):
-        completeSets.publicBuyCompleteSets(market.address + 1, fxpAmount, sender=tester.k1)
+        completeSets.publicBuyCompleteSets(0, fxpAmount, sender=tester.k1)
```
MicahZoltu: Did this start failing with `market.address + 1`?  I believe I intentionally didn't set it to `0` because I wanted to make sure it correctly failed if a non-special (but not a market) address was passed in.

tests/trading/test_makeOrder.py
```diff
@@ -155,7 +155,7 @@ def test_makeOrder_failure(contractsFixture):
 
     # makeOrder exceptions (pre-placeBid/placeAsk)
     with raises(TransactionFailed):
-        makeOrder.publicMakeOrder(ASK, fix('1'), fix('0.6'), market.address - 1, YES, 0, 0, 42, sender=tester.k1)
+        makeOrder.publicMakeOrder(ASK, fix('1'), fix('0.6'), market.address, YES, 0, 0, 42, sender=tester.k1)
```
MicahZoltu: I believe this should fail in this case, since `market.address` is a valid market, `publicMakeOrder` should not throw.  I'm surprised this change didn't break the test (suggesting the test has another problem).

src/extensions/MarketCreation.sol
MicahZoltu: I half talked myself out of `var` in my comment, your reply pushes my preference over the edge.  Lets go with no `var` unless someone poses a strong opinion with preference for `var`.

src/extensions/OrderBook.sol
nuevoalex: The OrderBook function does return a dynamic array actually. The comment and note is about the OrdersFetcher interface defined and used here having a uint256[100] signature since we must call it from this contract to get order ids.

src/libraries/arrays/AddressArrays.sol
nuevoalex: I added some tests for the uint256 variant. Parameterizing them for the address variant wouldn't be that hard if you think it's worth it.

tests/trading/test_completeSets.py
nuevoalex: market.address is now a hex string so the operation failed. I can modify this to be more like it was.

tests/trading/test_makeOrder.py
nuevoalex: I thought this one was just a typo. Yea that's kind of worrying if this wan't...

src/extensions/OrderBook.sol
MicahZoltu: Oh, we should remove `getOrderIDs` from `OrderBook` then.  Consider adding to this comment a reminder for that being a potential solution.  I personally don't think any of our internal contracts (not counting extensions) should be returning arbitrary sized arrays.

src/extensions/OrderBook.sol
nuevoalex: Not sure. I'll experiment and report back!

nuevoalex: Only possible in internal function calls :[

src/extensions/OrderBook.sol
nuevoalex: Only possible in internal function calls :[

src/extensions/OrderBook.sol
nuevoalex: Doesn't seem to work based on some very simple experimentation.

`
 struct Data {
        uint256 one;
        uint256 two;
    }

uint256[] memory retval = new uint256[](2);
Data memory newData = Data(1, 2);
retval = newData;
`

fails with

`TypeError: Type struct ArrayTest.Data memory is not implicitly convertible to expected type uint256[] memory.
`

tests/trading/test_makeOrder.py
nuevoalex: This was failing for _multiple_ reasons where it was placed in the test. I will move it to a more appropriate location.

src/reporting/ReportingToken.sol
```diff
@@ -0,0 +1,150 @@
+pragma solidity ^0.4.13;
+
+import 'ROOT/libraries/DelegationTarget.sol';
+import 'ROOT/libraries/Typed.sol';
+import 'ROOT/libraries/Initializable.sol';
+import 'ROOT/libraries/token/VariableSupplyToken.sol';
+import 'ROOT/reporting/Interfaces.sol';
+
+
+contract ReportingToken is DelegationTarget, Typed, Initializable, VariableSupplyToken, IReportingToken {
```
MicahZoltu: Hmm, something must be up with your merge.  This sholudn't be showing as a new file.

MicahZoltu: It appears that you both rebased onto and merged in the reporting branch.  Easiest solution is likely to do an interactive rebase on develop and cut out the merge commit (3a260cb123af0792a09e5d320db0b7d8319afea3).  I can help walk you through the process if you haven't done this before.

src/extensions/MarketCreation.sol
joeykrug: No var imo

PR #256, opened 2017-08-06
Migrates branch to solidity.

MicahZoltu:
Also adds some consistency to addresses being hex strings rather than sometimes longs.


MicahZoltu:
It should be noted that `block.timestamp` is uint256, not int256, so all timestamp storage is now in `uint256` so we don't have to be switching between types.  I don't think anything we need will ever need a timestamp before 1970, so this policy shouldn't cause any problems.  I have updated all that I could find in this PR, but I may have missed some.

PR #257, opened 2017-08-07
Updated parameter names and types for extensions/OrderBook.sol

tinybike:


src/extensions/OrderBook.sol
```diff
@@ -1,34 +1,36 @@
...
 import 'ROOT/Controller.sol';
+import 'ROOT/reporting/Interfaces.sol';
 
 
 // FIXME: Remove this stub once the contract is converted to Solidity
 contract OrdersFetcher {
-    function getOrderIDs(uint256 _type, address _marketID, uint256 _outcomeID, address _startingOrderID, uint256 _numOrdersToLoad) returns (address[100]);
-    function getOrder(address _orderID, uint256 _type, address _marketID, uint256 _outcomeID) returns (uint256[8]);
+    function getOrderIDs(uint256 _type, IMarket _market, uint256 _outcome, uint256 _startingOrderId, uint256 _numOrdersToLoad) returns (uint256[100]);
+    function getOrder(uint256 _orderId, uint256 _type, IMarket _market, uint256 _outcome) returns (uint256[8]);
```
MicahZoltu: You should update the Serpent to match, otherwise this will fail to work when called.

tinybike: Should I use `address` in place of `IMarket` in Serpent?

MicahZoltu:
Yeah, address for any contract parameter.

src/extensions/OrderBook.sol
tinybike: Updated in https://github.com/AugurProject/augur-core/commit/366ad0d5a2eee5f42fb9d92961a403d57bb70d74

src/trading/ordersFetcher.se
```diff
@@ -43,16 +43,16 @@ def getOrderIDs(type: int256, market: address, outcome: int256, startingOrderID:
     return(slice(orderIDs, items=0, items=i): arr)
 
 # @public
-def getOrder(orderID: address, type: int256, market: address, outcome: int256):
+def getOrder(_orderId: uint256, _type: uint256, _market: address, _outcome: uint256):
```
MicahZoltu: Note to whoever migrates this to Solidity (@adrake33?).  Type should be an enum with two options and `_outcome` should be a `uint8`.  Serpent doesn't support either of these, so that change will have to wait.

adrake33: Ok, thanks.  I'll make a note of this when migrating this file.

PR #258, opened 2017-08-08
First commit of delegator tests and possible repToken change

nuevoalex:
Just to confirm it would function as expected I also included a potential change to ReputationToken, but we should discuss what we actually want the recourse to be for that case.

### DIFF
--- a/src/reporting/ReputationToken.sol
+++ b/src/reporting/ReputationToken.sol
@@ -11,9 +11,10 @@ import 'ROOT/reporting/Interfaces.sol';
 contract ReputationToken is DelegationTarget, Typed, Initializable, StandardToken, IReputationToken {
     using SafeMath for uint256;
 
-    string public name;
-    string public symbol;
-    uint256 public decimals;
+    // Delegated constracts cannot use string values, so we use bytes32 values here
+    bytes32 constant public name = "Reputation";
+    bytes32 constant public symbol = "REP";
+    uint256 constant public decimals = 18;
     IBranch private branch;
     IReputationToken private topMigrationDestination;
 
@@ -21,9 +22,6 @@ contract ReputationToken is DelegationTarget, Typed, Initializable, StandardToke
         endInitialization();
         require(_branch != address(0));
         branch = _branch;
-        name = "Reputation";
-        symbol = "REP";
-        decimals = 18;
         // FIXME: DELETE THIS BEFORE LAUNCH
         var _reputationFaucet = controller.lookup("ReputationFaucet");
         balances[_reputationFaucet] = balances[_reputationFaucet].add(1000000 ether);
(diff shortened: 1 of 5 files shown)


src/reporting/ReputationToken.sol
```diff
@@ -11,19 +11,17 @@ import 'ROOT/reporting/Interfaces.sol';
 contract ReputationToken is DelegationTarget, Typed, Initializable, StandardToken, IReputationToken {
     using SafeMath for uint256;
 
-    string public name;
-    string public symbol;
-    uint256 public decimals;
+    // Delegated constracts cannot use string values, so we use bytes32 values here
+    bytes32 constant public name = "Reputation";
+    bytes32 constant public symbol = "REP";
```
MicahZoltu: The ERC20 interface for these is `function name() constant returns (string)` and `function symbol() constant returns (string)`.  I'm not sure what to do about this, since we can't return a string from a delegated contract, and this doesn't match the ERC20 interface.

MicahZoltu: Also, `constant public` is redundant I believe (though, if its not I need to fixup a bunch of the stuff I wrote).

tests/libraries/test_delegator.py
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/env python
+
+from ethereum.tools import tester
+from ethereum.tools.tester import TransactionFailed
+from pytest import fixture, mark, lazy_fixture, raises
+from ethereum.config import config_metropolis
+from utils import stringToBytes
+
+#config_metropolis['BLOCK_GAS_LIMIT'] = 2**128
```
MicahZoltu: Cleanup: Delete rather than leave commented out.

tests/libraries/test_delegator.py
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/env python
+
+from ethereum.tools import tester
+from ethereum.tools.tester import TransactionFailed
+from pytest import fixture, mark, lazy_fixture, raises
+from ethereum.config import config_metropolis
```
MicahZoltu: Not needed once you delete unused thing below.

src/reporting/ReputationToken.sol
nuevoalex: I can only think of very gross workarounds then. For example creating a subclass of Delegator for ERC20 that provides these string values :[

nuevoalex: Regarding `constant public` a very naive experiment in Remix suggests it is not redundant.

```
pragma solidity ^0.4.13;

contract Test {
    
    string constant notDecorated = "one";
    string public constant decorated = "two";
}
```

Only presents the decorated version so there is at least some differentiation even if there is a way to access it outside of their UI.

PR #259, opened 2017-08-08
Convert DisputeBondToken and RegistrationToken to Solidity

nuevoalex:


src/reporting/DisputeBondToken.sol
```diff
@@ -0,0 +1,107 @@
...
+        uint256 amountToTransfer = reputationToken.balanceOf(this);
+        bondRemainingToBePaidOut = bondRemainingToBePaidOut.sub(amountToTransfer);
+        reputationToken.transfer(bondHolder, amountToTransfer);
+        return true;
+    }
+
+    // FIXME: We should be minting coins in this scenario in order to achieve 2x
+    // target payout for bond holders during a fork.  Ideally, the amount minted is
+    // capped at the amount of tokens redeemed on other branches, so we may have to
+    // require the user to supply branches to deduct from with their call to this.
```
MicahZoltu: Style: Don't put hard newlines in prose text.  Editor softwraps solve this problem well and hard newlines like this will totally screw up viewing the file/diff on a phone, side-by-side in a smallish browser window, etc.

src/reporting/market.se
```diff
@@ -430,7 +430,7 @@ def getShareToken(outcome: int256):
 
 def getFinalPayoutDistributionHash():
     assertNoValue()
-    return(self.finalPayoutDistributionHash)
+    return(self.finalPayoutDistributionHash : bytes32)
```
MicahZoltu: Style: I think elsewhere we are doing `name: type` (no space before `:`).  I don't think we have discussed this, but it is inconsistent here even within this change.  Recommend either changing to `name: type` or starting a discussion around which style we want to go with as a team.

PR #260, opened 2017-08-09
Fixes bug that prevented cyclic imports.

MicahZoltu:
Also removes the interfaces for contracts already migrated to Solidity.


joeykrug:
Fixed merge conflict + made safemath lib import more explicit

PR #261, opened 2017-08-11
Convert ReportingWindow to Solidity

nuevoalex:


src/libraries/set.se
```diff
@@ -34,8 +34,11 @@ def remove(item):
     self.count -= 1
     return(1)
 
-def contains(item):
+def contains(item : address):
```
MicahZoltu: Same comment from other PR about style here (space before `:`).

src/reporting/ReportingWindow.sol
```diff
@@ -0,0 +1,289 @@
...
+        RegistrationTokenFactory _registrationTokenFactory = RegistrationTokenFactory(controller.lookup("RegistrationTokenFactory"));
+        registrationToken = _registrationTokenFactory.createRegistrationToken(controller, this);
+        SetFactory _setFactory = SetFactory(controller.lookup("SetFactory"));
+        markets = _setFactory.createSet(controller, this);
+        limitedReporterMarkets = _setFactory.createSet(controller, this);
+        allReporterMarkets = _setFactory.createSet(controller, this);
+        return true;
+    }
+
+    function createNewMarket(uint256 _endTime, int256 _numOutcomes, int256 _payoutDenominator, int256 _feePerEthInWei, address _denominationToken, address _creator, int256 _minDisplayPrice, int256 _maxDisplayPrice, address _automatedReporterAddress, int256 _topic) public payable returns (IMarket _newMarket) {
```
MicahZoltu: numOutcomes should be `uint8` as it can only be 2,3,4,5,6,7 or 8.

MicahZoltu: Though, if changing that ends up being a big change (shouldn't be, but might) we can put that change off until after the Solidity migration is complete.

PR #263, opened 2017-08-17
Ported claimProceeds to Solidity

redsquirrel:
### Overview
Porting claimProceeds.se to ClaimProceeds.sol.

This should be functionally equivalent.


src/trading/ClaimProceeds.sol
```diff
@@ -0,0 +1,51 @@
...
+import 'ROOT/Controller.sol';
+import 'ROOT/reporting/Interfaces.sol';
+import 'ROOT/extensions/MarketFeeCalculator.sol';
+
+/**
+ * @title ClaimProceeds
+ * @dev This allows users to claim their money from a market by exchanging their shares 
+ */
+contract ClaimProceeds is Controlled {
+	uint256 constant THREE_DAYS = 3 * 24 * 60 * 60;
```
adrake33: In our code, we decided to make all members private and then write getters for all of the members: https://augurproject.slack.com/archives/C031NRD9B/p1501974432517392.  (We may decide to change this policy later, though.)  This should probably be made private, but since this is a constant, I'm not sure if a getter is required here.  @MicahZoltu Are we also creating getters for all constants?

src/trading/ClaimProceeds.sol
```diff
@@ -0,0 +1,51 @@
...
+
+	function publicClaimProceeds(IMarket market) onlyInGoodTimes returns(uint256) {
```
adrake33: One of our style guidelines is to explicitly declare public functions, so the visibility here should be set to public.

src/trading/ClaimProceeds.sol
```diff
@@ -0,0 +1,51 @@
...
+		// TODO: controller.assertIsWhitelisted(msg.sender);
+		require(market.isFinalized());
+		require(block.timestamp > market.getFinalizationTime() + THREE_DAYS);
+
+	    uint8 outcome = 0;
```
adrake33: Another style guidelines of ours is to prepend all local variables with "_" so as to distinguish them from storage variables (i.e., "outcome" should become "_outcome", etc.).

src/trading/ClaimProceeds.sol
adrake33: Never mind about making it public.  I saw Micah's comment on Slack about making it internal instead.

src/reporting/Interfaces.sol
```diff
@@ -34,12 +33,22 @@ contract IMarket is Typed {
     function isDoneWithAllReporters() constant public returns (bool);
     function isDoneWithLimitedReporters() constant public returns (bool);
     function getReportingTokenOrZeroByPayoutDistributionHash(bytes32) constant public returns (ReportingToken);
+    function getFinalizationTime() constant public returns (uint);
```
MicahZoltu: Style: Prefer using explicitly sized types rather than implicitly sized types.  In this case, `uint256`

src/reporting/Interfaces.sol
```diff
@@ -34,12 +33,22 @@ contract IMarket is Typed {
...
+    function getDenominationToken() constant public returns (ERC20);
+    function getPayoutDenominator() constant public returns (int256);
```
MicahZoltu: This should (will) return `uint256`.  In Serpent, `int256` is really the only type which is why everything is `int256`.  As we have been migrating to Solidity, we have been adjusting things to their correct types and in this case that is unsigned.

MicahZoltu: Same for the other `int256` returning methods you added to this interface.

src/reporting/Interfaces.sol
```diff
@@ -34,12 +33,22 @@ contract IMarket is Typed {
...
+    function getMarketCreatorSettlementFeeInAttoethPerEth() constant public returns (int256);
+    function getCreator() constant public returns (address);
+    function shouldCollectReportingFees() constant public returns (bool);
 }
 
 
 contract IShareToken is Typed {
     function initialize(IMarket _market, uint8 _outcome) public returns (bool);
     function getMarket() constant returns (IMarket);
+    function balanceOf(address) constant returns (int256);
```
MicahZoltu: Rather than adding `balanceOf` here, recommend making `IShareToken` derive from `ERC20` which will bring in all of the token stuff.

src/reporting/Interfaces.sol
```diff
@@ -34,12 +33,22 @@ contract IMarket is Typed {
...
+    function destroyShares(address, int256 balance) public;
```
MicahZoltu: As mentioned above, this should be unsigned.

src/trading/ClaimProceeds.sol
```diff
@@ -0,0 +1,50 @@
...
+import 'ROOT/Controller.sol';
+import 'ROOT/reporting/Interfaces.sol';
+import 'ROOT/extensions/MarketFeeCalculator.sol';
+
+/**
+ * @title ClaimProceeds
+ * @dev This allows users to claim their money from a market by exchanging their shares 
+ */
+contract ClaimProceeds is Controlled {
+	uint256 private constant THREE_DAYS = 3 * 24 * 60 * 60;
```
MicahZoltu: Solidity has a concept of units so this can be simplified to `= 3 days`.  Because of that, this variable is actually no longer useful so I recommend just changing the place in code its used to be `3 days` and deleting this variable.

src/trading/ClaimProceeds.sol
```diff
@@ -0,0 +1,50 @@
...
+
+	function publicClaimProceeds(IMarket market) onlyInGoodTimes public returns(uint256) {
```
MicahZoltu: This function should return a `bool`.  As mentioned above, this was returning an int in Serpent only because there is no boolean type in Serpent.

src/trading/ClaimProceeds.sol
```diff
@@ -0,0 +1,50 @@
...
+		require(market.isFinalized());
+		require(block.timestamp > market.getFinalizationTime() + THREE_DAYS);
+
+	    uint8 outcome = 0;
+	    while (outcome < market.getNumberOfOutcomes()) {
```
MicahZoltu: Solidity has a for loop, so this should be changed from a while loop to `for (uint8 outcome = 0; outcome < market.getNumberOfOutcomes(); ++outcome)`.

It would also be good to extract `numberOfOutcomes` out into a variable so we don't have to do a cross-transaction call with every loop iteration.

src/trading/ClaimProceeds.sol
```diff
@@ -0,0 +1,50 @@
...
+	        int256 payoutNumerator = market.getFinalWinningReportingToken().getPayoutNumerator(outcome);
```
MicahZoltu: All of these int256 variables should be uint256.

src/trading/ClaimProceeds.sol
```diff
@@ -0,0 +1,50 @@
...
+	        IShareToken shareToken = market.getShareToken(outcome);
+	        int256 balance = 100;// TODO shareToken.balanceOf(msg.sender);
+	        int256 proceedAmount = balance * market.getCompleteSetCostInAttotokens() / 10**18 * payoutNumerator / market.getPayoutDenominator();
```
MicahZoltu: This should be using safe math functions rather than raw `*` and `/`.  You can use this by adding `using SafeMath for uint256;` to the top of the contract and then `balance.mul(...)`.

src/trading/ClaimProceeds.sol
```diff
@@ -0,0 +1,50 @@
...
+import 'ROOT/reporting/Interfaces.sol';
+import 'ROOT/extensions/MarketFeeCalculator.sol';
+
+/**
+ * @title ClaimProceeds
+ * @dev This allows users to claim their money from a market by exchanging their shares 
+ */
+contract ClaimProceeds is Controlled {
+	uint256 private constant THREE_DAYS = 3 * 24 * 60 * 60;
+
```
MicahZoltu: Can you please add a comment here (this is unrelated to your PR, but noticed in review):
```
AUDIT: Ensure that a malicious market can't subversively cause share tokens to be paid out incorrectly.
```

src/trading/ClaimProceeds.sol
```diff
@@ -0,0 +1,50 @@
...
+	            } else {
+	            	reportingFee = 0;
+	            }
+	            uint256 tokenHolderPayout = uint256(proceedAmount) - uint256(marketCreatorFee) - reportingFee;
+	            // NOTE: rounding error here will result in _very_ tiny amounts of denominationToken left in the market
+	            // require(market.getDenominationToken().transferFrom(market, msg.sender, tokenHolderPayout));
+	            // require(market.getDenominationToken().transferFrom(market, market.getCreator(), uint(marketCreatorFee)));
+	            // require(market.getDenominationToken().transferFrom(market, reportingWindow, reportingFee));       	
+	        }
+	    	outcome += 1;
```
MicahZoltu: Make sure to delete this when you change to a for loop.  ;)

src/trading/ClaimProceeds.sol
```diff
@@ -0,0 +1,50 @@
...
+
+	    var numberOfOutcomes = market.getNumberOfOutcomes();
+		for (uint8 outcome = 0; outcome < numberOfOutcomes; ++outcome) {
+	        uint256 payoutNumerator = uint256(market.getFinalWinningReportingToken().getPayoutNumerator(outcome));
+	        IShareToken shareToken = market.getShareToken(outcome);
+	        uint256 balance = 100; // shareToken.balanceOf(msg.sender);
+	        uint256 proceedAmount = balance.mul(market.getCompleteSetCostInAttotokens() / 10**18 * payoutNumerator / market.getPayoutDenominator());
+
+	        if (balance > 0) {
+	        	// TODO shareToken.destroyShares(msg.sender, balance);
```
joeykrug: Why's this commented with a todo

redsquirrel: Because I'm still working on it. 😸  Currently any calls to `shareToken` result in a failed transaction. 🙁  Trying to determine the cause. 🛠

src/trading/ClaimProceeds.sol
```diff
@@ -0,0 +1,50 @@
...
+
+	function publicClaimProceeds(IMarket market) onlyInGoodTimes public returns(bool) {
+		require(market.isFinalized());
+		require(block.timestamp > market.getFinalizationTime() + 3 days);
+
+	    var numberOfOutcomes = market.getNumberOfOutcomes();
+		for (uint8 outcome = 0; outcome < numberOfOutcomes; ++outcome) {
+	        uint256 payoutNumerator = uint256(market.getFinalWinningReportingToken().getPayoutNumerator(outcome));
+	        IShareToken shareToken = market.getShareToken(outcome);
+	        uint256 balance = 100; // shareToken.balanceOf(msg.sender);
```
redsquirrel: A bit stumped about why `shareToken.balanceOf(msg.sender)` blows up when uncommented. Need to dig into this further.

MicahZoltu:
File has mixed tabs and spaces.  Also, there should be two blank lines before the contract rather than 1 (both should go above contract comment).

Our team uses Solium linter, I recommend getting it setup so you can find these quickly and easily yourself.  If you use VSCode, you can install Solidity Extended extension to get in-editor support and syntax highlighting.

redsquirrel:
@MicahZoltu I installed Solium linter via npm. When I run `solium --file src/trading/ClaimProceeds.sol`, I get:

```
An error occurred while running the linter on src/trading/ClaimProceeds.sol:
Error: Rule comma-whitespace was not found
    at /usr/local/lib/node_modules/solium/lib/rules.js:128:11
    at Array.forEach (native)
    at Object.module.exports.load (/usr/local/lib/node_modules/solium/lib/rules.js:126:27)
    at EventEmitter.lint (/usr/local/lib/node_modules/solium/lib/solium.js:72:31)
    at lintString (/usr/local/lib/node_modules/solium/lib/cli.js:133:25)
    at lintFile (/usr/local/lib/node_modules/solium/lib/cli.js:166:9)
    at /usr/local/lib/node_modules/solium/lib/cli.js:187:11
    at Array.map (native)
    at lint (/usr/local/lib/node_modules/solium/lib/cli.js:186:33)
    at Object.execute (/usr/local/lib/node_modules/solium/lib/cli.js:311:19)
No issues found.
```

MicahZoltu:
Ah, I think you need to use the fork of Solium that is in the project's package.json.  If you are on a modern version of node you can do:
```
npm install
npx solium ...
```

redsquirrel:
@MicahZoltu Yep, that was the issue. Thanks.

src/trading/ClaimProceeds.sol
```diff
@@ -0,0 +1,56 @@
...
+import 'ROOT/Mutex.sol';
+import 'ROOT/reporting/Interfaces.sol';
+import 'ROOT/extensions/MarketFeeCalculator.sol';
+import 'ROOT/legacy_reputation/SafeMath.sol';
+
+
+/**
+ * @title ClaimProceeds
+ * @dev This allows users to claim their money from a market by exchanging their shares 
+ * AUDIT: Ensure that a malicious market can't subversively cause share tokens to be paid out incorrectly.
```
MicahZoltu: Can you put this as a line comment separate from the doc comment?  When searching for these I usually search for `// [A-Z][A-Z][A-Z][A-Z]` so I can find `// TODO`, `// FIXME` and `// AUDIT` all at once.

src/trading/ClaimProceeds.sol
```diff
@@ -0,0 +1,56 @@
...
+ */
+contract ClaimProceeds is Controlled {
+    using SafeMath for uint256;
+
+    function publicClaimProceeds(IMarket market) onlyInGoodTimes public returns(bool) {
```
MicahZoltu: Style: All local variables and parameters should be prefixed with `_`.  This is to avoid name collisions with storage variables as well as make it clear to the viewer the scope of the variable.

MicahZoltu:
I believe @adrake33 mentioned this earlier but it appears not addressed yet.  All local variables and parameters should be named `_camelCase` with a leading underscore.

redsquirrel:
@adrake33 @MicahZoltu This is ready for final review. All tests are passing for me.

src/macros/externs.sem
```diff
@@ -34,8 +34,9 @@ extern completeSetsExtern: [buyCompleteSets:[address,address,int256]:int256, get
...
 extern ordersFetcherExtern: [ascendOrderList:[int256,address,int256,int256,address]:int256[], descendOrderList:[int256,address,int256,int256,address]:int256[], findBoundingOrders:[int256,address,int256,int256,address,address,address,address]:int256[], getOrder:[uint256,uint256,address,uint256]:int256[], getOrderIDs:[int256,address,int256,address,int256]:int256[], initialize:[address]:_]
+<<<<<<< e26b60c78c3cc878ce9efafc5ad08a23731f6b25
```
adrake33: This line should be removed.

src/trading/ClaimProceeds.sol
```diff
@@ -0,0 +1,84 @@
...
+        _reporterShare = calculateReportingFee(_market, _proceeds);
+        _shareHolderShare = _proceeds.sub(_creatorShare).sub(_reporterShare);
+        return (_shareHolderShare, _creatorShare, _reporterShare);
+    }
+
+    function calculateProceeds(IMarket _market, ReportingToken _winningReportingToken, uint8 _outcome, uint256 _numberOfShares) constant returns (uint256) {
+        uint256 _completeSetCostInAttotokens = _market.getCompleteSetCostInAttotokens();
+        uint256 _payoutNumerator = _winningReportingToken.getPayoutNumerator(_outcome);
+        uint256 _getPayoutDenominator = _market.getPayoutDenominator();
+        return _numberOfShares.mul(_completeSetCostInAttotokens).div(10**18).mul(_payoutNumerator).div(_getPayoutDenominator);
```
adrake33: Along with my changes for Orders/OrdersFetcher, I added a new function in SafeMathUint256 called fxpMul(), which multiplies 2 numbers and divides by (10**18). It would probably be more straightforward to use this function once both of our changes are merged to the develop branch.

src/trading/ClaimProceeds.sol
```diff
@@ -0,0 +1,84 @@
...
+                require(_denominationToken.transferFrom(_market, _market.getReportingWindow(), _reporterShare));
+            }
+        }
+
+        _mutex.release();
+
+        return true;
+    }
+
+    function divideUpWinnings(IMarket _market, ReportingToken _winningReportingToken, uint8 _outcome, uint256 _numberOfShares) constant returns (uint256 _shareHolderShare, uint256 _creatorShare, uint256 _reporterShare) {
```
adrake33: This function and the ones that follow should all be explicitly declared as public.

src/macros/externs.sem
redsquirrel: Weird, not sure how I missed that.

src/reporting/Interfaces.sol
```diff
@@ -27,19 +27,28 @@ contract IMarket is Typed {
...
+    function getShareToken(uint8) constant public returns (IShareToken);
+    function getCompleteSetCostInAttotokens() constant public returns(uint256);
+    function getMarketCreatorSettlementFeeInAttoethPerEth() constant public returns (uint256);
+    function getCreator() constant public returns (address);
+    function shouldCollectReportingFees() constant public returns (bool);
 }
 
 
-contract IShareToken is Typed {
+contract IShareToken is ERC20, Typed {
```
MicahZoltu: Note to future us: We should change `ShareToken` to `VariableSupplyToken` when we port it to Solidity.

src/trading/ClaimProceeds.sol
```diff
@@ -0,0 +1,84 @@
...
+            IShareToken _shareToken = _market.getShareToken(_outcome);
+            uint256 _numberOfShares = _shareToken.balanceOf(msg.sender);
+            var (_shareHolderShare, _creatorShare, _reporterShare) = divideUpWinnings(_market, _winningReportingToken, _outcome, _numberOfShares);
+
+            // always destroy shares as it gives a minor gas refund and is good for the network
+            if (_numberOfShares > 0) {
+                _shareToken.destroyShares(msg.sender, _numberOfShares);
+            }
+            ERC20 _denominationToken = _market.getDenominationToken();
+            // NOTE: rounding error here will result in _very_ tiny amounts of denominationToken left in the market
```
MicahZoltu: This comment should be moved onto `calculateProceeds` since that is where the problem actually lies.

src/trading/ClaimProceeds.sol
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * @title ClaimProceeds
+ * @dev This allows users to claim their money from a market by exchanging their shares
+ */
+contract ClaimProceeds is Controlled {
+    using SafeMath for uint256;
+
+    function publicClaimProceeds(IMarket _market) onlyInGoodTimes public returns(bool) {
+        Mutex _mutex = Mutex(controller.lookup("Mutex"));
+        _mutex.acquire();
```
MicahZoltu: Switch this to use the `ReentrancyGuard` `nonReentrant` modifier rather than manually managing the mutex.

src/trading/ClaimProceeds.sol
```diff
@@ -0,0 +1,84 @@
...
+
+// AUDIT: Ensure that a malicious market can't subversively cause share tokens to be paid out incorrectly.
+/**
+ * @title ClaimProceeds
+ * @dev This allows users to claim their money from a market by exchanging their shares
+ */
+contract ClaimProceeds is Controlled {
+    using SafeMath for uint256;
+
+    function publicClaimProceeds(IMarket _market) onlyInGoodTimes public returns(bool) {
```
MicahZoltu: Rename this method to `claimProceeds`.  We used to put the public prefix on it in Serpent since Serpent didn't actually have visibility modifiers.

MicahZoltu: Change the visibility to `external`.  Anything that is protected by a mutex/reentrancy guard should be market as `external`.

src/trading/ClaimProceeds.sol
redsquirrel: @adrake33 I noticed that `fxpMul ` function is now is develop. But I don't see where it uses 10**18.

MicahZoltu: You pass in a base, in this case `10^18`.
`fxpMul(5*10**18, 12*10**17, 10**18)`

PR #264, opened 2017-08-18
Minor fixes in Cash.

MicahZoltu:
Changes Cash to be `VariableSupplyToken` which handles minting and burning.  Also removed the now redundant events (`VariableSupplyToken` fires a mint and burn event already).

Removes unused `_to` parameter on withdraw.

Removes the `public` prefix on deposit and withdraw, since neither of these will be called internally.


MicahZoltu:
@JohnDanz These changes may have an impact on documentation, if you have documented this class.
@tinybike These changes may have an impact on the middleware, if you were already consuming the events or calling the deposit/withdraw functions.

MicahZoltu:
All relevant changes are in Cash.sol.  The rest are just find & replace renames as a side effect of that change.

PR #268, opened 2017-08-20
Market solidity

nuevoalex:
To get this under the size limit (And to simplify the logic a little bit) I did refactor how reporting phases are determined internally by using a state Enum. With this refactor and some other cleanup I was actually able to get the size under the EVM limit without doing a major split of the market contract.

We may still want to do something larger to reduce its size but now it can be punted till after the Solidity conversion.

There was a fair bit of code I removed due to it being redundant (e.g removing a require that is called in a function immediately proceeding it) but it is possible I broke something or made a bad assumption and things just aren't fully tested, so the review here is extra appreciated.


src/reporting/ReportingWindow.sol
```diff
@@ -289,4 +291,22 @@ contract ReportingWindow is DelegationTarget, Typed, Initializable {
         numberOfReportsByMarket[_market] += 1;
         return true;
     }
+
+    function getWinningPayoutDistributionHashFromFork(Market _market) constant returns (bytes32) {
```
MicahZoltu: I'm assuming you moved this to help get the contract size down?  This function doesn't really make sense in ReportingWindow IMO since its a function for asking about details of a market.

nuevoalex: Yea, there were a few small functions toward the end of the reduction where I just moved them into another contract.

MicahZoltu: If that is what it comes to for reducing market contract size, I recommend we just create a `MarketExtensions` (name TBD) function.  That way it is at least clear, and users can do `using MarketExtensions for Market;` in their contract to make it _feel_ like its all one contract.

nuevoalex: I like that pattern! Will do.

nuevoalex: One oddity about this example and perhaps others is that it is the Market itself calling this function. I believe using the `MarketExtensions` library in the market contract itself would increase the size by the library size. That means for such cases I'd still be using other contracts as proxies to call the `MarketExtensions` library methods within the Market contract.

MicahZoltu: From market you could just do `MarketExtensions(controller.lookup('MarketExtensions')).getWinningPayoutDistributionHash(this)`

nuevoalex: I think it wouldn't be a library then and we couldn't use the `using MarketExtensions for Market` pattern then though correct? I'm fine with either, just making sure that's the understanding.

src/reporting/ReportingWindow.sol
```diff
@@ -289,4 +291,22 @@ contract ReportingWindow is DelegationTarget, Typed, Initializable {
...
+        if (getBranch().getForkingMarket() != _market) {
+            return 0;
+        }
+        ReputationToken _winningDestination = getReputationToken().getTopMigrationDestination();
+        if (address(_winningDestination) == address(0)) {
+            return 0;
+        }
+        if (_winningDestination.totalSupply() < 11 * 10**6 * 10**18 / 2 && block.timestamp < getBranch().getForkEndTime()) {
```
MicahZoltu: We should probably move this magic number into a variable, just so we can have it named.  I think constant variables are inlined in Solidity, so wouldn't increase contract size over this?

MicahZoltu: Note: There is a good chance that a lot of the comments I have are actually about the original code (which I wrote).  Feel free to ignore such comments until some time in the future, though I'm a fan of the campsite rule (which, when I Google it I only find [this](https://en.wikipedia.org/wiki/Savage_Love#Campsite_rule)... which is not what I had in mind but seems to apply well none the less).

src/reporting/Market.sol
```diff
@@ -0,0 +1,653 @@
...
+    bool private automatedReportReceived;
+    bytes32 private tentativeWinningPayoutDistributionHash;
+    bytes32 private finalPayoutDistributionHash;
+    DisputeBondToken private automatedReporterDisputeBondToken;
+    DisputeBondToken private limitedReportersDisputeBondToken;
+    DisputeBondToken private allReportersDisputeBondToken;
+    uint256 private validityBondAttoeth;
+    uint256 private automatedReporterBondAttoeth;
+
+    function initialize(ReportingWindow _reportingWindow, uint256 _endTime, uint8 _numOutcomes, uint256 _payoutDenominator, uint256 _feePerEthInAttoeth, Cash _denominationToken, address _creator, int256 _minDisplayPrice, int256 _maxDisplayPrice, address _automatedReporterAddress, bytes32 _topic) public payable beforeInitialized returns (bool _success) {
```
MicahZoltu: Since this contract is `initializable` I wonder if we should tag all of the methods in here with `afterInitialized` (or whatever the modifier is called).  I'm not sure yet how I feel about the practice yet, it is nice for small contracts that have a substantial amount of code executed during/before initialization, but in our case I don't think there is any pre-initialization code.

src/reporting/Market.sol
```diff
@@ -0,0 +1,653 @@
...
+        endInitialization();
+        require(address(_reportingWindow) != NULL_ADDRESS);
+        require(_payoutDenominator >= MIN_PAYOUT_DENOMINATOR);
```
MicahZoltu: This should actually be `MIN_NUM_OUTCOMES`.  That being said, 2 lines down we require that payout denominator must be evenly divisible by num outcomes, so I don't think this check is necessary.  Recommend removing it.

src/reporting/Market.sol
```diff
@@ -0,0 +1,653 @@
...
+    // CONSIDER: we really don't need these
+    int256 private maxDisplayPrice;
+    int256 private minDisplayPrice;
+
+    // CONSIDER: figure out approprate values for these
+    uint256 private constant AUTOMATED_REPORTER_DISPUTE_BOND_AMOUNT = 11 * 10**20;
+    uint256 private constant LIMITED_REPORTERS_DISPUTE_BOND_AMOUNT = 11 * 10**21;
+    uint256 private constant ALL_REPORTERS_DISPUTE_BOND_AMOUNT = 11 * 10**22;
+
+    uint256 private constant MIN_NUM_OUTCOMES = 2;
```
MicahZoltu: Style: I'm generally not a fan of constants that are only used once/in one place.  I am a fan of creating variables when a name helps make the code more readable though, and it is possible that you were running into stack size issues which is why these were extracted.  In some cases, I feel like it is obvious and the name doesn't add value, in which case I would inline (like `require(numOutcomes >= 2)` already conveys that 2 is the minimum number of outcomes.  Naming the `2` doesn't really add any value IMO.

src/reporting/Market.sol
```diff
@@ -0,0 +1,653 @@
...
+    function initialize(ReportingWindow _reportingWindow, uint256 _endTime, uint8 _numOutcomes, uint256 _payoutDenominator, uint256 _feePerEthInAttoeth, Cash _denominationToken, address _creator, int256 _minDisplayPrice, int256 _maxDisplayPrice, address _automatedReporterAddress, bytes32 _topic) public payable beforeInitialized returns (bool _success) {
+        endInitialization();
+        require(address(_reportingWindow) != NULL_ADDRESS);
+        require(_payoutDenominator >= MIN_PAYOUT_DENOMINATOR);
+        require(_numOutcomes >= MIN_NUM_OUTCOMES);
+        // payoutDenominator must be a multiple of numOutcomes so we can evenly split complete set share payout on indeterminate
+        require((_payoutDenominator % _numOutcomes) == 0);
+        require(feePerEthInAttoeth <= MAX_FEE_PER_ETH_IN_ATTOETH);
+        require((DISPLAY_PRICE_MIN <= _maxDisplayPrice) && (_maxDisplayPrice <= DISPLAY_PRICE_MAX));
+        require((DISPLAY_PRICE_MIN <= _minDisplayPrice) && (_minDisplayPrice <= DISPLAY_PRICE_MAX));
```
MicahZoltu: I don't believe these two are necessary anymore now that we have real types.  display price can be any int256 but we had the check previously I think out of habit because Serpent didn't have types so everything was confusing around the edges (this particular check probably wasn't necessary in Serpent either).

src/reporting/Market.sol
```diff
@@ -0,0 +1,653 @@
...
+        require((1 <= uint256(_maxDisplayPrice.sub(_minDisplayPrice))) && (uint256(_maxDisplayPrice.sub(_minDisplayPrice)) <= SET_COST_MULTIPLIER_MAX));
```
MicahZoltu: This check can probably be split into multiple checks for simplicity.  If we have space on the stack, I recommend an intermediate variable as well:
```
require(_minDisplayPrice < _maxDisplayPrice);
uint256 _completeSetCost = _maxDisplayPrice - minDisplayPrice;
require(1 <= _completeSetCost && _completeSetCost <= 2^254);
```
Also, you can drop the check for `<=2^254`.  That was just to deal with the fact that Serpent didn't have unsigned ints.  In general, anywhere we are checking to see if a number is less than 2^254 can be deleted and you should just use `uint256` instead.  Anywhere we check that a number is >= 0 can similarly just be done with types.

MicahZoltu: Actually, I think you can just have the one check `require(_minDisplayPrice < _maxDisplayPrice)` now that we have a type system, since between that and uint256, the following require can never fail.

src/reporting/Market.sol
```diff
@@ -0,0 +1,653 @@
...
+    DisputeBondToken private limitedReportersDisputeBondToken;
+    DisputeBondToken private allReportersDisputeBondToken;
+    uint256 private validityBondAttoeth;
+    uint256 private automatedReporterBondAttoeth;
+
+    function initialize(ReportingWindow _reportingWindow, uint256 _endTime, uint8 _numOutcomes, uint256 _payoutDenominator, uint256 _feePerEthInAttoeth, Cash _denominationToken, address _creator, int256 _minDisplayPrice, int256 _maxDisplayPrice, address _automatedReporterAddress, bytes32 _topic) public payable beforeInitialized returns (bool _success) {
+        endInitialization();
+        require(address(_reportingWindow) != NULL_ADDRESS);
+        require(_payoutDenominator >= MIN_PAYOUT_DENOMINATOR);
+        require(_numOutcomes >= MIN_NUM_OUTCOMES);
```
MicahZoltu: I don't see a check for `require(_numOutcomes <= 8)`, which I _thought_ was in the old contracts (if not it should have been).

src/reporting/Market.sol
```diff
@@ -0,0 +1,653 @@
...
+
+        // TODO: we need to update this signature (and all of the places that call it) to allow the creator (UI) to pass in a number of other things which will all be logged here
+        // TODO: log short description
+        // TODO: log long description
+        // TODO: log min display price
+        // TODO: log max display price
+        // TODO: log tags (0-2)
+        // TODO: log outcome labels (same number as numOutcomes)
+        // TODO: log type (scalar, binary, categorical)
+        // TODO: log any immutable data associated with the market (e.g., endTime, numOutcomes, payoutDenominator, denominationToken address, etc.)
```
MicahZoltu: These comments should go inside of the `initialize` function I think, or at least have their indentation fixed.

src/reporting/Market.sol
```diff
@@ -0,0 +1,653 @@
...
+
+    // this will need to be called manually for each open market if a spender contract is updated
+    function approveSpenders() private returns (bool) {
+        bytes32[5] memory _names = [bytes32("cancelOrder"), bytes32("completeSets"), bytes32("takeOrder"), bytes32("tradingEscapeHatch"), bytes32("claimProceeds")];
+        uint8 i = 0;
+        for (i = 0; i < 5; i++) {
```
MicahZoltu: Recommend looping to `_names.length()`  rather than `5` for clarity.

src/reporting/Market.sol
```diff
@@ -0,0 +1,653 @@
...
+        // TODO: log max display price
+        // TODO: log tags (0-2)
+        // TODO: log outcome labels (same number as numOutcomes)
+        // TODO: log type (scalar, binary, categorical)
+        // TODO: log any immutable data associated with the market (e.g., endTime, numOutcomes, payoutDenominator, denominationToken address, etc.)
+
+    // this will need to be called manually for each open market if a spender contract is updated
+    function approveSpenders() private returns (bool) {
+        bytes32[5] memory _names = [bytes32("cancelOrder"), bytes32("completeSets"), bytes32("takeOrder"), bytes32("tradingEscapeHatch"), bytes32("claimProceeds")];
+        uint8 i = 0;
```
MicahZoltu: Style: I'm guessing you did this because Solidity cries about re-using the variable name, despite the scope difference (if that is the case I encourage you to submit a bug to Solidity since `for(uint8 i = 0; i < x; i++) { }` should scope `i` to inside of the for loop)?  Assuming that is the case, I think I would personally prefer to just have one loop use `i` and the other use `j` than declare it up here.  Doing it this way gave me pause, but `i` and `j` wouldn't have.

src/reporting/Market.sol
nuevoalex: Sounds reasonable to me, will do

nuevoalex: Interestingly this increases contract size considerably. For now I'll add it selectively to non-constant functions where possible given that.

src/reporting/Market.sol
nuevoalex: Not sure why I removed that. Adding back.

src/reporting/Market.sol
nuevoalex: I'll submit a bug and use the i,j,k pattern going forward.

src/reporting/Market.sol
MicahZoltu: Eh, if it has a big impact on contract size I would say lets leave it out of Market.  We already have enough contract size problems with this one as it is, no need to add this since it isn't going to provide any significant value (other than consistency).

joeykrug:
As a side note, lemme know when this is ready for review by me

src/reporting/Market.sol
```diff
@@ -0,0 +1,653 @@
...
+            shareTokens[i].approve(controller.lookup("takeOrder"), APPROVAL_AMOUNT);
+        }
+        return true;
+    }
+
+    function changeCreator(address _newCreator) public returns (bool) {
+        require(msg.sender == creator);
+        creator = _newCreator;
+        return true;
+    }
```
MicahZoltu: I wonder if we should change this contract to inherit from `Ownable` and change creator to owner everywhere in here.  This would allow us to use the `onlyOwner` modifier on this method and the other one(s) that need access control, and it would help get a little bit of code out of this contract (I recognize this won't help with contract size, but it will help with readability).

Note: Such a change would also require the factory to change the owner as part of contract initialization.

nuevoalex: That seems like a very reasonable change to me. Will do

src/reporting/Market.sol
```diff
@@ -0,0 +1,653 @@
...
+        require(isInAutomatedDisputePhase());
+        automatedReporterDisputeBondToken = DisputeBondTokenFactory(controller.lookup("DisputeBondTokenFactory")).createDisputeBondToken(controller, this, msg.sender, AUTOMATED_REPORTER_DISPUTE_BOND_AMOUNT, tentativeWinningPayoutDistributionHash);
+        reportingState = ReportingState.LIMITED;
+        fundDisputeBondWithReputation(msg.sender, automatedReporterDisputeBondToken, AUTOMATED_REPORTER_DISPUTE_BOND_AMOUNT);
+        reportingWindow.updateMarketPhase();
+        return true;
+    }
+
+    function disputeLimitedReporters() public returns (bool) {
+        migrateThroughAllForks();
```
MicahZoltu: I wonder if it would improve readability to have a modifier for `triggersMigration` that does `migrateThroughAllForks` rather than having it be the first line of a bunch of different functions?

There _may_ be similar readability benefits for things like `onlyInPhase(X)` (we would want an enum for all the possible phases in this case).

nuevoalex: The `triggersMigration` change sounds good so I'll work on that.

Having an enum for all possible phases is a little tricky since there is a dependence on timing, so the state may be incorrect if enough time has passed. That said we could use the limited enum and apply it in the suggested way with some additional checks if we still think it's worth it.

src/reporting/Market.sol
```diff
@@ -0,0 +1,653 @@
...
+
+    enum ReportingState {
+        AUTOMATED,
+        LIMITED,
+        ALL,
+        FORK,
+        FINALIZED
+    }
+
+    ReportingState private reportingState = ReportingState.AUTOMATED;
```
MicahZoltu: Unfortunately, I believe that this won't work correctly.  The most immediate problem is that automated reporting is an optional phase and the market automatically switches from automated reporting phase to limited reporting phase based on `block.timestamp`, not based on a transaction trigger.  Because contracts don't support any kind of alarm clock/trigger system we need to dynamically check what state the contract is in based on what the current `block.timestamp` is.

nuevoalex: The code does try to accommodate that bit by in some cases actually checking for `> LIMITED` rather than `==`. Additionally we usually use methods that ask if we're in a particular phase rather than just providing a getter to the state enum value. That said I understand this state value is confusing since it may be incorrect. I could rename the first value to reflect that it is somewhat ambiguous?

src/reporting/Market.sol
```diff
@@ -0,0 +1,653 @@
...
+    address private constant NULL_ADDRESS = address(0);
+
+    ReportingWindow private reportingWindow;
+    uint256 private endTime;
+    uint8 private numOutcomes;
+    uint256 private marketCreationBlock;
+    bytes32 private topic;
+    address private automatedReporterAddress;
+    mapping(bytes32 => ReportingToken) private reportingTokens;
+    Cash private denominationToken;
```
MicahZoltu: Suggest renaming this `cash` since that is now what it is (same comment all other places this variable is used).

src/reporting/Market.sol
```diff
@@ -0,0 +1,653 @@
...
+        marketCreationBlock = block.number;
+        topic = _topic;
+        automatedReporterAddress = _automatedReporterAddress;
+        denominationToken = _denominationToken;
+        creator = _creator;
+        for (uint8 _outcome = 0; _outcome < numOutcomes; _outcome++) {
+            shareTokens.push(createShareToken(_outcome));
+        }
+        approveSpenders();
+        require(controller.lookup("Cash") == address(denominationToken) || getBranch().isContainerForShareToken(denominationToken));
```
MicahZoltu: This can be deleted since we are removing support for combinatorial markets for launch.

src/extensions/MarketExtensions.sol
```diff
@@ -0,0 +1,27 @@
...
+        if (address(_winningDestination) == address(0)) {
+            return 0;
+        }
+        uint256 _halfTotalSupply = 11 * 10**6 * 10**18 / 2;
+        if (_winningDestination.totalSupply() < _halfTotalSupply && block.timestamp < _reportingWindow.getBranch().getForkEndTime()) {
+            return 0;
+        }
+        return _winningDestination.getBranch().getParentPayoutDistributionHash();
+    }
+}
```
adrake33: Are we still continuing the style of ending every file with an empty line like we did in Serpent?

src/reporting/ReportingWindow.sol
```diff
@@ -44,9 +46,9 @@ contract ReportingWindow is DelegationTarget, Typed, Initializable {
         return true;
     }
 
-    function createNewMarket(uint256 _endTime, uint256 _numOutcomes, uint256 _payoutDenominator, int256 _feePerEthInWei, address _denominationToken, address _creator, int256 _minDisplayPrice, int256 _maxDisplayPrice, address _automatedReporterAddress, int256 _topic) public payable returns (IMarket _newMarket) {
-        require(MINIMUM_OUTCOMES <= _numOutcomes && _numOutcomes <= MAXIMUM_OUTCOMES);
-        require(MINIMUM_PAYOUT_DENOMINATOR <= _payoutDenominator);
```
adrake33: Should these constants be removed from this contract since it looks like they aren't being used anymore?

src/extensions/MarketExtensions.sol
nuevoalex: I think we are, my bad. Will fix.

MicahZoltu:
I recommend rebasing onto or merging into develop.

src/reporting/Market.sol
```diff
@@ -0,0 +1,604 @@
...
+
+    enum ReportingState {
+        PRE_REPORTING,
+        AUTOMATED_REPORTING,
+        AUTOMATED_DISPUTE,
+        LIMITED_REPORTING,
+        LIMITED_DISPUTE,
+        ALL_REPORTING,
+        ALL_DISPUTE,
+        FORK,
```
MicahZoltu: Consider renaming this to `FORKING` to align a bit better with the tense of the others (e.g., `REPORTING`).  Alternatively, consider changing `*_REPORTING` to `*_REPORT`.  The latter would result in more consistent tense for all, but quite as natural.  Your call with however you want to name the enum items, I tend to favor consistency but I don't care enough or have a good naming scheme to push very hard for it.  :)

MicahZoltu: We need one more phase, `AWAITING_FINALIZATION` (name TBD).  It is the state where reporting is over and the market is ready to be finalized but it has not yet been finalized.  We _could_ make it so this function finalizes when in such a state, but that would mean `getReportingPhase` can't be `constant` anymore.  :/

I think for the sake of having a nice clean constant method we should go with having a phase that indicates finalizable, but not yet finalized.

nuevoalex: I'll do `FORKING` since it sounds better for some reason to me

src/reporting/Market.sol
```diff
@@ -0,0 +1,604 @@
...
+        }
+        if (address(limitedReportersDisputeBondToken) != NULL_ADDRESS && limitedReportersDisputeBondToken.getDisputedPayoutDistributionHash() == finalPayoutDistributionHash) {
+            _reputationToken.trustedTransfer(limitedReportersDisputeBondToken, getFinalWinningReportingToken(), _reputationToken.balanceOf(limitedReportersDisputeBondToken));
+        }
+        return true;
+    }
+
+    function derivePayoutDistributionHash(uint256[] _payoutNumerators) public constant returns (bytes32) {
+        uint256 _sum = 0;
+        require(_payoutNumerators.length == numOutcomes);
```
MicahZoltu: I think we can delete this require, since the `require(_sum == payoutDenominator)` means no individual payout numerator can be greater than payout denominator.

This was needed previously to prevent someone including a negative number in _payoutNumerators, but now we have unsigned types for that.  If you do remove this, we should change the sum to `_sum = _sum.add(_payoutNumerators[i]);` to prevent overflows dodging the check.  We should generally be using the safe math function here anyway rather than trusting ourselves to correctly prevent overflow with require statements.

src/reporting/Market.sol
```diff
@@ -0,0 +1,604 @@
...
+        bytes32 _payoutDistributionHash = derivePayoutDistributionHash(_payoutNumerators);
+        ReportingToken _reportingToken = reportingTokens[_payoutDistributionHash];
+        if (address(_reportingToken) == NULL_ADDRESS) {
+            _reportingToken = ReportingTokenFactory(controller.lookup("ReportingTokenFactory")).createReportingToken(controller, this, _payoutNumerators);
+            reportingTokens[_payoutDistributionHash] = _reportingToken;
+        }
+        return _reportingToken;
+    }
+
+    function fundDisputeBondWithReputation(address _bondHolder, DisputeBondToken _disputeBondToken, uint256 _bondAmount) private returns (bool) {
```
MicahZoltu: I think we can safely move this function into `DisputeBond`.  For this to work you will need make `ReputationToken`'s `trustedTransfer` method trust the dispute token, but the groundwork for that should already be laid.  The call would move into initialize of `DisputeBond`.

All of the above will remove a function definition and 3 call sites from market, helping us get the size down.

src/reporting/Market.sol
```diff
@@ -0,0 +1,604 @@
...
+
+    function isDoneWithLimitedReporters() public constant returns (bool) {
+        return getReportingState() > ReportingState.LIMITED_REPORTING;
+    }
+
+    function isDoneWithAllReporters() public constant returns (bool) {
+        return getReportingState() > ReportingState.ALL_REPORTING;
+    }
+
+    function isFinalized() public constant returns (bool) {
```
MicahZoltu: We could also delete this function and replace it with `getReportingState`.  Though, isFinalized is called enough that there may be value in keeping it separate.

nuevoalex: For consistency I think I'll move the logic into `getReportingState` but keep this function and make it a convenience for the reason you stated.

src/reporting/Market.sol
```diff
@@ -0,0 +1,604 @@
...
+        return sha3(_payoutNumerators);
+    }
+
+    function getReportingTokenOrZeroByPayoutDistributionHash(bytes32 _payoutDistributionHash) public constant returns (ReportingToken) {
+        return reportingTokens[_payoutDistributionHash];
+    }
+
+    ////////
+    //////// Getters
+    ////////
```
MicahZoltu: So many slashes!  Recommend just:
```
//
// Getters
//
```
Also, we should be consistent with spacing, Helpers has an empty line after the comment, Getters does not.

src/reporting/Market.sol
```diff
@@ -0,0 +1,604 @@
...
+    }
+
+    function isInAllDisputePhase() public constant returns (bool) {
+        return getReportingState() == ReportingState.ALL_DISPUTE;
+    }
+
+    function isContainerForReportingToken(ReportingToken _shadyToken) public constant returns (bool) {
+        if (address(_shadyToken) == NULL_ADDRESS) {
+            return false;
+        }
```
MicahZoltu: We can remove this check, since the next line will throw in Solidity anyway (test to validate, but I believe calling a method on address 0 fails in Solidity).

src/reporting/Market.sol
```diff
@@ -0,0 +1,604 @@
...
+        bytes32 _shadyId = _shadyToken.getPayoutDistributionHash();
+        ReportingToken _reportingToken = reportingTokens[_shadyId];
+        if (address(_reportingToken) == NULL_ADDRESS) {
+            return false;
+        }
+        if (_reportingToken != _shadyToken) {
+            return false;
+        }
+        return true;
```
MicahZoltu: Can simplify last two to: `return _reportingToken == _shadyToken`

src/reporting/Market.sol
```diff
@@ -0,0 +1,604 @@
...
+    }
+
+    function isContainerForShareToken(IShareToken _shadyShareToken) public constant returns (bool) {
+        if (_shadyShareToken.getTypeName() != "ShareToken") {
+            return false;
+        }
+        return(getShareToken(_shadyShareToken.getOutcome()) == _shadyShareToken);
```
MicahZoltu: Remove parenthesis around return expression.

src/reporting/Market.sol
```diff
@@ -0,0 +1,604 @@
...
+    }
+
+    function isContainerForDisputeBondToken(DisputeBondToken _shadyBondToken) public constant returns (bool) {
+        if (_shadyBondToken.getTypeName() != "DisputeBondToken") {
+            return false;
+        }
```
MicahZoltu: The type check here is unnecessary since we won't be calling any methods on the contract.

src/reporting/Market.sol
```diff
@@ -0,0 +1,604 @@
...
+        if (automatedReporterDisputeBondToken == _shadyBondToken) {
+            return true;
+        } else if (limitedReportersDisputeBondToken == _shadyBondToken) {
+            return true;
+        } else if (allReportersDisputeBondToken == _shadyBondToken) {
+            return true;
+        }
```
MicahZoltu: Style: Recommend just `if(...) { return ... }` rather than `else if`.  They are functionally the same when a return is involved, but I find clear guard clauses to be slightly more readable.

MicahZoltu:
Make sure to delete `market.se` sometime before this gets merged.

src/reporting/Market.sol
```diff
@@ -0,0 +1,604 @@
...
+        return uint256(maxDisplayPrice.sub(minDisplayPrice));
+    }
+
+    function getTopic() public constant returns (bytes32) {
+        return topic;
+    }
+
+    function shouldCollectReportingFees() public constant returns (bool) {
+        return !getBranch().isContainerForShareToken(cash);
+    }
```
MicahZoltu: With the removal of combinatorial markets for MVP, we can remove this function (you'll need to adjust claimFees as well).  Someone else is in the middle of migrating ClaimFees, so you may want to just make a comment to remove this function in the future and have it hard-coded to return true for the time being.

src/reporting/Market.sol
```diff
@@ -0,0 +1,604 @@
...
+
+    function getAutomatedReportDueTimestamp() public constant returns (uint256) {
+        return endTime + AUTOMATED_REPORTING_DURATION_SECONDS;
+    }
+
+    function getAutomatedReportDisputeDueTimestamp() public constant returns (uint256) {
+        return getAutomatedReportDueTimestamp() + AUTOMATED_REPORTING_DISPUTE_DURATION_SECONDS;
+    }
+
+    function getReportingState() public constant returns (ReportingState) {
```
MicahZoltu: I wonder if this should have a check for `needsMigration`  in it that causes it to return... something different than it would otherwise.  We could add a new state for `AWAITING_MIGRATION` (a market can be in this state for up to a couple months).  During this time no reporting on the market happens and it is effectively in limbo.  Once migration occurs, the state should drop back to LIMITED_REPORTING (will happen automatically if the migration code is correct).

MicahZoltu: As far as order goes, AWAITING_MIGRATION probably should live between automated reporting and limited reporting?  A market won't migrate until automated reporting/dispute are over and the market would otherwise enter LIMITED_REPORTING.

nuevoalex: Is the thought also that this would make us no longer need to do implicit migration when we try to finalize?

src/reporting/Market.sol
nuevoalex: I ended up just removing the function which did cut down size a bit since it has to come after the dispute bond creation.

src/factories/MarketFactory.sol
```diff
@@ -2,15 +2,16 @@ pragma solidity ^0.4.13;
...
+import 'ROOT/reporting/Market.sol';
+import 'ROOT/trading/Cash.sol';
 import 'ROOT/Controller.sol';
 
 
 contract MarketFactory {
-    function createMarket(Controller _controller, ReportingWindow _reportingWindow, uint256 _endTime, uint256 _numOutcomes, uint256 _payoutDenominator, int256 _feePerEthInWei, address _denominationToken, address _creator, int256 _minDisplayPrice, int256 _maxDisplayPrice, address _automatedReporterAddress, int256 _topic) payable returns (IMarket _market) {
-        Delegator _delegator = new Delegator(_controller, "market");
-        _market = IMarket(_delegator);
+    function createMarket(Controller _controller, ReportingWindow _reportingWindow, uint256 _endTime, uint8 _numOutcomes, uint256 _payoutDenominator, uint256 _feePerEthInWei, Cash _denominationToken, address _creator, int256 _minDisplayPrice, int256 _maxDisplayPrice, address _automatedReporterAddress, bytes32 _topic) payable returns (Market _market) {
```
adrake33: This should be explicitly declared as public.

src/factories/ReportingTokenFactory.sol
```diff
@@ -2,10 +2,11 @@ pragma solidity ^0.4.13;
 
 import 'ROOT/libraries/Delegator.sol';
 import 'ROOT/reporting/ReportingToken.sol';
+import 'ROOT/reporting/Market.sol';
 
 
 contract ReportingTokenFactory {
-    function createReportingToken(Controller _controller, IMarket _market, int256[] _payoutNumerators) returns (ReportingToken) {
+    function createReportingToken(Controller _controller, Market _market, uint256[] _payoutNumerators) returns (ReportingToken) {
```
adrake33: This should also be declared as public.

adrake33: Also ShareTokenFactory.createShareToken(), Branch.getReportingWindowForForkEndTime(), and 3 of the functions in ReportingToken.sol.

src/reporting/Market.sol
MicahZoltu: Possibly, we'll need to think through all of the scenarios and make sure all of the code paths either do the migration themselves, or function properly without the migration.

src/reporting/Market.sol
```diff
@@ -0,0 +1,593 @@
...
+        if (isFinalized()) {
+            return ReportingState.FINALIZED;
+        }
+
+        if (needsMigration()) {
+            return ReportingState.AWAITING_MIGRATION;
+        }
```
MicahZoltu: I wonder if we should just integrate `needsMigration` into this method instead of having it be a separate method?  More specifically, I wonder if up in `migrateThroughOneFork` we should call `getReportingState` and compare it against `AWAITING_MIGRATION` instead of calling `needsMigration` directly.  This would make `needsMigration` just an implementation detail of `getReportingState` rather than something that needs to stand on its own.

nuevoalex: I like that. Will do.

src/reporting/Market.sol
```diff
@@ -0,0 +1,593 @@
...
+
+        if (address(allReportersDisputeBondToken) != NULL_ADDRESS) {
```
MicahZoltu: I _think_ it may be safer to check `getBranch().getForkingMarket() == this` to determine if this is a forking market.  The check you have here I believe is correct, but I'm less confident that it will remain an authoritative source of "are we forking" than the branch's forkingMarket state member.

src/reporting/Market.sol
```diff
@@ -0,0 +1,593 @@
...
+    function getAutomatedReportDisputeDueTimestamp() public constant returns (uint256) {
+        return getAutomatedReportDueTimestamp() + AUTOMATED_REPORTING_DISPUTE_DURATION_SECONDS;
+    }
+
+    function getReportingState() public constant returns (ReportingState) {
+        if (isFinalized()) {
+            return ReportingState.FINALIZED;
+        }
+
+        if (needsMigration()) {
```
MicahZoltu: Consider an early exit (to save on gas) for if the market end timestamp is less than block.timestamp.  At the moment, any calls to this before the market has ended will go through all of the checks before returning when that very cheap check could have caused an early bail.

This change is less important if the trading code paths never check the state.  If any of the trading code paths result in a state check though, we definitely should try to exit early for pre-reporting.

src/reporting/Market.sol
```diff
@@ -0,0 +1,593 @@
...
+                return ReportingState.ALL_DISPUTE;
+            }
+            return ReportingState.ALL_REPORTING;
+        }
+
+        if (reportingWindow.isDisputeActive()) {
+            return ReportingState.LIMITED_DISPUTE;
+        }
+
+        if (canBeReportedOn()) {
```
MicahZoltu: I think this would be more clear (given context) if it just did `if (!reportingWindow.isReportingActive())` instead.  This would align better with the check above this, and also it would avoid the excess `isFinalized` check done in `canBeReportedOn`.

src/reporting/Market.sol
```diff
@@ -0,0 +1,593 @@
...
+            return ReportingState.LIMITED_REPORTING;
+        }
+
+        if (block.timestamp > getAutomatedReportDisputeDueTimestamp() && (canFinalizeReporting() || canFinalizeAutomated())) {
```
MicahZoltu: We check `canFinalizeReporting` a couple times throughout this process.  I think it may simplify things if we just did a `canFinalizeReporting` check up front (perhaps just after checking if the market is already finalized).  This check is quite cheap, so it makes for a good early exit.

There may actually be value in removing the `canFinalizeReporting` method and directly checking the timestamp against reporting window end time here.  If you do this, you may need to switch some callsites over to use the new `getReportingState` function instead of `canFinalizeReporting`.

src/reporting/Market.sol
```diff
@@ -0,0 +1,593 @@
...
+        // FIXME: if automated report is right, transfer automated report bond to market creator
+        // FIXME: handle markets that get 0 reports during their scheduled reporting window
+    }
+
+    function canFinalizeAutomated() private returns (bool) {
+        if (!automatedReportReceived) {
+            return false;
+        }
+        return true;
+    }
```
MicahZoltu: This may have been a pre-existig bug, but this needs to check more than just `automatedReportReceived`.  At the least, it needs to verify no dispute bond was placed and it is after the dispute period.  As with several other methods in this contract, consider removing it and replacing its callsites with `getReportingState`.

As you may be catching on to, now that we have a consolidated "Reporting State" system I'm a big fan of moving more things over to it rather than having all of these one-off methods that do some subset of state checks.  There are some minor performance advantages to checking only a subset of the state, but I think we should opt for cleaner code at the moment and performance optimize later if necessary.

nuevoalex: Yea that makes sense and will look a lot clearer.

nuevoalex:
I definitely like that pattern. Since we're already refactoring a bit we may as well shift full on to it. Will work on that tomorrow.

Given the other incoming PRs and potential bottlenecks with parallelizable work on the conversion I may spend some time this week writing some of those tests too.

PR #269, opened 2017-08-23
Orders & Orders Fetcher Solidity migration

adrake33:
﻿# Pull Request Template
### Overview

Migration of orders.se and ordersFetcher.se to Solidity.  NewOrders.sol was modified to use one-dimensional arrays instead of four-dimensional arrays.

### Examples (if necessary)


src/trading/NewOrders.sol
```diff
@@ -0,0 +1,342 @@
...
+    }
+
+    struct MarketOrders {
+        uint256 volume;
+        uint256[] prices;
+    }
+
+    // Trade types
+    // FIXME: Ideally, this should probably be an enum
+    uint8 private constant BID = 1;
```
nuevoalex: These could be made an enum now I think

src/trading/NewOrders.sol
```diff
@@ -0,0 +1,342 @@
...
+            orders[_worseOrderId].betterOrderId = _orderId;
+            orders[_orderId].worseOrderId = _worseOrderId;
+        }
+        return true;
+    }
+
+    // FIXME: Remove unused parameters orderID and gasPrice
+    function saveOrder(address orderId, uint256 _type, IMarket _market, uint256 _fxpAmount, uint256 _fxpPrice, address _sender, uint8 _outcome, uint256 _fxpMoneyEscrowed, uint256 _fxpSharesEscrowed, bytes20 _betterOrderId, bytes20 _worseOrderId, uint256 _tradeGroupId, uint256 gasPrice) public onlyWhitelistedCallers returns (bytes20 _orderId) { 
+        require(_type == BID || _type == ASK);
+        require(0 <= _outcome || _outcome < _market.getNumberOfOutcomes());
```
nuevoalex: I think the `0 <= _outcome` isn't needed anymore since its an unsigned type

src/trading/NewOrders.sol
```diff
@@ -0,0 +1,342 @@
...
+        orders[_orderId].fxpMoneyEscrowed = 0;
+        orders[_orderId].fxpSharesEscrowed = 0;
+        orders[_orderId].gasPrice = 0;
+        return true;
+    }
+
+    function fillOrder(bytes20 _orderId, uint256 _orderType, IMarket _market, uint8 _orderOutcome, uint256 _sharesFilled, uint256 _tokensFilled) public onlyWhitelistedCallers returns (bool) { 
+        // FIXME: Should eventually be changed to `require(_market.getTypeName() == "Market")`
+        require(_market != address(0));
+        require(0 <= _orderOutcome && _orderOutcome < _market.getNumberOfOutcomes());
```
nuevoalex: same note about `0 <= _orderOutcome` w/ regard to not being needed anymore

src/trading/NewOrders.sol
adrake33: Yep, I'll remove it.

src/trading/NewOrdersFetcher.sol
```diff
@@ -0,0 +1,149 @@
...
+        _betterOrderId = _orders.getBetterOrderId(_orderId, _type, _market, _outcome);
+        _worseOrderId = _orders.getWorseOrderId(_orderId, _type, _market, _outcome);
+        _gasPrice = _orders.getGasPrice(_orderId, _type, _market, _outcome);
+        return (_attoshares, _displayPrice, _owner, _tokensEscrowed, _sharesEscrowed, _betterOrderId, _worseOrderId, _gasPrice);
+    }
+
+    function ascendOrderList(uint256 _type, IMarket _market, uint8 _outcome, uint256 _fxpPrice, bytes20 _lowestOrderId) public constant returns (bytes20 _betterOrderId, bytes20 _worseOrderId) {
+        _worseOrderId = _lowestOrderId;
+        bool _isWorstPrice;
+        var _orders = NewOrders(controller.lookup("NewOrders"));
```
nuevoalex: I think we decided to avoid `var`

src/trading/NewOrders.sol
adrake33: I suppose this enum should probably be put in a separate library, since most of the trading contracts use these constants.

src/libraries/Trading.sol
```diff
@@ -0,0 +1,11 @@
+/**
+ * Copyright (C) 2015 Forecast Foundation OU, full GPL notice in LICENSE
+ */
+
+pragma solidity ^0.4.13;
+
+library Trading {
```
MicahZoltu: Hmm, I wonder how this actually works... I see in other files you are importing it and then doing `Trading.Bid` and `Trading.Ask`, but I thought libraries in Solidity needed to be linked in via an external address at compile time?  Maybe that doesn't apply for nested types like Enums?

src/libraries/math/SafeMathUint256.sol
```diff
@@ -45,4 +44,22 @@ library SafeMathUint256 {
             return b;
         }
     }
+
+    function safeToMultiply(uint256 a, uint256 b) internal constant returns (bool) {
```
MicahZoltu: Curious: what made you decide to extract this out into a separate method rather than inlining into `fxpMul`?

src/libraries/math/SafeMathUint256.sol
```diff
@@ -45,4 +44,22 @@ library SafeMathUint256 {
...
+        uint256 c = a * b;
+        if (a == 0 || c / a == b) {
+            return true;
+        }
+        return false;
+    }
+
+    // Float [fixed point] Operations
+    function fxpMul(uint256 a, uint256 b) internal constant returns (uint256) {
```
MicahZoltu: Given the name (`fxpMul`) this should _probably_ include a base parameter.  I have argued against fixed point math in the past and lost, so I won't try to convince you to remove it.  ;)

At the moment, this implementation is 10^18 fixed point multiply, yet the name doesn't convey that it is 10^18 fixed point.  Recommend either changing the name to be more clear, or adding a base parameter.

src/libraries/math/SafeMathUint256.sol
```diff
@@ -45,4 +44,22 @@ library SafeMathUint256 {
...
+        require(safeToMultiply(a, b));
```
MicahZoltu: With the `safeToMultiply` check extracted out, this ends up doing the arithmetic twice.  Consider merging in that method so we don't have to dot he math twice.  Though, if you have a compelling reason for keeping it extracted the argument for gas savings is pretty insignificant.

src/libraries/math/SafeMathUint256.sol
```diff
@@ -45,4 +44,22 @@ library SafeMathUint256 {
...
+        return a * b / 10 ** 18;
+    }
+
+    function fxpDiv(uint256 a, uint256 b) internal constant returns (uint256) {
+        return a * 10 ** 18 / b;
```
MicahZoltu: This needs some checks to make sure it doesn't overflow.  `a * 10^18` could overflow, resulting in an incorrect number being divided by.  Probably easiest to change this to `return div(mul(a, 10**18), b)`.

src/libraries/math/SafeMathUint256.sol
MicahZoltu: Consider just changing this to `div(mul(a, b), 10**18)`

src/libraries/Trading.sol
adrake33: Good question -- I'm actually not sure either.  Do you think this should be done in a different way?

src/trading/NewOrdersFetcher.sol
```diff
@@ -0,0 +1,146 @@
...
+import 'ROOT/libraries/arrays/Bytes20Arrays.sol';
+import 'ROOT/libraries/Trading.sol';
+import 'ROOT/reporting/Interfaces.sol';
+import 'ROOT/trading/NewOrders.sol';
+
+
+/**
+ * @title OrdersFetcher
+ */
+// FIXME: This was named NewOrdersFetcher to prevent naming conflicts with ordersFetcher.se. Should be renamed to OrdersFetcher.
```
MicahZoltu: Don't forget to apply this fixme before merging into develop.

src/libraries/Trading.sol
MicahZoltu: No, if this works I'm generally happy with it.  I'm just surprised it works.

src/trading/NewOrdersFetcher.sol
```diff
@@ -0,0 +1,146 @@
...
+contract NewOrdersFetcher is Controlled {
+    using Bytes20Arrays for bytes20[];
+
+    event Log(uint256 fxpPrice);
```
MicahZoltu: I'm guessing this was added for testing?  Recommend removing if so, if not it is currently unused and should either be removed or used.

src/trading/NewOrdersFetcher.sol
```diff
@@ -0,0 +1,146 @@
...
+        require(_type == Trading.TradeTypes.Bid || _type == Trading.TradeTypes.Ask);
+        require(0 <= _outcome && _outcome < _market.getNumberOfOutcomes());
+        NewOrders _orders = NewOrders(controller.lookup("NewOrders"));
+        if (_startingOrderId == 0) {
+            _startingOrderId = _orders.getBestOrderId(_type, _market, _outcome);
+        }
+        bytes20[] _orderIds;
+        _orderIds[0] = _startingOrderId;
+        uint256 _i = 0;
+        while (_i < _numOrdersToLoad && _orders.getWorseOrderId(_orderIds[_i], _type, _market, _outcome) != 0) {
```
MicahZoltu: Change this to a for loop, since that is really what it is.  the `_orders.getWorseOrderId(...) != 0` can be changed to a `if (...) { break; }` statement inside the loop.

src/libraries/math/SafeMathUint256.sol
adrake33: I did it because the mul() and fxpMul() functions were both using this same logic, and this is how it was being done in safeMath.sem.  I realize it's kind of awkward, though.  I'd be ok with putting it inline in these functions instead.

src/trading/NewOrdersFetcher.sol
```diff
@@ -0,0 +1,146 @@
...
+// FIXME: This was named NewOrdersFetcher to prevent naming conflicts with ordersFetcher.se. Should be renamed to OrdersFetcher.
+contract NewOrdersFetcher is Controlled {
+    using Bytes20Arrays for bytes20[];
+
+    event Log(uint256 fxpPrice);
+    
+    /**
+     * @dev Get orders for a particular market, type, and outcome (chunked)
+     */
+    function getOrderIds(Trading.TradeTypes _type, IMarket _market, uint8 _outcome, bytes20 _startingOrderId, uint256 _numOrdersToLoad) public constant returns (bytes20[]) {
```
MicahZoltu: Change from `public` to `external` since this is not intended to ever be called internally.

src/libraries/math/SafeMathUint256.sol
adrake33: Ok, I agree that adding the base would make it clearer.  Maybe call it fxpMulBase18() instead?

src/libraries/math/SafeMathUint256.sol
adrake33: The main reason I did it was because the logic was shared between mul() and fxpMul(), but I think the savings in gas is a much more compelling reason to do the logic inline.

src/libraries/math/SafeMathUint256.sol
MicahZoltu: I'm not sure, the actual base is `10^18` (different from `2^18` base for example).  This is part of why I'm generally against fixed point math, because doing it well is hard.  :/  I'll leave it up to you to make the final call of leaving it as is, changing the name (and if so, what to change it to), or adding a base parameter.  Good luck!

src/trading/NewOrders.sol
```diff
@@ -0,0 +1,338 @@
...
+import 'ROOT/libraries/Trading.sol';
+import 'ROOT/reporting/Interfaces.sol';
+import 'ROOT/trading/NewOrdersFetcher.sol';
+
+
+/**
+ * @title Orders
+ * @dev Storage of all data associated with orders
+ */
+// FIXME: This was named NewOrders to prevent naming conflicts with orders.se. Should be renamed to Orders.
```
MicahZoltu: Be sure to apply fixme before merge.

src/trading/NewOrders.sol
```diff
@@ -0,0 +1,338 @@
...
+
+    struct Order {
+        uint256 fxpAmount;
+        uint256 fxpPrice;
+        address owner;
+        uint256 fxpSharesEscrowed;
+        uint256 fxpMoneyEscrowed;
+        bytes20 betterOrderId;
+        bytes20 worseOrderId;
+        uint256 gasPrice; // FIXME: Remove deprecated gasPrice
```
MicahZoltu: I wonder if we can safely remove this now, or would that require touching a bunch of other files?

src/trading/NewOrders.sol
```diff
@@ -0,0 +1,338 @@
...
+ * @dev Storage of all data associated with orders
+ */
+// FIXME: This was named NewOrders to prevent naming conflicts with orders.se. Should be renamed to Orders.
+contract NewOrders is Controlled {
+    using SafeMathUint256 for uint256;
+
+    event CancelOrder(address indexed market, address indexed sender, uint256 fxpPrice, uint256 fxpAmount, bytes20 orderId, uint8 outcome, Trading.TradeTypes orderType, uint256 cashRefund, uint256 sharesRefund);
+    event CompleteSets(address indexed sender, address indexed market, Trading.TradeTypes indexed orderType, uint256 fxpAmount, uint256 numOutcomes, uint256 marketCreatorFee, uint256 reportingFee);
+    event MakeOrder(address indexed market, address indexed sender, Trading.TradeTypes indexed orderType, uint256 fxpPrice, uint256 fxpAmount, uint8 outcome, bytes20 orderId, uint256 fxpMoneyEscrowed, uint256 fxpSharesEscrowed, uint256 tradeGroupId);
+    event TakeOrder(address indexed market, uint256 indexed outcome, Trading.TradeTypes indexed orderType, bytes20 orderId, uint256 price, address maker, address taker, uint256 makerShares, uint256 makerTokens, uint256 takerShares, uint256 takerTokens, uint256 tradeGroupId);
```
MicahZoltu: `outcome` is `uint8` now I believe.

src/trading/NewOrdersFetcher.sol
adrake33: I haven't updated all of the other test scripts to use the Solidity version of Orders/OrdersFetcher (only test_orders.py), so that's why I was using this naming convention.  Should I go ahead and update the other test scripts as well before merging to develop?

src/trading/NewOrdersFetcher.sol
adrake33: Ah, good catch.  I'll remove that.

src/trading/NewOrdersFetcher.sol
MicahZoltu: Yeah, this the goal for merging this PR is `orders.se` is deleted and `Orders.sol` fully replaces it (meaning all tests pass using Orders.sol instead of orders.se).

src/trading/NewOrders.sol
adrake33: I can remove it from where it's being used in NewOrders.sol, but removing it as an input param from functions might affect a fair number of other files.

MicahZoltu: I think leaving it as a parameter for now is fine if we don't need to remove it for stack depth reasons.  I think it would still be good to remove it from this struct and usage in all functions (internally).

src/trading/NewOrders.sol
```diff
@@ -0,0 +1,338 @@
...
+    function getWorstOrderId(Trading.TradeTypes _type, IMarket _market, uint8 _outcome) public constant returns (bytes20) { 
+        return worstOrder[getBestOrderWorstOrderHash(_market, _outcome, _type)];
+    }
+
+    // FIXME: Currently, getOrderIdHash(), getBestOrderWorstOrderHash(), and makeOrder.makeOrder() are creating ripemd160 hash values of type bytes20.  This should probably be changed to create sha3/sha256 hash values of type bytes32.
+    function getOrderIdHash(Trading.TradeTypes _type, IMarket _market, uint256 _fxpAmount, uint256 _fxpPrice, address _sender, uint256 _blockNumber, uint256 _outcome, uint256 _fxpMoneyEscrowed, uint256 _fxpSharesEscrowed) public constant returns (bytes20) {
+        return ripemd160(_type, _market, _fxpAmount, _fxpPrice, _sender, _blockNumber, _outcome, _fxpMoneyEscrowed, _fxpSharesEscrowed);
+    }
+
+    function getBlockNumber() public constant returns (uint256) {
```
MicahZoltu: This isn't used in this file, nor does it appear to be used anywhere else.  We should delete it.

src/trading/NewOrders.sol
```diff
@@ -0,0 +1,338 @@
...
+
+    function getBetterOrderId(bytes20 _orderId, Trading.TradeTypes _type, IMarket _market, uint8 _outcome) public constant returns (bytes20) { 
+        return orders[_orderId].betterOrderId;
+    }
+
+    function getWorseOrderId(bytes20 _orderId, Trading.TradeTypes _type, IMarket _market, uint8 _outcome) public constant returns (bytes20) {
+        return orders[_orderId].worseOrderId;
+    }
+
+    function getGasPrice(bytes20 _orderId, Trading.TradeTypes _type, IMarket _market, uint8 _outcome) public constant returns (uint256) {
```
MicahZoltu: This is only used in orders Fetcher, so we can safely delete it.

src/trading/NewOrders.sol
```diff
@@ -0,0 +1,338 @@
...
+        require(!isBetterPrice(_type, _market, _outcome, _fxpPrice, _betterOrderId));
+        return true;
+    }
+
+    function assertIsNotWorsePrice(Trading.TradeTypes _type, IMarket _market, uint8 _outcome, uint256 _fxpPrice, bytes20 _worseOrderId) public returns (bool) { 
+        require(!isWorsePrice(_type, _market, _outcome, _fxpPrice, _worseOrderId));
+        return true;
+    }
+
+    function insertOrderIntoList(bytes20 _orderId, Trading.TradeTypes _type, IMarket _market, uint8 _outcome, uint256 _fxpPrice, bytes20 _betterOrderId, bytes20 _worseOrderId) internal onlyWhitelistedCallers returns (bool) {
```
MicahZoltu: Functions that are `internal` don't need `onlyWhitelistCallers`.  Serpent didn't have `private` or `internal` functions so we ended up flagging such methods as `onlyWhitelistedCallers`.

In general, any method that had `assertPrivate` or `onlyWhitelistedCallers` and was never called outside of the contract should be changed to `private`.  `internal` should only be used if you are leveraging inheritance and need to expose something to a derived contract (pretty rare).  If it had `onlyWhitelistedCallers` and was called from other Augur contracts, then it should be public with the `onlyWhitelistedCallers` modifier.

src/trading/NewOrders.sol
```diff
@@ -0,0 +1,338 @@
...
+        require(orders[_orderId].fxpPrice <= _market.getMaxDisplayPrice());
+        require(orders[_orderId].fxpPrice >= _market.getMinDisplayPrice());
+        require(_market.getMaxDisplayPrice() + _market.getMinDisplayPrice() <= 2**254);
+        uint256 _fill = 0;
+        if (_orderType == Trading.TradeTypes.Bid) {
+            // We can't use safeSub here because it disallows subtracting negative numbers. Worst case here is an operation of 2**254 - 1 as required above, which won't overflow
+            _fill = _sharesFilled + (_tokensFilled.fxpDiv(orders[_orderId].fxpPrice - _market.getMinDisplayPrice()));
+        }
+        if (_orderType == Trading.TradeTypes.Ask) {
+            // We can't use safeSub here because it disallows subtracting negative numbers. Worst case here is an operation of 2**254 - 1 as required above, which won't overflow
```
MicahZoltu: At some point (doesn't have to be in this PR) we should add a `safeMathInt256` (if we haven't already) that understands how to safely subtract and add negative numbers.  Then we can delete this comment and use the safe math functions for `int256`.

For now, feel free to just update the comment to be a TODO indicating that we should make this use a safe math library that supports int256.

src/trading/NewOrders.sol
adrake33: It's used in test_orders.py for calling orders.getOrderIdHash().  I figured having the function getOrderIdHash might also be useful in some scenarios if we want to get the orderId hash without actually calling saveOrder().  If you still think I should remove it, how should I be setting the better/worse order IDs that are used for the asserts in test_orders.py?

src/trading/NewOrders.sol
adrake33: Yep, I'll change this.

src/trading/NewOrders.sol
adrake33: I recently created a SafeMathInt256.  I'll update this line to use it.

src/trading/NewOrders.sol
```diff
@@ -0,0 +1,338 @@
...
+            worstOrder[getBestOrderWorstOrderHash(_market, _outcome, _type)] = _betterOrderId;
+        }
+        if (_betterOrderId != 0) { 
+            orders[_betterOrderId].worseOrderId = _worseOrderId;
+        }
+        if (_worseOrderId != 0) { 
+            orders[_worseOrderId].betterOrderId = _betterOrderId;
+        }
+        orders[_orderId].betterOrderId = 0;
+        orders[_orderId].worseOrderId = 0;
```
MicahZoltu: `solc` gives me a warning here indicating that assigning a decimal value to `bytesXX` may not behave as expected.

src/trading/NewOrders.sol
```diff
@@ -0,0 +1,338 @@
...
+        _market.getBranch().getTopics().updatePopularity(_market.getTopic(), _fxpAmount);
+        return true;
+    }
+
+    function setPrice(IMarket _market, uint8 _outcome, uint256 _fxpPrice) internal returns (bool) { 
+        marketOrderData[_market].prices[_outcome] = _fxpPrice;
+        return true;
+    }
+
+    function removeOrderFromList(bytes20 _orderId, Trading.TradeTypes _type, IMarket _market, uint8 _outcome) private returns (bool) {
```
MicahZoltu: Nit: A lot of lines in this file have an extraneous space at the end, before the newline.

src/trading/NewOrders.sol
```diff
@@ -0,0 +1,338 @@
...
+            return (_fxpPrice < orders[_orderId].fxpPrice);
+        }
+    }
+
+    function isWorsePrice(Trading.TradeTypes _type, IMarket _market, uint8 _outcome, uint256 _fxpPrice, bytes20 _orderId) public constant returns (bool) { 
+        if (_type == Trading.TradeTypes.Bid) { 
+            return (_fxpPrice < orders[_orderId].fxpPrice);
+        } else {
+            return (_fxpPrice > orders[_orderId].fxpPrice);
+        }
```
MicahZoltu: Style: I personally prefer switch statements for things like this as it ensures that all of the enum values are handled.  For a 2-item switch it doesn't really make a huge difference, but I think it adds some clarity.  Right now this code makes it look like the second codepath should be followed if _type is anything other than Bid, which isn't correct.

If you prefer to keep it as an if statement, this function should `require(_type == Trading.TradeTypes.Bid || _type == Trading.TradeTypes.Ask)` or it should be if, else if, else throw.  We want to protect ourselves from ourselves and someone passing something other than one of those two values into this function.

src/trading/NewOrders.sol
```diff
@@ -0,0 +1,338 @@
...
+        if (_worseOrderId != 0) {
+            orders[_worseOrderId].betterOrderId = _orderId;
+            orders[_orderId].worseOrderId = _worseOrderId;
+        }
+        return true;
+    }
+
+    // FIXME: Remove unused parameters orderID and gasPrice
+    function saveOrder(address orderId, Trading.TradeTypes _type, IMarket _market, uint256 _fxpAmount, uint256 _fxpPrice, address _sender, uint8 _outcome, uint256 _fxpMoneyEscrowed, uint256 _fxpSharesEscrowed, bytes20 _betterOrderId, bytes20 _worseOrderId, uint256 _tradeGroupId, uint256 gasPrice) public onlyWhitelistedCallers returns (bytes20 _orderId) {
```
MicahZoltu: For unused parameters I recommend removing the parameter name in order to silence the checker.  `saveOrder(address, Trading.TradeTypes _type, ..., uint256)` (note the missing parameter names for `orderId` and `gasPrice`.

src/trading/NewOrders.sol
```diff
@@ -0,0 +1,338 @@
...
+        uint256[] prices;
+    }
+
+    mapping(bytes20 => Order) private orders;
+    mapping(address => MarketOrders) private marketOrderData;
+    mapping(bytes20 => bytes20) private bestOrder;
+    mapping(bytes20 => bytes20) private worstOrder;
+
+    // Getters
+    function getOrders(bytes20 _orderId, Trading.TradeTypes _type, IMarket _market, uint8 _outcome) public constant returns (uint256, uint256, address, uint256, uint256, bytes20, bytes20, uint256) {
```
MicahZoltu: Now that the Market migration has been merged in, you'll need to do a global find and replace of `IMarket` with `Market` and you'll need to `import 'ROOT/reporting/market.sol';` in any file that needs it.

src/trading/makeOrder.se
```diff
@@ -70,15 +70,16 @@ def makeOrder(sender: address, type: int256, attoshares: int256, displayPrice: i
     fxpMoneyEscrowed = escrow[0]
     fxpSharesEscrowed = escrow[1]
 
-    order = [type, market, attoshares, displayPrice, sender, block.number, outcome, fxpMoneyEscrowed, fxpSharesEscrowed, tx.gasprice]
-    orderID = ripemd160(order, items=10)
+    order = [type, market, attoshares, displayPrice, sender, block.timestamp, outcome, fxpMoneyEscrowed, fxpSharesEscrowed]
```
MicahZoltu: Changing this from `block.number` to `block.timestamp` is a pretty significant change.  I'm pretty sure this should be `block.number`.  What was the reason for the change to `block.timestamp`?

tests/trading/test_makeOrder.py
```diff
@@ -50,12 +55,12 @@ def test_setController_failure(contractsFixture):
...
     cash = contractsFixture.cash
     market = contractsFixture.binaryMarket
-    ordersFetcher = contractsFixture.contracts['ordersFetcher']
+    ordersFetcher = contractsFixture.contracts['NewOrdersFetcher']
     makeOrder = contractsFixture.contracts['makeOrder']
     cash.depositEther(value = 10**17)
     cash.approve(makeOrder.address, 10**17)
 
-    orderID = makeOrder.publicMakeOrder(BID, 10**18, 10**17, market.address, 1, 0, 0, 7)
+    orderID = makeOrder.publicMakeOrder(BID, 10**18, 10**17, market.address, 1, HASHED_ORDER_ID_0, HASHED_ORDER_ID_0, 7)
```
MicahZoltu: I believe this should be a 20 byte array of 0s, not a hash of the number `0`.  In one of the other tests you correctly have `ZEROED_ORDER_ID = '\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'`, recommend using that here as well.

MicahZoltu: Same comment throughout the test changes.  When we were passing in `0` for a better order ID or worse order ID it was meant to represent `null`, which would be a 20-byte null array.

tests/trading/test_orders.py
```diff
@@ -23,207 +29,215 @@
...
+        "betterOrderID": HASHED_ORDER_ID_0,
+        "worseOrderID": HASHED_ORDER_ID_0,
         "tradeGroupID": 0
     }
-    assert(orders.saveOrder(order["orderID"], order["type"], market.address, order["fxpAmount"], order["fxpPrice"], order["sender"], order["outcome"], order["fxpMoneyEscrowed"], order["fxpSharesEscrowed"], order["betterOrderID"], order["worseOrderID"], order["tradeGroupID"], 1) == 1), "Save order"
-    bestOrderID = int(orders.getBestOrderID(1, market.address, outcomeID), 16)
-    worstOrderID = int(orders.getWorstOrderID(1, market.address, outcomeID), 16)
-    assert(bestOrderID == 5)
-    assert(worstOrderID == 5)
+    hashedOrderID5 = orders.saveOrder(HASHED_ORDER_ID_0, order["type"], market.address, order["fxpAmount"], order["fxpPrice"], order["sender"], order["outcome"], order["fxpMoneyEscrowed"], order["fxpSharesEscrowed"], order["betterOrderID"], order["worseOrderID"], order["tradeGroupID"])
```
MicahZoltu: Recommend naming this just `orderId`.  Order IDs are always 20-byte arrays whose provenance is a hash but that isn't relevant to the reader.

tests/trading/test_orders.py
```diff
@@ -23,207 +29,215 @@
...
+
 def test_walkOrderList_bids(contractsFixture):
     market = contractsFixture.binaryMarket
-    orders = contractsFixture.contracts['orders']
-    ordersFetcher = contractsFixture.contracts['ordersFetcher']
+    orders = contractsFixture.contracts['NewOrders']
+    ordersFetcher = contractsFixture.contracts['NewOrdersFetcher']
     outcomeID = 1
     order = {
         "orderID": 5,
```
MicahZoltu: This shouldn't be `5` anymore, since `5` isn't a valid 20-byte array.  Then below we can continue doing `order["orderID"]` instead of having to reference the hard coded value.

Really all we need is a unique 20-byte array for these, we used numbers previously for simplicity but it could be a randomly generated UUID or even a 20-byte string (since Python considers strings and byte arrays to be the same).

There may be value in adding a function to `utils.py` like `longToBytes(value, length=32)` that would turn a long into a byte array padded with zeros out to `length`.  That way we could just do `longToBytes(5, 20)` here rather than having to do `\x00\x00...\x05` each time.

src/trading/makeOrder.se
adrake33: I did it to match with the removal of the getBlockNumber().  Since you said I should use the timestamp in test_orders.py, I had to use the timestamp in the contracts to get the same hash when calling orders.getOrderIdHash().  I think I had originally tried using contractsFixture.chain.head_state.block_number in test_orders.py, but it wasn't returning the same block number for some reason.

src/trading/NewOrders.sol
adrake33: Hmm, how can we set a bytes20 variable to 20-byte array of zeros?  This line in trade.se also seems to be causing an error because we're passing in zeros: `return MAKEORDER.makeOrder(sender, direction, fxpAmount, fxpPrice, market, outcome, 0, 0, tradeGroupID)`.

src/trading/makeOrder.se
MicahZoltu: `head_state.block_number` should match.  It is possible it is off by one (1 number too low).  This should be pretty easy to test and find out.  When I said you should remove `getBlockNumber()` I just meant that you should be getting the block number from pyethereum, not from the contracts.  My guess for why you were seeing different results is that Pyethereum might be displaying the last mined block, while the contracts will see the pending block (which will be one number higher).

src/trading/NewOrders.sol
MicahZoltu: You should test first, but I'm guessing for the case of `0` you could just do `bytes20(0)`.  For any other values I suspect you'll need to do something different but we should never be hard-coding non-zero byte arrays.  You should test and verify that Solidity `bytes20(0)` results in a byte array of 20 bytes, to make sure my assumptions are correct.

src/trading/NewOrdersFetcher.sol
```diff
@@ -0,0 +1,148 @@
...
+        _orderIds[0] = _startingOrderId;
+        uint256 _i = 0;
+        while (_i < _numOrdersToLoad && _orders.getWorseOrderId(_orderIds[_i], _type, _market, _outcome) != 0) {
+            _orderIds[_i + 1] = _orders.getWorseOrderId(_orderIds[_i], _type, _market, _outcome);
+            _i += 1;
+        }
+        return (_orderIds.slice(0, _i));
+    }
+
+    function getOrder(bytes20 _orderId, uint8 _type, Market _market, uint8 _outcome) public constant returns (uint256 _attoshares, int256 _displayPrice, address _owner, int256 _tokensEscrowed, uint256 _sharesEscrowed, bytes20 _betterOrderId, bytes20 _worseOrderId, uint256 _gasPrice) {
```
MicahZoltu: `tokensEscrowed` should be `uint256`.

adrake33: It's being set to orders.getOrderMoneyEscrowed(), which returns fxpMoneyEscrowed.  I thought you had said previously that this should be int256: https://augurproject.slack.com/archives/C031NRD9B/p1503637460000088.  Is this not the case?

MicahZoltu: Price should be `int256` (for the time being), but the amounts being escrowed can never be negative.

src/trading/NewOrders.sol
adrake33: I think bytes20(0) might work in Solidity, but it is causing errors in Serpent.  For example, I tried changing some lines in `trade.takeBestOrder()` from using 0 to using bytes20(0), and I'm seeing `Invalid argument count or LLL function: (bytes20 0)`

MicahZoltu: `new bytes20(0)` maybe?  I'll try to fool around with it in the sandbox and see if I can figure out how to create a `bytes20`.

src/trading/NewOrders.sol
adrake33: I don't think Solidity actually supports switch statements (http://solidity.readthedocs.io/en/develop/control-structures.html#control-structures), but I'll change the `else` to an `else if` and add the `require()` line you mentioned above.

MicahZoltu: Odd.  It appears that Solidity supports switch statements in inline assembly, but not in normal Solidity.  Very bizarre.

adrake33: Hmm, that is really weird.  Well, hopefully Solidity will add switch statements at some point in the future.

src/libraries/math/SafeMathInt256.sol
```diff
@@ -54,4 +54,22 @@ library SafeMathInt256 {
...
+    }
+
+    // Float [fixed point] Operations
+    function fxpMul(int256 a, int256 b) internal constant returns (int256) {
+        require(safeToMultiply(a, b));
+        return a * b / 10 ** 18;
+    }
+
+    function fxpDiv(int256 a, int256 b) internal constant returns (int256) {
+        return a * 10 ** 18 / b;
```
MicahZoltu: This library should be changed to match the style of `SafeMathUint256.sol` and have `fxpMul` and `fxpDiv` just be composition of `mul` and `div`.  That way we are covered from underflow/overflow issues that are not currently accounted for in these and we avoid having to extract out `safeToMultiply`.

src/reporting/Interfaces.sol
```diff
@@ -9,12 +9,15 @@ import 'ROOT/reporting/ReportingToken.sol';
 import 'ROOT/reporting/ReportingWindow.sol';
 import 'ROOT/reporting/DisputeBondToken.sol';
 import 'ROOT/reporting/RegistrationToken.sol';
-
+import 'ROOT/trading/Cash.sol';
```
MicahZoltu: This shouldn't be imported here, instead it should be imported into whichever contracts need `Cash`.  As we migrate more and more code to Solidity, this file should shrink until it is deleted.

src/trading/Orders.sol
```diff
@@ -0,0 +1,320 @@
...
+    function getBestOrderId(Trading.TradeTypes _type, Market _market, uint8 _outcome) public constant returns (bytes20) {
+        return bestOrder[getBestOrderWorstOrderHash(_market, _outcome, _type)];
+    }
+
+    function getWorstOrderId(Trading.TradeTypes _type, Market _market, uint8 _outcome) public constant returns (bytes20) {
+        return worstOrder[getBestOrderWorstOrderHash(_market, _outcome, _type)];
+    }
+
+    // FIXME: Currently, getOrderIdHash() & getBestOrderWorstOrderHash() are creating ripemd160 hash values of type bytes20.  This should probably be changed to create sha3/sha256 hash values of type bytes32.
+    function getOrderIdHash(Trading.TradeTypes _type, Market _market, uint256 _fxpAmount, int256 _fxpPrice, address _sender, uint256 _blockNumber, uint8 _outcome, uint256 _fxpMoneyEscrowed, uint256 _fxpSharesEscrowed) public constant returns (bytes20) {
```
MicahZoltu: This should be renamed to `getOrderId`.  The fact that it happens to be a hash is an implementation detail.

src/trading/makeOrder.se
```diff
@@ -70,15 +70,14 @@ def makeOrder(sender: address, type: int256, attoshares: int256, displayPrice: i
...
 
-    order = [type, market, attoshares, displayPrice, sender, block.number, outcome, fxpMoneyEscrowed, fxpSharesEscrowed, tx.gasprice]
-    orderID = ripemd160(order, items=10)
+    orderID = ORDERS.getOrderIdHash(type, market, attoshares, displayPrice, sender, block.number, outcome, fxpMoneyEscrowed, fxpSharesEscrowed)
     require(ORDERS.getAmount(orderID, type, market, outcome) == 0)
-    ORDERS.saveOrder(orderID, type, market, attoshares, displayPrice, sender, outcome, fxpMoneyEscrowed, fxpSharesEscrowed, betterOrderID, worseOrderID, tradeGroupID, tx.gasprice)
+    orderID = ORDERS.saveOrder(orderID, type, market, attoshares, displayPrice, sender, outcome, fxpMoneyEscrowed, fxpSharesEscrowed, betterOrderID, worseOrderID, tradeGroupID, tx.gasprice)
 
-    return(orderID)
+    return(orderID:bytes20)
```
MicahZoltu: Style: Missing space between `: bytes20`.

src/trading/trade.se
```diff
@@ -79,23 +78,24 @@ def publicTakeBestOrder(direction: int256, market: address, outcome: int256, fxp
...
-def trade(sender: address, direction: int256, market: address, outcome: int256, fxpAmount: int256, fxpPrice: int256, tradeGroupID: int256):
+def trade(sender: address, direction: uint8, market: address, outcome: uint8, fxpAmount: uint256, fxpPrice: int256, tradeGroupID: uint256):
     assertNoValue()
     self.controller.assertIsWhitelisted(msg.sender)
     fxpAmount = self.takeBestOrder(sender, direction, market, outcome, fxpAmount, fxpPrice, tradeGroupID)
     if(fxpAmount > 0 and msg.gas >= MINIMUM_GAS_NEEDED):
-        return MAKEORDER.makeOrder(sender, direction, fxpAmount, fxpPrice, market, outcome, 0, 0, tradeGroupID)
-    return 1
+        orderID = MAKEORDER.makeOrder(sender, direction, fxpAmount, fxpPrice, market, outcome, 0, 0, tradeGroupID)
+        return (orderID:bytes20)
```
MicahZoltu: Style: remove the intermediate local variable and just `return MAKEORDER...`.

src/trading/Orders.sol
```diff
@@ -0,0 +1,320 @@
...
+import 'ROOT/libraries/Trading.sol';
+import 'ROOT/reporting/Interfaces.sol';
+import 'ROOT/trading/OrdersFetcher.sol';
+
+
+/**
+ * @title Orders
+ * @dev Storage of all data associated with orders
+ */
+contract Orders is Controlled {
```
MicahZoltu: This needs to derive from `DelegationTarget` as the first base class: `contract Orders is DelegationTarget, Controlled { ... }`.

tests/conftest.py
```diff
@@ -223,7 +223,7 @@ def uploadAllContracts(self):
                 if name == 'controller': continue
                 # remove this and Interfaces.sol when https://github.com/ethereum/solidity/issues/2665 is fixed
                 if name == 'Interfaces': continue
-                contractsToDelegate = ['orders','tradingEscapeHatch']
+                contractsToDelegate = ['tradingEscapeHatch']
```
MicahZoltu: Orders should still be delegated, so this change should be rolled back.

tests/conftest.py
```diff
@@ -340,4 +340,4 @@ def sessionFixture():
 @fixture
 def contractsFixture(sessionFixture):
     sessionFixture.resetSnapshot()
-    return sessionFixture
+    return sessionFixture
```
MicahZoltu: Style: Files should all end in a newline.  Recommend adding EditorConfig support to whatever editor you are using so this happens automatically.

tests/trading/test_cancelOrder.py
```diff
@@ -22,10 +22,12 @@
 WORSE_ORDER_ID = 6
 GAS_PRICE = 7
 
+ZEROED_ORDER_ID = '\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
```
MicahZoltu: This can be simplified to `bytearray(32)`.  Consider also removing this global variable and just inlining `bytearray(32)` instead of `ZEROD_ORDER_ID` (which is about as long).

tests/trading/test_cancelOrder.py
```diff
@@ -105,8 +108,14 @@ def test_exceptions(contractsFixture):
...
     makerInitialCash = cash.balanceOf(tester.a1)
     marketInitialCash = cash.balanceOf(market.address)
-    orderID = makeOrder.publicMakeOrder(orderType, fxpAmount, fxpPrice, market.address, outcomeID, tradeGroupID, sender=tester.k1)
-    assert(orderID != 0), "Order ID should be non-zero"
+
+    # yesShareToken = contractsFixture.applySignature('shareToken', market.getShareToken(YES))
+    # print yesShareToken.balanceOf(tester.a1)
+    # print "Making order"
+    orderID = makeOrder.publicMakeOrder(orderType, fxpAmount, fxpPrice, market.address, outcomeID, ZEROED_ORDER_ID, ZEROED_ORDER_ID, tradeGroupID, sender=tester.k1)
+    # print yesShareToken.balanceOf(tester.a1)
```
MicahZoltu: Delete dead code rather than leaving it commented out.

tests/trading/test_cancelOrder.py
```diff
@@ -105,8 +108,14 @@ def test_exceptions(contractsFixture):
...
+    order = ordersFetcher.getOrder(orderID, orderType, market.address, outcomeID)
```
MicahZoltu: This doesn't appear to be used for anything or asserted on.  Delete it if it is unnecessary.  If it is necessary provide a comment as to why.

tests/trading/test_cancelOrder.py
```diff
@@ -116,9 +125,9 @@ def test_exceptions(contractsFixture):
 
     # cancelOrder exceptions
     with raises(TransactionFailed):
-        cancelOrder.publicCancelOrder(0, orderType, market.address, outcomeID, sender=tester.k1)
+        cancelOrder.publicCancelOrder(ZEROED_ORDER_ID, orderType, market.address, outcomeID, sender=tester.k1)
     with raises(TransactionFailed):
-        cancelOrder.publicCancelOrder(orderID + 1, orderType, market.address, outcomeID, sender=tester.k1)
+        cancelOrder.publicCancelOrder('\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01', orderType, market.address, outcomeID, sender=tester.k1)
```
MicahZoltu: You can simplify `'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01'` to `'\x01'.rjust(32, '\x00')`.

tests/trading/test_cancelOrder.py
adrake33: Sorry, I had meant to remove this, but I missed it before committing.

tests/trading/test_cancelOrder.py
MicahZoltu: Same comment for other test files.

tests/trading/test_orders.py
```diff
@@ -23,207 +23,215 @@
...
+        "betterOrderID": ZEROED_ORDER_ID,
+        "worseOrderID": ZEROED_ORDER_ID,
         "tradeGroupID": 0
     }
-    assert(orders.saveOrder(order["orderID"], order["type"], market.address, order["fxpAmount"], order["fxpPrice"], order["sender"], order["outcome"], order["fxpMoneyEscrowed"], order["fxpSharesEscrowed"], order["betterOrderID"], order["worseOrderID"], order["tradeGroupID"], 1) == 1), "Save order"
-    bestOrderID = int(orders.getBestOrderID(1, market.address, outcomeID), 16)
-    worstOrderID = int(orders.getWorstOrderID(1, market.address, outcomeID), 16)
-    assert(bestOrderID == 5)
-    assert(worstOrderID == 5)
+    hashedOrderID5 = orders.saveOrder(ZEROED_ORDER_ID, order["type"], market.address, order["fxpAmount"], order["fxpPrice"], order["sender"], order["outcome"], order["fxpMoneyEscrowed"], order["fxpSharesEscrowed"], order["betterOrderID"], order["worseOrderID"], order["tradeGroupID"])
```
MicahZoltu: Naming: Variable should just be called `orderId`.  As mentioned previously, the fact that it is a hash is an implementation detail.  Also, it is unclear why this has `5`  on the end?

tests/trading/test_orders.py
```diff
@@ -23,207 +23,215 @@
...
+
 def test_walkOrderList_bids(contractsFixture):
     market = contractsFixture.binaryMarket
-    orders = contractsFixture.contracts['orders']
-    ordersFetcher = contractsFixture.contracts['ordersFetcher']
+    orders = contractsFixture.contracts['Orders']
+    ordersFetcher = contractsFixture.contracts['OrdersFetcher']
     outcomeID = 1
     order = {
         "orderID": 5,
```
MicahZoltu: This should be `'orderId': longTo32Bytes(5),`.  You'll have to create a method for `longTo32Bytes`, there are some helpful tips on how to do this on StackOverflow.

MicahZoltu: Alternatively, a randomly generated 32-byte array would serve the purposes here.  It just needs to be unique.

MicahZoltu: Same comment for other tests in this file.

tests/trading/test_orders.py
MicahZoltu: Same comment for other tests in this file.

tests/trading/test_orders.py
```diff
@@ -23,207 +23,215 @@
...
+        "worseOrderID": ZEROED_ORDER_ID,
         "tradeGroupID": 0
     }
-    assert(orders.saveOrder(order["orderID"], order["type"], market.address, order["fxpAmount"], order["fxpPrice"], order["sender"], order["outcome"], order["fxpMoneyEscrowed"], order["fxpSharesEscrowed"], order["betterOrderID"], order["worseOrderID"], order["tradeGroupID"], 1) == 1), "Save order"
-    bestOrderID = int(orders.getBestOrderID(1, market.address, outcomeID), 16)
-    worstOrderID = int(orders.getWorstOrderID(1, market.address, outcomeID), 16)
-    assert(bestOrderID == 5)
-    assert(worstOrderID == 5)
+    hashedOrderID5 = orders.saveOrder(ZEROED_ORDER_ID, order["type"], market.address, order["fxpAmount"], order["fxpPrice"], order["sender"], order["outcome"], order["fxpMoneyEscrowed"], order["fxpSharesEscrowed"], order["betterOrderID"], order["worseOrderID"], order["tradeGroupID"])
+    assert(isinstance(hashedOrderID5, bytes)), "Save order"
```
MicahZoltu: We can probably remove this assert.  I don't think it adds much value.  Probably good to `assert orderId != bytearray(32)` though (to make sure the test fails here if we get back `0`).

tests/trading/test_cancelOrder.py
adrake33: This was part of the debugging code that I commented out above.  I'll remove it.

tests/trading/test_orders.py
adrake33: I added the `5` on the end to differentiate it from some of the other orderIds in the file.  Previously, we were just using integers, so lines like these:

`    assert(bestOrderID == 5)
    assert(worstOrderID == 6)`

had to be changed to:

`    assert(bestOrderID == hashedOrderID5)
    assert(worstOrderID == hashedOrderID6)`

I'll change the `hashedOrderID` prefix to `orderId`.

tests/trading/test_orders.py
```diff
@@ -552,576 +597,600 @@ def runtest(testCase):
...
 
 def test_saveOrder(contractsFixture):
     market = contractsFixture.binaryMarket
-    orders = contractsFixture.contracts['orders']
-    ordersFetcher = contractsFixture.contracts['ordersFetcher']
-    order1 = 12345
-    order2 = 98765
+    orders = contractsFixture.contracts['Orders']
+    ordersFetcher = contractsFixture.contracts['OrdersFetcher']
+    hashedOrderID1 = orders.saveOrder(ZEROED_ORDER_ID, BID, market.address, fix('10'), fix('0.5'), tester.a1, NO, 0, fix('10'), ZEROED_ORDER_ID, ZEROED_ORDER_ID, 1)
```
MicahZoltu: Style: I personally prefer a blank line between the test setup (like getting all of the contracts from the fixture) and when the test starts, and then when assertions happen.  Not all of our tests have clear separation like this, but this one could benefit from a blank line above this line I think.

src/trading/Orders.sol
```diff
@@ -0,0 +1,320 @@
...
+
+    function getBestOrders(Trading.TradeTypes _type, Market _market, uint8 _outcome) public constant returns (bytes20) {
+        return bestOrder[getBestOrderWorstOrderHash(_market, _outcome, _type)];
+    }
+
+    function getWorstOrders(Trading.TradeTypes _type, Market _market, uint8 _outcome) public constant returns (bytes20) {
+        return worstOrder[getBestOrderWorstOrderHash(_market, _outcome, _type)];
+    }
+
+    function getAmount(bytes20 _orderId, Trading.TradeTypes _type, Market _market, uint8 _outcome) public constant returns (uint256) {
```
MicahZoltu: Solc is complaining about unused variables for a lot of these.  I know we are keeping them in for now so we don't have to change too much, but we should remove the parameter names and just have `(bytes20 _orderId, Trading.TradeTypes, Market, uint8)` as the parameter list so solc understands that we intentionally aren't using them.

tests/trading/test_cancelOrder.py
adrake33: I tried converting to bytearray(20) (since the orderId is currently 20 bytes, not 32), but that's throwing an error.  I think it's getting passed into the function slightly differently and may need to be formatted.

adrake33: Passing in `str(bytearray(20))` seems to work.

tests/conftest.py
adrake33: I updated the contract declaration in Orders.sol to `contract Orders is DelegationTarget, Controlled {`.  However, when I tried to change this line to `contractsToDelegate = ['tradingEscapeHatch']`, every script started failing with errors like this:

```
request = <FixtureRequest for <Function 'test_cancelBid'>>

    def fill(request):
        item = request._pyfuncitem
        fixturenames = item.fixturenames
        argnames = item._fixtureinfo.argnames
    
        for fname in fixturenames:
            if fname not in item.funcargs and fname not in argnames:
>               item.funcargs[fname] = request.getfixturevalue(fname)

/usr/local/lib/python2.7/site-packages/pytest_lazyfixture.py:35: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
conftest.py:338: in sessionFixture
    return ContractsFixture()
conftest.py:155: in __init__
    self.initializeAllContracts()
conftest.py:252: in initializeAllContracts
    self.contracts[contractName].initialize(self.controller.address)
/usr/local/lib/python2.7/site-packages/ethereum/tools/tester.py:97: in kall
    startgas=kwargs.get('startgas', STARTGAS)
/usr/local/lib/python2.7/site-packages/ethereum/tools/tester.py:147: in tx
    o = self.direct_tx(transaction)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <ethereum.tools.tester.Chain object at 0x7f1986be0910>, transaction = <Transaction(08fa)>

    def direct_tx(self, transaction):
        self.last_tx, self.last_sender = transaction, None
        success, output = apply_transaction(self.head_state, transaction)
        self.block.transactions.append(transaction)
        if not success:
>           raise TransactionFailed()
E           TransactionFailed

/usr/local/lib/python2.7/site-packages/ethereum/tools/tester.py:140: TransactionFailed
```

Reverting this line back to `contractsToDelegate = ['tradingEscapeHatch']` caused the errors stop occurring.

tests/utils.py
```diff
@@ -12,6 +13,9 @@ def unfix(n):
 def stringToBytes(value):
     return value.ljust(32, '\x00')
 
+def longTo32Bytes(value):
+    return pack(">l", value).rjust(32, '\x00')
```
MicahZoltu: Very nice.  This is even less code than I expected.

adrake33: Thanks, I had seen some examples using `pack()` on StackOverflow, but since they weren't producing the desired number of bytes, I got the idea to use `rjust()` from the stringToBytes() function above.

PR #271, opened 2017-08-28
Flesh out tests for reporting and fix 2 obvious bugs

nuevoalex:


src/reporting/Market.sol
```diff
@@ -445,7 +445,7 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable {
         }
 
         // Automated reporting period has not passed yet
-        if (block.timestamp < getAutomatedReportDueTimestamp()) {
+        if (!automatedReportReceived && block.timestamp < getAutomatedReportDueTimestamp()) {
```
MicahZoltu: 🤔 Out of scope of this PR, but it would probably be better if we made it so that automated dispute began (3 day timer started) as soon as the automated report was received.  That would reduce the "fastest resolution" down to just over 3 days, rather than 6 days as it is now.

src/reporting/ReportingWindow.sol
```diff
@@ -67,7 +67,7 @@ contract ReportingWindow is DelegationTarget, Typed, Initializable {
     function migrateMarketInFromNibling() public afterInitialized returns (bool) {
         Market _market = Market(msg.sender);
         Branch _shadyBranch = _market.getBranch();
-        require(branch.isParentOf(_shadyBranch));
+        require(_shadyBranch.isParentOf(branch));
```
MicahZoltu: This check is insecure as `_shadyBranch` may lie about who its parent is.  If this is the direction we want, then this should be changed to `branch.getParentBranch() == _shadyBranch`.

tests/reporting/test_reporting.py
```diff
@@ -1,108 +1,442 @@
...
+    for (key, address) in [(tester.k1, tester.a1), (tester.k2, tester.a2), (tester.k3, tester.a3)]:
+        firstRegistrationToken.register(sender=key)
+        assert firstRegistrationToken.balanceOf(address) == 1
+        assert reputationToken.balanceOf(address) == 1 * 10**6 * 10**18 - 10**18
+
+    sessionFixture.reputationToken = reputationToken
+    sessionFixture.reportingWindow = reportingWindow
+    sessionFixture.firstRegistrationToken = firstRegistrationToken
+    sessionFixture.reportingTokenNo = reportingTokenNo
+    sessionFixture.reportingTokenYes = reportingTokenYes
```
MicahZoltu: This will add these variables to the `sessionFixture` for everyone.  This isn't directly a problem, but it may result in unexpected behavior where if the tests are run before a test that uses `reportingSnapshot` they will behave differently from tests that are run after a test that references `reportingSnapshot`.

I recommend creating a separate class to hold these objects rather than applying them to `sessionFixture`.  This way you only get access to these variables if you depend on `reportingSnapshot` and you can't get access to them by only depending on `sessionFixture` and happening to have some other test run prior to yours that depends on `reportingSnapshopt`.

tests/reporting/test_reporting.py
```diff
@@ -1,108 +1,442 @@
...
+    cash = reportingFixture.cash
+    market = reportingFixture.binaryMarket
+    reputationToken = reportingFixture.reputationToken
+    branch = reportingFixture.branch
+    reportingTokenNo = reportingFixture.reportingTokenNo
+    reportingTokenYes = reportingFixture.reportingTokenYes
+
+    # We can't yet report on the market as it's in the pre reporting phase
+    assert market.getReportingState() == reportingFixture.constants.PRE_REPORTING()
+    with raises(TransactionFailed, message="Reporting cannot be done in the PRE REPORTING state"):
```
MicahZoltu: What does the `message` parameter here do?

tests/reporting/test_reporting.py
```diff
@@ -1,108 +1,442 @@
...
+from pytest import fixture, mark, lazy_fixture, raises
 from datetime import timedelta
 
 tester.STARTGAS = long(6.7 * 10**6)
 
-def test_reporting(contractsFixture):
-    # seed legacy rep contract
-    legacyRepContract = contractsFixture.contracts['legacyRepContract']
+@fixture(scope="session")
+def reportingSnapshot(sessionFixture):
```
MicahZoltu: Style: For test files, I tend to prefer putting the tests themselves at the top, and dependencies and helper methods at the bottom.  The idea is to make it easy for someone reviewing test failures to see the tests first and only dig into the implementation details if necessary (scroll down).

src/reporting/Market.sol
nuevoalex: Yea I agree that makes more sense. I can do that in this PR or add a comment to do so.

src/reporting/ReportingWindow.sol
nuevoalex: Ah good point. Yea I believe that will be what we have to do. The code before was just logically incorrect.

tests/reporting/test_reporting.py
nuevoalex: Yup!

PR #273, opened 2017-08-29
Automated report timing

nuevoalex:


src/reporting/Market.sol
```diff
@@ -160,6 +161,7 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable {
         // we have to create the reporting token so the rest of the system works (winning reporting token must exist)
         getReportingToken(_payoutNumerators);
         automatedReportReceived = true;
+        automatedReportReceivedTime = block.timestamp;
```
MicahZoltu: We can probably get rid of `automatedReportReceived` and just use `automatedReportReceviedTimestamp` instead.  If it is `0` it means no report received.  This will save us a couple state updates.

PR #275, opened 2017-08-30
Rep faucet

nuevoalex:
I think this is ready to look at initially.

Notes:

- The commented out behavior we had implied the faucet funded test accounts by transferring from a provided branch. Right now I'm instead making the legacy rep contract itself the faucet balance holder. If this should go back to the branch it should be an easy switch, but this seemed cleaner.

- There is no explicit protection against this being uploaded and used on the main net yet. One thing we could do is add some sort of marker contract to the controller that suggests we're on the test net and check for it's existence. Curious on other's thoughts.


MicahZoltu:
Easiest way to check for Foundation branch is probably to hard-code the Foundation branch REP contract and call `totalSupply` on it.  If you get back 11,000,000 then we are on Foundation, if you get back 0, we are on a test network.

MicahZoltu:
I would put the check in the constructor so upload of this contract will fail on Foundation.

src/LegacyRepContract.sol
```diff
@@ -0,0 +1,82 @@
+pragma solidity ^0.4.13;
+
+import 'ROOT/legacy_reputation/StandardToken.sol';
+import 'ROOT/legacy_reputation/Ownable.sol';
+import 'ROOT/Controller.sol';
+
+
+contract LegacyRepContract is StandardToken, Controlled, Ownable {
```
MicahZoltu: This should be a `VariableSupplyToken` as we'll be increasing total supply anytime someone wants REP.

src/LegacyRepContract.sol
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    function transfer(address _to, uint256 _value) public doneCreating returns (bool) {
+        balances[msg.sender] = balances[msg.sender].sub(_value);
+        balances[_to] = balances[_to].add(_value);
+        Transfer(msg.sender, _to, _value);
+        return true;
+    }
+
+    function setSaleDistribution(address[] _addresses, uint256[] _balances) public onlyOwner stillCreating returns (bool) {
```
MicahZoltu: This method can go away, we won't be seeding it with a sale, it'll start with `0` REP and anytime someone calls the faucet function REP will be minted.  `VariableSupplyToken` has a mint function you can call to make the process easy.

src/LegacyRepContract.sol
nuevoalex: Ah ok. We don't need to allocate a balance for the token itself then if we're doing that. Will do!

src/LegacyRepContract.sol
```diff
@@ -0,0 +1,82 @@
...
+            }
+        }
+        if (totalSupply == 11000000 ether) {
+            seeded = true;
+        }
+        require(totalSupply <= 11000000 ether);
+        return true;
+    }
+
+    function getRidOfDustForLaunch() public onlyOwner returns (bool) {
```
MicahZoltu: Also no need for this, or any of the other initialization/seeding functions.  Also no need to add any additional protections on transfer or anything.  This doesn't need to be particularly secure.  Also, the modifiers can all likely go away.

tests/conftest.py
```diff
@@ -247,7 +247,7 @@ def whitelistLibraryTestContracts(self):
             self.controller.addToWhitelist(self.contracts[name].address)
 
     def initializeAllContracts(self):
-        contractsToInitialize = ['Mutex', 'Cash','orders','completeSets','makeOrder','takeBidOrder','takeAskOrder','takeOrder','cancelOrder','trade','ClaimProceeds','tradingEscapeHatch','ordersFetcher']
+        contractsToInitialize = ['Mutex', 'Cash','orders','completeSets','makeOrder','takeBidOrder','takeAskOrder','takeOrder','cancelOrder','trade','ClaimProceeds','tradingEscapeHatch','ordersFetcher', 'LegacyRepContract']
```
MicahZoltu: I believe we can drop this, the REP contract shouldn't need any initialization.

src/LegacyRepContract.sol
MicahZoltu: Also, I don't think this contract needs to be Controlled or Ownable.  Just `VariableSupplyToken`.

src/LegacyRepContract.sol
nuevoalex: Right now a good bit of code works by migrating REP from whatever the controller is pointing to at 'LegacyRepContract' onto the ReputationToken. Do we want to refactor this in some fashion?

MicahZoltu: No, that is desirable.  However, that shouldn't have any impact on this contract?  The `LegacyRepContract` returned from the controller can be _any_ ERC20 token.  Just needs an approve and a `transferFrom` method.  We want this contract to be a dead simple ERC20 token that allows anyone to mint REP for themselves for testing.

nuevoalex: Got it, will do

src/LegacyRepContract.sol
```diff
@@ -0,0 +1,40 @@
...
+        address _foundationRepAddress = address(0xe94327d07fc17907b4db788e5adf2ed424addff6);
+        uint256 size;
+        assembly { size := extcodesize(_foundationRepAddress) }
+        require(size == 0);
+        // Store the creator for use in automated testing funding
+        creator = msg.sender;
+    }
+
+    function faucet() public returns (bool) {
+        require(balanceOf(msg.sender) == 0);
```
MicahZoltu: No need for this.  The Faucet should _always_ give out REP to anyone who asks, even if they already have some.

src/LegacyRepContract.sol
```diff
@@ -0,0 +1,40 @@
...
+        mint(msg.sender, FAUCET_AMOUNT);
+        FundedAccount(this, msg.sender, FAUCET_AMOUNT, block.timestamp);
+        return true;
+    }
+
+    // Function for automated testing
+    function testFund(uint256 _amount) public returns (bool) {
+        require(msg.sender == creator);
+        mint(msg.sender, _amount);
+    }
```
MicahZoltu: Can you help me understand this?  Why not just call `faucet`?  If its because you want the ability to ask for a certain amount, we can just have an overload of `faucet(uint256)` that takes in an amount.  There isn't any real need to secure it against someone asking for a lot.  Maybe have a cap at something like 2^128 just so someone can't ask for so much REP that the totalSupply caps out (would require them calling that method 2^128 times to cap out totalSupply).

src/LegacyRepContract.sol
```diff
@@ -0,0 +1,40 @@
...
+contract LegacyRepContract is VariableSupplyToken {
+    event FundedAccount(address indexed _branch, address indexed _sender, uint256 _repBalance, uint256 _timestamp);
+
+    uint256 private constant FAUCET_AMOUNT = 47 ether;
+
+    string public constant name = "Reputation";
+    string public constant symbol = "REP";
+    uint256 public constant decimals = 18;
+
+    address private creator;
```
MicahZoltu: See comments below, I believe we can remove this.

src/LegacyRepContract.sol
nuevoalex: Yea given we don't care about restricting it I'll just make this take an argument as you suggested

src/LegacyRepContract.sol
```diff
@@ -0,0 +1,40 @@
...
+        FundedAccount(this, msg.sender, FAUCET_AMOUNT, block.timestamp);
+        return true;
+    }
+
+    // Function for automated testing
+    function testFund(uint256 _amount) public returns (bool) {
+        require(msg.sender == creator);
+        mint(msg.sender, _amount);
+    }
+}
```
MicahZoltu: Style: Missing newline at end of file.  Recommend adding EditorConfig support to whatever your editor of choice is so this is handled for you.

nuevoalex: Done! Kept meaning to do this...

src/LegacyRepContract.sol
```diff
@@ -0,0 +1,33 @@
+pragma solidity ^0.4.13;
+
+import 'ROOT/libraries/token/VariableSupplyToken.sol';
+import 'ROOT/libraries/token/ERC20.sol';
```
MicahZoltu: Recommend removing this import, its not directly used anywhere in this contract.

PR #276, opened 2017-08-31
Cancel order migration

adrake33:
﻿# Pull Request Template
### Overview

Replacing cancelOrder.se with CancelOrder.sol

### Examples (if necessary)


tests/trading/test_cancelOrder.py
```diff
@@ -109,19 +109,13 @@ def test_exceptions(contractsFixture):
     orderID = makeOrder.publicMakeOrder(orderType, fxpAmount, fxpPrice, market.address, outcomeID, longTo32Bytes(0), longTo32Bytes(0), tradeGroupID, sender=tester.k1)
     assert(orderID != bytearray(32)), "Order ID should be non-zero"
 
-    # Permissions exceptions
-    with raises(TransactionFailed):
-        cancelOrder.cancelOrder(tester.a1, orderID, orderType, market.address, outcomeID, sender=tester.k1)
-    with raises(TransactionFailed):
-        cancelOrder.refundOrder(tester.a1, orderType, 0, fxpAmount, market.address, outcomeID, sender=tester.k1)
```
adrake33: I removed these two tests because `cancelOrder()` used to be a function that was not public.  (Now `publicCancelOrder()` and `cancelOrder()` have been combined.)  Also, `refundOrder()` is private, and so this line was giving an error saying that cancelOrder did not contain `refundOrder()` (as opposed to a transaction failure when using Serpent).

src/trading/CancelOrder.sol
```diff
@@ -0,0 +1,74 @@
...
+    function cancelOrder(bytes32 _orderId, Trading.TradeTypes _type, Market _market, uint8 _outcome) nonReentrant external returns (bool) {
+        require(_orderId != bytes32(0));
+        require(_market.getTypeName() == "Market");
+
+        // Look up the order the sender wants to cancel
+        Orders _orders = Orders(controller.lookup("Orders"));
+        uint256 _fxpMoneyEscrowed = _orders.getOrderMoneyEscrowed(_orderId, _type, _market, _outcome);
+        uint256 _fxpSharesEscrowed = _orders.getOrderSharesEscrowed(_orderId, _type, _market, _outcome);
+
+        // Check that the order ID is correct, the type is bid (1) or ask (2), and that the sender owns the order
```
MicahZoltu: Recommend removing the `(1)` and `(2)` from the comment since Bid and Ask are now enum members and we plan to change their indexing at some point.

src/trading/CancelOrder.sol
```diff
@@ -0,0 +1,74 @@
...
+        require(_type == Trading.TradeTypes.Bid || _type == Trading.TradeTypes.Ask);
```
MicahZoltu: We can probably delete this now that we have an enum.  This was originally intended to protect us in Serpent when this function just took in a number for `_type`.

adrake33:
The requested changes have been made.  Please review again and let me know if this looks ready to merge.

PR #277, opened 2017-09-01
Ported ShareToken to Solidity

redsquirrel:
### Overview
Rewrote `shareToken.se` to `ShareToken.sol`

Should be functionally equivalent (see below for one exception)

### Notes
The one difference in behavior between `BasicToken.transfer` (which `ShareToken` inherits) and the original `shareToken.se` was this check `require(1 <= value and value <= self.accounts[msg.sender].balance)`. `BasicToken`'s implementation doesn't validate the value.


src/trading/ShareToken.sol
```diff
@@ -0,0 +1,51 @@
...
+        market = _market;
+        outcome = _outcome;
+    }
+
+    function createShares(address _owner, uint256 _fxpValue) onlyWhitelistedCallers external returns(bool) {
+        mint(_owner, _fxpValue);
+        return true;
+    }
+
+    function destroyShares(address _owner, uint256 _fxpValue) onlyWhitelistedCallers external returns(bool) {
```
MicahZoltu: I think at some point we should rename this to `burn` and the other to `mint` and then have them apply the modifier and call `super.mint()`/`super.burn()` (or whatever the syntax for that is).  That is probably better as a separate PR, possibly after we have finished the Solidity migration.

PR #278, opened 2017-09-02
Changes pattern to using interfaces.

MicahZoltu:
Interfaces drastically improve compilation time and IDE responsiveness as it prevents compiling the world when you only want to make small changes.  It also will enable us to create Mocks of most of our contracts which should facilitate unit testing, without the need to setup the world.

The interfaces _do_ result in some redundancies and maintenance burden, but I believe the performance and testing benefits outweigh those costs.

There may be a couple contracts that don't follow the interface pattern yet, and that is fine and we can address as we see them, this PRs intent is to convert _most_ things to this new pattern.

Other changes that came up while doing this:
* extracted out Controlled from Controller
* removed some unused parameters from Orders.getOrders and OrdersFetcher.getOrder
* refactored OrderBook.getOrderBook to address stack depth problems
* deleted Interfaces.sol, it was meant to be temporary and this extracts out the last couple interfaces
* changed isContainerFor* methods in Market to match patterns in Branch
* moved `Trading.sol` to `trading` folder from libraries
* bumped solidity version during debugging, figured it didn't hurt to leave it


Dockerfile-test
```diff
@@ -4,7 +4,7 @@ RUN curl -sL https://deb.nodesource.com/setup_8.x | bash - \
 	&& apt-get install -y nodejs
 RUN npm install -g AugurProject/solium
 
-RUN wget --quiet --output-document /usr/local/bin/solc https://github.com/ethereum/solidity/releases/download/v0.4.13/solc-static-linux \
+RUN wget --quiet --output-document /usr/local/bin/solc https://github.com/ethereum/solidity/releases/download/v0.4.16/solc-static-linux \
```
redsquirrel: Looks like you upgraded to Solidity 0.4.16. How come you left it at 0.4.13 in the Solidity files?

MicahZoltu: The upgrade to Solidity was done as part of troubleshooting, and it seemed reasonable to keep it since everything worked as is and the new version has some useful features.  I wanted to avoid touching any more files than necessary in this PR though, and bumping solidity version would touch _all_ .sol files.  I do plan on a follow-up PR that bumps us to Solidity 0.4.16 or later.

MicahZoltu:
@nuevoalex @adrake33 I'm going to merge this now rather than waiting on you two for review as I don't want a big PR like this to sit too long and you are both out today I believe.  Feel free to leave comments on this PR even after I merge it and I'll address them in a follow-up PR.

src/libraries/IOwnable.sol
```diff
@@ -0,0 +1,7 @@
+pragma solidity ^0.4.13;
+
+
+contract IOwnable {
+    function getOwner() public constant returns (address);
+    function transferOwnership(address newOwner) public returns (bool);
```
adrake33: Are modifiers not required for interface function declarations (such as `onlyOwner` in this case)?

src/reporting/IDisputeBond.sol
```diff
@@ -0,0 +1,12 @@
+pragma solidity ^0.4.13;
+
+import 'ROOT/libraries/Typed.sol';
+import 'ROOT/reporting/IMarket.sol';
+
+
+contract IDisputeBond is Typed {
```
adrake33: Perhaps we should either rename DisputeBondToken to DisputeBond to match this, or rename this to IDisputeBondToken.sol?

src/trading/IClaimProceeds.sol
```diff
@@ -0,0 +1,6 @@
+pragma solidity ^0.4.13;
+
+
+contract IClaimProceeds {
+
```
adrake33: Why does this not have any functions declared?

src/libraries/IOwnable.sol
MicahZoltu: Correct, they are not required and in fact not allowed in this case since this interface doesn't know what the modifiers mean.  The modifiers are an implementation detail.  In theory, we could make the various modifiers available in interfaces, at which point we could apply them as interfaces.  I'm on the fence as to whether that is a good thing or not.

src/reporting/IDisputeBond.sol
MicahZoltu: The long-term plan is to rename `DisputeBondToken` to `DisputeBond` and make it no longer a token and instead an `Ownable`.  However, I wanted to wait until after this PR to address that (trying to avoid growing this PR any more than necessary).

src/trading/IClaimProceeds.sol
MicahZoltu: No other contracts call anything on it.  This is basically just a placeholder in order to keep the pattern consistently followed.  I suspect at some point we'll have contracts that call this, at which point the interface will be filled out.

src/Controlled.sol
```diff
@@ -0,0 +1,43 @@
+pragma solidity ^0.4.13;
+
+import 'ROOT/IControlled.sol';
+import 'ROOT/IController.sol';
+
+
+contract Controlled is IControlled {
+    IController public controller;
```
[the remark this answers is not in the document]
MicahZoltu: Hmm, I needed it to be public for something but I don't remember what.  I just changed it to internal and everything works (tests pass) so I'll submit a follow-up PR to make it internal.

PR #279, opened 2017-09-04
Migrates `completeSets.se` to `CompleteSets.sol`.

MicahZoltu:
Along the way fixes an outstanding bug where share prices weren't correctly calculated for scalar markets and were only correct for binary/categorical markets by happy coincidence.  The fix was a functional change that I originally intended on addressing in the future, which is to make it so share purchases are 1:1 with denomination tokens.  If you buy a 5 complete sets it will cost 5 indivisible units of denomination token and you will receive 5 indivisible shares of each outcome.  This is now the case no matter what the range of the market is.

Fixes a bug with `fundedRepSnapshot` where it would get bleed-over from previous tests if it wasn't used in a test first.  Pytest fixtures are evaluated lazily, even if they are `session` so we need to first reset the base fixture when we want to create an extended fixture.


src/trading/ClaimProceeds.sol
```diff
@@ -57,11 +57,10 @@ contract ClaimProceeds is Controlled, ReentrancyGuard, IClaimProceeds {
     }
 
     function calculateProceeds(IMarket _market, IReportingToken _winningReportingToken, uint8 _outcome, uint256 _numberOfShares) public constant returns (uint256) {
-        uint256 _completeSetCostInAttotokens = _market.getCompleteSetCostInAttotokens();
         uint256 _payoutNumerator = _winningReportingToken.getPayoutNumerator(_outcome);
-        uint256 _getPayoutDenominator = _market.getPayoutDenominator();
+        uint256 _payoutDenominator = _market.getPayoutDenominator();
         // NOTE: rounding error here will result in _very_ tiny amounts of denominationToken left in the market
-        return _numberOfShares.mul(_completeSetCostInAttotokens).div(10**18).mul(_payoutNumerator).div(_getPayoutDenominator);
```
nuevoalex: Just curious if you know why we were doing `mul(_completeSetCostInAttotokens).div(10**18)` before and why we aren't now. I'm assuming the payoutNumerator includes the total set cost in its calculation then but this should've been wrong before then.

src/trading/CompleteSets.sol
```diff
@@ -0,0 +1,73 @@
...
+    using SafeMathUint256 for uint256;
+
+    /**
+     * Buys `_amount` shares of every outcome in the specified market.
+    **/
+    function publicBuyCompleteSets(IMarket _market, uint256 _amount) external onlyInGoodTimes nonReentrant returns (bool) {
+        return this.buyCompleteSets(msg.sender, _market, _amount);
+    }
+
+    function buyCompleteSets(address _sender, IMarket _market, uint256 _amount) external onlyWhitelistedCallers returns (bool) {
```
nuevoalex: Why make this external? Seems like private or internal is more appropriate.

src/trading/CompleteSets.sol
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    function sellCompleteSets(address _sender, IMarket _market, uint256 _amount) external onlyWhitelistedCallers returns (bool) {
+        require(_sender != address(0));
+        require(_market != IMarket(0));
+
+        uint8 _numOutcomes = _market.getNumberOfOutcomes();
+        ERC20 _denominationToken = _market.getDenominationToken();
+        uint256 _marketCreatorFeeRate = _market.getMarketCreatorSettlementFeeInAttoethPerEth();
+        uint256 _marketCreatorFee = _amount.mul(_marketCreatorFeeRate).div(10**18);
```
nuevoalex: Small nit but I prefer `1 ether` instead of `10**18`

MicahZoltu: Fixed.

src/trading/CompleteSets.sol
MicahZoltu: It is intended to be called by other Augur contracts (trusted callers).  `publicBuyCompleteSets` is intended to be called by untrusted external sources.  The `whitelistCallersOnly` modifier checks `msg.sender` which is why we need to call this via `this.buyCompleteSets` even when we are calling it privately because otherwise `msg.sender` will be the original `msg.sender` and the whitelisted callers check will fail.

src/trading/ClaimProceeds.sol
MicahZoltu: This was the bug I referred to in the PR description.  Things were broken before, and this fixes it by changing the way complete set size is calculated.  Now 1 indivisible complete set == 1 indivisible share in each outcome and it costs 1 indivisible unit of currency.  Previously there were some weird calculations involved (like randomly dividing by 10**18) that broke for scalar markets.

src/trading/takeBidOrder.se
```diff
@@ -149,9 +146,9 @@ def takeBidOrder(taker: address, orderID: bytes32, market: address, outcome: uin
     # maker is opening a long, taker is opening a short
     if takerDesiredSharesForThisOrder:
         # transfer cash from both parties into this contract for complete set purchase
-        takerPortionOfCompleteSetCost = safeFxpMul(takerDesiredSharesForThisOrder, sharePriceShort)
+        takerPortionOfCompleteSetCost = safeDiv(safeMul(takerDesiredSharesForThisOrder, sharePriceShort), market.getMaxDisplayPrice() - market.getMinDisplayPrice())
         denominationToken.transferFrom(taker, self, takerPortionOfCompleteSetCost)
-        makerPortionOfCompleteSetCost = safeFxpMul(takerDesiredSharesForThisOrder, sharePriceLong)
```
nuevoalex: I think we can remove `sharePriceLong` from this function now.

MicahZoltu: I'm working on migrating `takeBidOrder` to solidity right now, which still has `sharePriceLong` but it does remove some other pieces.  Recommend bringing this back up once that PR is out if you believe removal still makes sense (and why).

PR #281, opened 2017-09-04
Moving SafeMath*Testers under /tests

redsquirrel:
At @MicahZoltu's request, moving these helpers.


tests/libraries/math/test_safeMathUint256.py
```diff
@@ -4,14 +4,25 @@
 from ethereum.tools.tester import TransactionFailed
 from pytest import fixture, mark, lazy_fixture, raises
 
+@fixture(scope='session')
+def testerSnapshot(sessionFixture):
+    sessionFixture.uploadAndAddToController('solidity_test_helpers/SafeMathUint256Tester.sol')
```
MicahZoltu: This needs to do `sessionFixture.resetSnapshot()` first, otherwise it could end up working with an already used fixture.

tests/libraries/math/test_safeMathUint256.py
```diff
@@ -4,14 +4,25 @@
...
+    return sessionFixture.chain.snapshot()
+
+@fixture
+def testerContractsFixture(sessionFixture, testerSnapshot):
+    sessionFixture.chain.revert(testerSnapshot)
```
MicahZoltu: I believe you need to do `sessionFixture.resetSnapshot()` before this as well (though I'm less confident than above).

tests/libraries/math/test_safeMathInt256.py
```diff
@@ -4,6 +4,17 @@
 from ethereum.tools.tester import TransactionFailed
 from pytest import fixture, mark, lazy_fixture, raises
 
+@fixture(scope='session')
+def testerSnapshot(sessionFixture):
+    sessionFixture.uploadAndAddToController('solidity_test_helpers/SafeMathInt256Tester.sol')
```
MicahZoltu: You should be able to just `upload(...)` here, no need to add it to the controller.  Same comment below.

tests/libraries/math/test_safeMathInt256.py
```diff
@@ -13,8 +24,8 @@
     (1, 0, 0),
     (1, 1, 1)
 ])
-def test_mul(a, b, expectedResult, contractsFixture):
-    SafeMathInt256Tester = contractsFixture.contracts['SafeMathInt256Tester']
+def test_mul(a, b, expectedResult, testerContractsFixture):
+    SafeMathInt256Tester = testerContractsFixture.contracts['SafeMathInt256Tester']
```
MicahZoltu: Style: first letter should be lower case for local variables.

MicahZoltu: Same comment throughout.

PR #282, opened 2017-09-04
Topics solidity

nuevoalex:


nuevoalex:
I'm going to rebase and port this to use Interfaces as per the new style

PR #283, opened 2017-09-05
Topics solidity

nuevoalex:


tests/trading/test_topics.py
```diff
@@ -49,6 +49,6 @@ def test_topics(fundedRepFixture):
         topics.getTopicByOffset(0): topics.getPopularityByOffset(0),
         topics.getTopicByOffset(1): topics.getPopularityByOffset(1),
     } == {
-        bytesToLong('Politics'.ljust(32, '\x00')): 10,
-        bytesToLong('Sports'.ljust(32, '\x00')): fix('2.4'),
+        'Politics'.ljust(32, '\x00'): 10,
+        'Sports'.ljust(32, '\x00'): fix('2.4'),
```
redsquirrel: I'm curious about this. Was the `bytesToLong` never necessary? Or is Solidity's behavior different than Serpent in this case?

nuevoalex: Solidity's behavior is different in this case.

src/trading/Topics.sol
```diff
@@ -0,0 +1,55 @@
...
+contract Topics is DelegationTarget, Typed, Initializable, ITopics {
+    using SafeMathUint256 for uint256;
+
+    // Mapping of topic to popularity
+    IIterableMap private topics;
+
+    // FIXME: topics needs to be sorted on-chain, which means we need a sorted linked list collection with hinted sorting at insertion/update time instead of a simple iterableMap
+    function initialize(IController _controller) public beforeInitialized returns (bool) {
+        endInitialization();
+        controller = _controller;
```
MicahZoltu: As this is a `DelegationTarget`, it already has a `controller` set, so this shouldn't be necessary.  I believe the next line (instantiating the map) is still necessary, so I think we need to keep `initialize` but it can be parameterless.

src/trading/ITopics.sol
```diff
@@ -1,8 +1,10 @@
 pragma solidity ^0.4.13;
 
+import 'ROOT/IController.sol';
```
MicahZoltu: If you remove `IController` on initialize then make sure to remove this import.

PR #285, opened 2017-09-06
Trades solidity

nuevoalex:


src/trading/Trade.sol
```diff
@@ -0,0 +1,78 @@
...
+        uint256 _fxpAmountRemaining = this.takeBestOrder(msg.sender, _direction, _market, _outcome, _fxpAmount, _fxpPrice, _tradeGroupID);
+        return _fxpAmountRemaining;
+    }
+
+    function trade(address _sender, Trading.TradeDirections _direction, IMarket _market, uint8 _outcome, uint256 _fxpAmount, int256 _fxpPrice, uint256 _tradeGroupID) external onlyWhitelistedCallers returns (bytes32) {
+        uint256 _bestFxpAmount = this.takeBestOrder(_sender, _direction, _market, _outcome, _fxpAmount, _fxpPrice, _tradeGroupID);
+        if (_bestFxpAmount > 0 && msg.gas >= MINIMUM_GAS_NEEDED) {
+            return IMakeOrder(controller.lookup("makeOrder")).makeOrder(_sender, _direction, _bestFxpAmount, _fxpPrice, _market, _outcome, 0, 0, _tradeGroupID);
+        }
+        return 1;
```
MicahZoltu: This has always bothered me.  I wonder if we should change this method to have two return values, `success` and `orderId`?  Also, I believe the compiler will give a warning about implicit casting of uint to bytes32 here, you should wrap the `1` in `bytes32(1)`.

src/trading/Trade.sol
```diff
@@ -0,0 +1,78 @@
...
+    function takeOrder(address, bytes32, Trading.TradeTypes, IMarket, uint8, uint256, uint256) returns (uint256);
+}
+
+
+contract Trade is Controlled, ReentrancyGuard {
+    uint256 private constant MINIMUM_GAS_NEEDED = 300000;
+
+    function publicBuy(IMarket _market, uint8 _outcome, uint256 _fxpAmount, int256 _fxpPrice, uint256 _tradeGroupID) external onlyInGoodTimes nonReentrant returns (bytes32) {
+        bytes32 _output = this.trade(msg.sender, Trading.TradeDirections.Buying, _market, _outcome, _fxpAmount, _fxpPrice, _tradeGroupID);
+        return _output;
```
MicahZoltu: Style: Remove the unnecessary local variable.  Same for following functions.

src/trading/Trade.sol
```diff
@@ -0,0 +1,78 @@
...
+        uint256 _bestFxpAmount = this.takeBestOrder(_sender, _direction, _market, _outcome, _fxpAmount, _fxpPrice, _tradeGroupID);
+        if (_bestFxpAmount > 0 && msg.gas >= MINIMUM_GAS_NEEDED) {
+            return IMakeOrder(controller.lookup("makeOrder")).makeOrder(_sender, _direction, _bestFxpAmount, _fxpPrice, _market, _outcome, 0, 0, _tradeGroupID);
+        }
+        return 1;
+    }
+
+    function takeBestOrder(address _sender, Trading.TradeDirections _direction, IMarket _market, uint8 _outcome, uint256 _fxpAmount, int256 _fxpPrice, uint256 _tradeGroupID) external onlyWhitelistedCallers returns (uint256 _bestFxpAmount) {
+        // we need to take a BID if we want to SELL and we need to take an ASK if we want to BUY
+        Trading.TradeTypes _type = (_direction == Trading.TradeDirections.Selling) ? Trading.TradeTypes.Bid : Trading.TradeTypes.Ask;
```
MicahZoltu: What?!  Solidity has a ternary operator!  Ugh, I could have really used that in some other places...

src/trading/Trading.sol
```diff
@@ -10,4 +10,9 @@ library Trading {
         // FIXME: None is a placeholder to force Bid to equal 1 and Ask to equal 2, since these are their values in the Serpent files and test scripts.  After the Solidity migration is done, we should remove "None" and change Bid to 0 and Ask to 1 everywhere.
         None, Bid, Ask
     }
+    
+    enum TradeDirections {
+        // FIXME: Same as above
+        None, Buying, Selling
```
MicahZoltu: I believe this enumeration is only used in `Trade.sol`, so it should be same/easy to have it be zero indexed now rather than waiting.

src/trading/Trade.sol
```diff
@@ -0,0 +1,78 @@
...
+        IOrders _orders = IOrders(controller.lookup("Orders"));
+        bytes32 _orderID = _orders.getBestOrderId(_type, _market, _outcome);
```
MicahZoltu: Style: `orderId` (lowercase `d`).

src/trading/Trade.sol
```diff
@@ -0,0 +1,78 @@
...
+        bytes32 _output = this.trade(msg.sender, _direction, _market, _outcome, _fxpAmount, _fxpPrice, _tradeGroupID);
+        return _output;
+    }
+
+    function publicTakeBestOrder(Trading.TradeDirections _direction, IMarket _market, uint8 _outcome, uint256 _fxpAmount, int256 _fxpPrice, uint256 _tradeGroupID) external onlyInGoodTimes nonReentrant returns (uint256) {
+        uint256 _fxpAmountRemaining = this.takeBestOrder(msg.sender, _direction, _market, _outcome, _fxpAmount, _fxpPrice, _tradeGroupID);
+        return _fxpAmountRemaining;
+    }
+
+    function trade(address _sender, Trading.TradeDirections _direction, IMarket _market, uint8 _outcome, uint256 _fxpAmount, int256 _fxpPrice, uint256 _tradeGroupID) external onlyWhitelistedCallers returns (bytes32) {
```
MicahZoltu: Style: `_tradeGroupId` (lowercase `d`)

src/trading/Trade.sol
```diff
@@ -0,0 +1,78 @@
...
+        uint256 _bestFxpAmount = this.takeBestOrder(_sender, _direction, _market, _outcome, _fxpAmount, _fxpPrice, _tradeGroupID);
+        if (_bestFxpAmount > 0 && msg.gas >= MINIMUM_GAS_NEEDED) {
+            return IMakeOrder(controller.lookup("makeOrder")).makeOrder(_sender, _direction, _bestFxpAmount, _fxpPrice, _market, _outcome, 0, 0, _tradeGroupID);
+        }
```
MicahZoltu: Style: Consider inverting the if statement here so this turns into a guard clause.  This case is a bit borderline, but my general philosophy is to have the "most interesting/most common code path" be least indented, and indents for side paths (less interesting or less common).  In this case, exiting early is not particularly interesting, but making an order is.

Also, consider breaking this into multiple simpler if statements like:
```
uint256 _bestFxpAmount = takeBestOrder(...);
if (_bestFxpAmount == 0) {
  return 1;
}
if (msg.gas < MINIMUM_GAS_NEEDED) {
  return 1;
}
return IMakeOrder(...).makeOrder(...);
```

src/trading/Trade.sol
```diff
@@ -0,0 +1,78 @@
...
+        return _output;
+    }
+
+    function publicTakeBestOrder(Trading.TradeDirections _direction, IMarket _market, uint8 _outcome, uint256 _fxpAmount, int256 _fxpPrice, uint256 _tradeGroupID) external onlyInGoodTimes nonReentrant returns (uint256) {
+        uint256 _fxpAmountRemaining = this.takeBestOrder(msg.sender, _direction, _market, _outcome, _fxpAmount, _fxpPrice, _tradeGroupID);
+        return _fxpAmountRemaining;
+    }
+
+    function trade(address _sender, Trading.TradeDirections _direction, IMarket _market, uint8 _outcome, uint256 _fxpAmount, int256 _fxpPrice, uint256 _tradeGroupID) external onlyWhitelistedCallers returns (bytes32) {
+        uint256 _bestFxpAmount = this.takeBestOrder(_sender, _direction, _market, _outcome, _fxpAmount, _fxpPrice, _tradeGroupID);
```
MicahZoltu: Don't call with `this.`, instead do a local call.  Because this function is whitelisted, we can guarantee that even with a local call `msg.sender` will be whitelisted.  It is only functions intended to be called by untrusted callers that have to use the `this.` convention (which costs notably more gas).

src/trading/Trade.sol
MicahZoltu: Can you check to see if `trade` is actually called from elsewhere in the contracts?  If not, then perhaps we can just mark it as `private` and not deal with the `this.` calling convention.

I suspect it _is_ called from elsewhere in the contracts so we have to continue to follow this pattern, but would be nice to verify that.

nuevoalex:
Will address all of these!

src/trading/Trade.sol
nuevoalex: I'll add a comment to that effect for the first suggestion and will do an explicit cast.

src/trading/Trading.sol
nuevoalex: It's used in makeOrder and takeOrder. It should be really simple once they're converted.

src/trading/Trade.sol
nuevoalex: Looks like it is only called internally so I will make it so!

src/trading/Trade.sol
```diff
@@ -24,53 +24,53 @@ contract ITakeOrder {
...
+        if (msg.gas < MINIMUM_GAS_NEEDED) {
+            return bytes32(1);
+        }
+        return IMakeOrder(controller.lookup("makeOrder")).makeOrder(_sender, _direction, _bestFxpAmount, _fxpPrice, _market, _outcome, 0, 0, _tradeGroupId);
     }
 
-    function takeBestOrder(address _sender, Trading.TradeDirections _direction, IMarket _market, uint8 _outcome, uint256 _fxpAmount, int256 _fxpPrice, uint256 _tradeGroupID) external onlyWhitelistedCallers returns (uint256 _bestFxpAmount) {
-        // we need to take a BID if we want to SELL and we need to take an ASK if we want to BUY
+    function takeBestOrder(address _sender, Trading.TradeDirections _direction, IMarket _market, uint8 _outcome, uint256 _fxpAmount, int256 _fxpPrice, uint256 _tradeGroupId) internal returns (uint256 _bestFxpAmount) {
+        // we need to take a BId if we want to SELL and we need to take an ASK if we want to BUY
```
MicahZoltu: Heh, find and replace I'm guessing?  ;)

nuevoalex: haha yup

PR #286, opened 2017-09-07
MakeOrder migration

adrake33:
Migration of makeOrder.se to Solidity.

Major changes:
- The library PlaceOrder.sol was created to work around stack depth limitations.
- orderIds are now generated using sha256() instead of ripemd160().
- Changed a few hard-coded values of 10**18 to 1 ether.

Notes:
The struct in PlaceOrder includes `betterOrderId`, `worseOrderId`, and `tradeGroupId`.  These are not actually used in any of the functions in PlaceOrder, but I added them to the struct because the call to `orders.saveOrder()` was still hitting the stack depth limit.  If you have any suggestions for better workarounds, please let me know.


src/libraries/trading/PlaceOrder.sol
```diff
@@ -0,0 +1,135 @@
...
+
+        uint256 _attotokensEscrowed = 0;
+        uint256 _attosharesEscrowed = 0;
+        uint256 _attosharesToCover = _data.attoshares;
+        uint8 _numberOfOutcomes = _data.market.getNumberOfOutcomes();
+
+        // Figure out how many almost-complete-sets (just missing `outcome` share) the maker has
+        uint256 _attosharesHeld = 2**254;
+        uint8 _i = 0;
+        while (_i < _numberOfOutcomes) {
```
MicahZoltu: This should be a for loop (Serpent didn't have for loops which is why everything was a while loop).

MicahZoltu: Same comment for most other while loops.

src/libraries/trading/PlaceOrder.sol
```diff
@@ -0,0 +1,135 @@
...
+
+pragma solidity ^0.4.13;
+
+import 'ROOT/IController.sol';
+import 'ROOT/libraries/math/SafeMathUint256.sol';
+import 'ROOT/trading/Trading.sol';
+import 'ROOT/trading/IOrders.sol';
+
+
+library PlaceOrder {
```
MicahZoltu: I wonder if we can come up with a better name for this.  Fundamentally, it should be named like we would name a class that holds the items in `data`.  I think my suggestion would be to name it `Order`.  I'm a mildly worried that may conflict with other objects named Order (like in my taking branch) leading to confusion, though I think in this case it is the most appropriate name and perhaps I can integrate it into my branch when the time comes.

src/libraries/trading/PlaceOrder.sol
```diff
@@ -0,0 +1,135 @@
...
+
+    uint256 constant minOrderValue = 10**14;
+
+    struct Data {
+        // Contracts
+        IOrders orders;
+        IMarket market;
+
+        // Order
+        address sender;
```
MicahZoltu: We should probably rename this variable to `maker`.

src/libraries/trading/PlaceOrder.sol
```diff
@@ -0,0 +1,135 @@
...
+        uint8 outcome;
+        Trading.TradeTypes tradeType;
```
MicahZoltu: I wonder if we should move the enum definition for `TradeTypes` into this contract.  It feels like the right place as Bid/Ask are types of orders, and this contract is an in-memory order contract.

src/libraries/trading/PlaceOrder.sol
```diff
@@ -0,0 +1,135 @@
...
+        uint256 attoshares;
+        int256 displayPrice;
+        bytes32 betterOrderId;
+        bytes32 worseOrderId;
+        uint256 tradeGroupId;
```
MicahZoltu: I suspect we can leave this out of the object since it is only used for logging at the end, after the order is created.  If possible, I recommend doing so to keep this structure purely about the order, not metadata associated with the transaction that is creating the order.

src/libraries/trading/PlaceOrder.sol
adrake33: Ok, sounds good.  I'll change this to a for loop.

src/libraries/trading/PlaceOrder.sol
adrake33: Yeah, I considered naming it `Order` as well, but I was also concerned about potential naming conflicts.  If you don't think it's likely to pose too big of a problem, I'll go ahead and rename it to `Order`.

src/libraries/trading/PlaceOrder.sol
adrake33: Yeah, I like `maker` better.  I'll update this.

src/libraries/trading/PlaceOrder.sol
```diff
@@ -0,0 +1,135 @@
...
+            worseOrderId: _worseOrderId,
+            tradeGroupId: _tradeGroupId
+        });
+    }
+
+    //
+    // "public" functions
+    //
+
+    function placeBid(PlaceOrder.Data _data) internal returns (uint256, uint256) {
```
MicahZoltu: Rather than having this function return two values, consider mutating the object to contain the values instead.  Then the caller doesn't need to deal with finding space on the stack for them.  In this case, I believe this function is calculating how many shares/tokens need to be escrowed, then escrowing them, then returning the values.  Instead, it could be renamed to `escrowFundsForBid` and do all of the same and then save the escrowed funds as members of the structure (which would default to `0` at construction time).

The same recommendation for `placeAsk` => `escrowFundsForAsk`.

src/trading/MakeOrder.sol
```diff
@@ -0,0 +1,32 @@
...
+
+contract MakeOrder is Controlled, ReentrancyGuard {
+    using PlaceOrder for PlaceOrder.Data;
+
+    function makeOrder(Trading.TradeTypes _type, uint256 _attoshares, int256 _displayPrice, IMarket _market, uint8 _outcome, bytes32 _betterOrderId, bytes32 _worseOrderId, uint256 _tradeGroupId) onlyInGoodTimes nonReentrant external returns (bytes32 _orderId) {
+        PlaceOrder.Data memory _placeOrderData = PlaceOrder.create(controller, msg.sender, _outcome, _type, _attoshares, _displayPrice, _market, _betterOrderId, _worseOrderId, _tradeGroupId);
+
+        uint256 _fxpMoneyEscrowed;
+        uint256 _fxpSharesEscrowed;
+        if (_placeOrderData.tradeType == Trading.TradeTypes.Ask) {
```
MicahZoltu: Building on the comment above about `placeBid` and `placeAsk` functions, consider having a function on `Order` that is `escrowFunds` that contains this logic and then calls the appropriate method (`escrowFundsForBid` or `escrowFundsForAsk`) internally.  Then the caller (this code) would merely need to call `escrowFunds` and the `Order` class would do the rest.

src/libraries/trading/PlaceOrder.sol
adrake33: Ok, and the TradeDirections enums and new `getOrderTradingTypeFromMakerDirection()`/ `getOrderTradingTypeFromTakerDirection()` functions should go here as well?

src/libraries/trading/PlaceOrder.sol
adrake33: Ok, I'll try doing this, and if there aren't any stack depth issues, I'll remove it from the struct.

src/trading/MakeOrder.sol
```diff
@@ -0,0 +1,32 @@
...
+            (_fxpMoneyEscrowed, _fxpSharesEscrowed) = PlaceOrder.placeAsk(_placeOrderData);
+        } else if (_placeOrderData.tradeType == Trading.TradeTypes.Bid) {
+            (_fxpMoneyEscrowed, _fxpSharesEscrowed) = PlaceOrder.placeBid(_placeOrderData);
+        }
+
+        _orderId = _placeOrderData.orders.getOrderId(_placeOrderData.tradeType, _placeOrderData.market, _placeOrderData.attoshares, _placeOrderData.displayPrice, _placeOrderData.sender, block.number, _placeOrderData.outcome, _fxpMoneyEscrowed, _fxpSharesEscrowed);
```
MicahZoltu: Consider making this a method on the `Order` class as well, or better yet memoize `orderId` lazily.  Really we don't need the order ID until we call `saveOrder`, and we can make a method on `Order` that checks to see if `Order.orderId == 0` and if so it gets the `orderId` from the current data and if not uses whatever `orderId` is set to.  Then the `saveOrder` function can call into `orders.saveOrder` with the appropriate parameters.

With the above, this callsite would simplify to just `_orderData.saveOrder();` and then on the last line of this function we would just have `return _orderData.orderId();`

The goal with these changes is to make this function easy to read and we hide away all of the complex function calls (calls with lots of parameters) and switching logic inside the Order class functions.

src/libraries/trading/PlaceOrder.sol
MicahZoltu: That seems reasonable to me.

adrake33:
I added publicMakeOrder() back in.  It's being called in most parts of the code, but Trade.sol is calling makeOrder().  Let me know if you think these should be consolidated.  (We probably should consolidate them since publicMakeOrder() is just calling makeOrder() and passing in msg.sender.)

src/trading/MakeOrder.sol
```diff
@@ -0,0 +1,25 @@
...
+    using Order for Order.Data;
+
+    function publicMakeOrder(Order.TradeTypes _type, uint256 _attoshares, int256 _displayPrice, IMarket _market, uint8 _outcome, bytes32 _betterOrderId, bytes32 _worseOrderId, uint256 _tradeGroupId) external onlyInGoodTimes nonReentrant returns (bytes32) {
+        return this.makeOrder(msg.sender, _type, _attoshares, _displayPrice, _market, _outcome, _betterOrderId, _worseOrderId, _tradeGroupId);
+    }
+
+    function makeOrder(address _maker, Order.TradeTypes _type, uint256 _attoshares, int256 _displayPrice, IMarket _market, uint8 _outcome, bytes32 _betterOrderId, bytes32 _worseOrderId, uint256 _tradeGroupId) public onlyWhitelistedCallers returns (bytes32) {
+        Order.Data memory _orderData = Order.create(controller, _maker, _outcome, _type, _attoshares, _displayPrice, _market, _betterOrderId, _worseOrderId);
+        Order.escrowFunds(_orderData);
+        require(_orderData.orders.getAmount(_orderData.getOrderId(), _orderData.tradeType, _orderData.market, _orderData.outcome) == 0);
```
MicahZoltu: This check should be inside `saveOrder`, as validating the orderID doesn't already exist is part of the order creation process.

src/trading/Orders.sol
```diff
@@ -161,48 +152,46 @@ contract Orders is DelegationTarget, IOrders {
...
     // FIXME: Remove first and last parameters, as they are no longer used
-    function saveOrder(bytes32, Trading.TradeTypes _type, IMarket _market, uint256 _fxpAmount, int256 _fxpPrice, address _sender, uint8 _outcome, uint256 _fxpMoneyEscrowed, uint256 _fxpSharesEscrowed, bytes32 _betterOrderId, bytes32 _worseOrderId, uint256 _tradeGroupId, uint256) public onlyWhitelistedCallers returns (bytes32 _orderId) {
-        require(_type == Trading.TradeTypes.Bid || _type == Trading.TradeTypes.Ask);
+    function saveOrder(bytes32, Order.TradeTypes _type, IMarket _market, uint256 _fxpAmount, int256 _fxpPrice, address _sender, uint8 _outcome, uint256 _fxpMoneyEscrowed, uint256 _fxpSharesEscrowed, bytes32 _betterOrderId, bytes32 _worseOrderId, uint256 _tradeGroupId, uint256) public onlyWhitelistedCallers returns (bytes32 _orderId) {
         require(_outcome < _market.getNumberOfOutcomes());
         _orderId = getOrderId(_type, _market, _fxpAmount, _fxpPrice, _sender, block.number, _outcome, _fxpMoneyEscrowed, _fxpSharesEscrowed);
         insertOrderIntoList(_orderId, _type, _market, _outcome, _fxpPrice, _betterOrderId, _worseOrderId);
         orders[_orderId].fxpPrice = _fxpPrice;
```
MicahZoltu: The new `Order` object has additional fields that are not populated here.  We should make sure to fully populate the object (e.g., `orderId`) so someone can't write code that accesses uninitialized fields on the `Order`.  I think there may be a couple more besides `orderId`?

src/trading/Orders.sol
```diff
@@ -161,48 +152,46 @@ contract Orders is DelegationTarget, IOrders {
...
+        if (_orderType == Order.TradeTypes.Bid) {
             // We can't use safeSub here because it disallows subtracting negative numbers. Worst case here is an operation of 2**254 - 1 as required above, which won't overflow
-            _fill = _sharesFilled + _tokensFilled.fxpDiv(uint(orders[_orderId].fxpPrice - _market.getMinDisplayPrice()), 10**18);
-        }
-        if (_orderType == Trading.TradeTypes.Ask) {
+            _fill = _sharesFilled + _tokensFilled.fxpDiv(uint(orders[_orderId].fxpPrice - _market.getMinDisplayPrice()), 1 ether);
+        } else if (_orderType == Order.TradeTypes.Ask) {
             // We can't use safeSub here because it disallows subtracting negative numbers. Worst case here is an operation of 2**254 - 1 as required above, which won't overflow
-            _fill = _sharesFilled + _tokensFilled.fxpDiv(uint(_market.getMaxDisplayPrice() - orders[_orderId].fxpPrice), 10**18);
+            _fill = _sharesFilled + _tokensFilled.fxpDiv(uint(_market.getMaxDisplayPrice() - orders[_orderId].fxpPrice), 1 ether);
```
MicahZoltu: I know this was pre-exesting, but you shouldn't need to ever do `fxpDiv(x, 10^18)` as `fxpDiv` is already dividing by 10^18.  We can investigate later if the tests are passing, but something is fishy here.

src/trading/MakeOrder.sol
MicahZoltu: On second thought, probably would be better put into the `getOrderId` function inside the if block, that way the check is only executed once when the order ID is first generated, to make sure its not already in use.

adrake33: Ok, I'll move it to getOrderId().

src/trading/Orders.sol
adrake33: Ok, I'll add a FIXME here to remind us to revisit this.

src/trading/Orders.sol
adrake33: Yeah, I can add:
```
        orders[_orderId].orders = this;
        orders[_orderId].market = _market;
        orders[_orderId].id = _orderId;
        orders[_orderId].betterOrderId = _betterOrderId;
        orders[_orderId].worseOrderId = _worseOrderId;
```
(I'll leave out tradeType and outcome, since you said they are already in your branch.)

adrake33: Don't we need to add lines to removeOrder() for these values as well?

MicahZoltu: Yeah, probably.  We want removeOrder to clear out everything.  Though, maybe `delete orders[_orderId]` is enough?  Would need to test to see if you get a full gas refund for that operation.

adrake33: Ok, should I add a TODO for testing `delete orders[_orderId]`, or do you think this should be done before merging?

MicahZoltu: Either update `removeOrder` to clear out all members and add a TODO to investigate `delete orders[_orderId]` or test `delete orders[_orderId]` and see if it refunds gas.  Up to you which one you want to do.  😄

adrake33: I'll add a TODO to look at this next week so we can get this merged a bit sooner.

PR #287, opened 2017-09-08
Fixmes

nuevoalex:
Regarding the fix for redeeming registration tokens:

- A sentinel value couldn't be used since we don't actually tally reports, rather we have a set of markets reported on that we count
- Keeping a tally of markets that require reporting became necessary for the edge case of when 0 markets in a window ever need reporting

There are other FIXMEs to look at still but this seems like an appropriate chunk.


MicahZoltu:
(haven't looked at PR yet, just description): I believe the plan for 0 markets scenario is to create a faux market for the sake of there always being _something_ to report on.

src/reporting/Reporting.sol
```diff
@@ -0,0 +1,18 @@
+pragma solidity ^0.4.13;
+
+
+library Reporting {
+    uint256 private constant AUTOMATED_REPORTING_DURATION_SECONDS = 3 days;
+    uint256 private constant AUTOMATED_REPORTING_DISPUTE_DURATION_SECONDS = 3 days;
+    uint256 private constant REPORTING_DURATION_SECONDS = 27 * 1 days;
+    uint256 private constant REPORTING_DISPUTE_DURATION_SECONDS = 3 days;
+    uint256 private constant CLAIM_PROCEEDS_WAIT_TIME = 3 days;
+    uint256 private constant REGISTRATION_TOKEN_BOND_AMOUNT = 1 ether;
```
MicahZoltu: Is it not possible to expose these directly (without writing a function) by making them `internal` instead of `private`?  I would be OK with going against our normal patterns for this use-case (centrally defined constants) if it works.

nuevoalex: That unfortunately doesn't work :[ This method was the on;y way to make them accessible from a shared source that I could find

src/reporting/Reporting.sol
```diff
@@ -0,0 +1,18 @@
...
+
+    function automatedReportingDurationSeconds() internal constant returns (uint256) { return AUTOMATED_REPORTING_DURATION_SECONDS; }
+    function automatedReportingDisputeDurationSeconds() internal constant returns (uint256) { return AUTOMATED_REPORTING_DISPUTE_DURATION_SECONDS; }
+    function reportingDurationSeconds() internal constant returns (uint256) { return REPORTING_DURATION_SECONDS; }
+    function reportingDisputeDurationSeconds() internal constant returns (uint256) { return REPORTING_DISPUTE_DURATION_SECONDS; }
+    function claimProceedsWaitTime() internal constant returns (uint256) { return CLAIM_PROCEEDS_WAIT_TIME; }
+    function getRegistrationTokenBondAmount() internal constant returns (uint256) { return REGISTRATION_TOKEN_BOND_AMOUNT; }
```
MicahZoltu: BOND_AMOUNT will likely become dynamic at some point.  This change is fine for now, but we should figure out where the right place to do bond calculation is (probably `ReportingWindow` or `Branch`) and then move the constant to that location as a first step towards making it dynamic.

src/reporting/ReportingWindow.sol
```diff
@@ -24,10 +25,10 @@ contract ReportingWindow is DelegationTarget, Typed, Initializable, IReportingWi
     Set private markets;
     Set private limitedReporterMarkets;
     Set private allReporterMarkets;
+    uint256 private totalMarketsThatRequiredReporting;
     mapping(address => Set) private reportsByReporter;
+    mapping(address => bool) private reporterDoneReporting;
```
MicahZoltu: Consider having a single `mapping(address => ReportingStatus) reportingStatii` (come up with better name) where
```
struct ReportingStatus {
    Set marketsReportedOn;
    bool finishedReporting;
}
```
I'm curious if this would help make some of the other code for reporter tracking more readable (last I checked a while ago, the code was difficult to follow).

src/reporting/ReportingWindow.sol
```diff
@@ -90,19 +92,25 @@ contract ReportingWindow is DelegationTarget, Typed, Initializable, IReportingWi
...
-            limitedReporterMarkets.remove(_market);
-            return false;
+
+        // If we recieve an automated report we may not need any reporting. We will uppdate the counter if the market proceeds to limited reporting
+        if (_state == IMarket.ReportingState.AUTOMATED_DISPUTE) {
+            totalMarketsThatRequiredReporting -= 1;
         }
-        if (_state > IMarket.ReportingState.LIMITED_REPORTING) {
+
+        if (_state != IMarket.ReportingState.ALL_REPORTING) {
```
MicahZoltu: Consider inverting this if statement.  I find `!=` to be harder to mentally parse than `==`.

MicahZoltu: Same with the one below.

src/reporting/ReportingWindow.sol
nuevoalex: Will try this!

src/reporting/ReportingWindow.sol
```diff
@@ -216,7 +226,13 @@ contract ReportingWindow is DelegationTarget, Typed, Initializable, IReportingWi
...
 
     function getRequiredReportsPerReporterForlimitedReporterMarkets() public afterInitialized constant returns (uint256) {
-        return getTargetReportsPerLimitedReporterMarket() * limitedReporterMarkets.count() / registrationToken.totalSupply();
+        uint256 _numLimitedReporterMarkets = limitedReporterMarkets.count();
+        uint256 _requiredReports = getTargetReportsPerLimitedReporterMarket() * _numLimitedReporterMarkets / registrationToken.totalSupply();
+        // We shouldn't require more reporting than is possible.
+        if (_requiredReports > _numLimitedReporterMarkets) {
+            return _numLimitedReporterMarkets;
+        }
+        return _requiredReports;
```
MicahZoltu: Consider: This could be simplified with `return _requiredReports.min(_numLimitedReporterMarkets);`

src/reporting/ReportingWindow.sol
```diff
@@ -262,7 +278,7 @@ contract ReportingWindow is DelegationTarget, Typed, Initializable, IReportingWi
     }
 
     function isDoneReporting(address _reporter) public afterInitialized constant returns (bool) {
-        return getReportsByReporter(_reporter).count() >= getTargetReportsPerReporter();
+        return reporterDoneReporting[_reporter] || (totalMarketsThatRequiredReporting == 0);
```
MicahZoltu: For reasons*, we cannot ever have a reporting window with 0 markets being reported on, so I don't think this check is necessary.

* We need to be able to identify who showed up for reporting and who didn't so we can correctly reward people that showed up during the month, even if they didn't have to do any meaningful work beyond blindly clicking a button in the UI.

nuevoalex: Ok! For now should I just comment out tests that would fail since we allow this case? If we know roughly how we want to implement this I could also just do so.

MicahZoltu: I don't think we have a plan on the exact details of how to do this.  The theory is that if there are no markets in a reporting window when the reporting window starts, a placeholder market will be reported on with a question like, "Report YES if you want to receive reporting fees this period."  This market likely won't be able to go through the normal market creation codepath since it'll be created _after_ the reporting window starts.  Maybe now that we are using interfaces we can actually have the reporting window create an `IMarket` with a different implementation?

Alternatively we could come up with some other solution for allowing the user to sign an "I'm here" transaction for the special case of no markets to be reported on.  This may be simpler, but it would require the UI to have a separate code path for "no markets available".

Easiest solution test wise is to just (for now) have all tests include at least one market to be reported on and have one test that is skipped that has no markets to report on (since the functionality isn't implemented yet).

If you have some ideas on how to implement the faux market it may be a good next task, though it may be too big for this PR (depending on what the final solution is).

MicahZoltu: 🤔 more on it, does everything "work" with this change alone with regards to no markets?  If so, maybe we should leave this in for now until we implement the better solution later and just add a FIXME?

nuevoalex: I think for now I'll remove that logic since it'll end up being a turd and comment out the test with a note.

IMO the cleanest solution would be that every reporting window generates a special IMarket on construction that can only be reported on in the case of no reportable markets. This way no transaction is required to create the special market and it _should_ just work passively by having the appropriate functions that retrieve markets handle this.

I agree that should probably go in another PR though if that sounds reasonable.

MicahZoltu: Hmm, actually... reporting windows are only created when someone creates a market, so maybe all of this is moot since there will be no reporting window in the first place.  :)

nuevoalex: That market may not need reporting though so I think it's still a thing

tests/reporting/reporting_utils.py
```diff
@@ -0,0 +1,168 @@
...
+def proceedToAutomatedReporting(testFixture, market):
+    cash = testFixture.cash
+    branch = testFixture.branch
+    reputationToken = testFixture.applySignature('ReputationToken', branch.getReputationToken())
+    reportingWindow = testFixture.applySignature('ReportingWindow', market.getReportingWindow())
+
+    # We can't yet do an automated report on the market as it's in the pre reporting phase
+    if (market.getReportingState() == testFixture.constants.PRE_REPORTING()):
+        with raises(TransactionFailed, message="Reporting cannot be done in the PRE REPORTING state"):
+            market.automatedReport([0,2], sender=tester.k0)
```
MicahZoltu: This will only work with binary market.  Consider taking in the array as a parameter to this function so this will work with n-outcome markets and scalar markets. (same comment elsewhere in this file).

tests/reporting/reporting_utils.py
```diff
@@ -0,0 +1,168 @@
...
+    reputationToken = testFixture.applySignature('ReputationToken', branch.getReputationToken())
+
+    # Proceed to the ALL REPORTING phase
+    if (market.getReportingState() != testFixture.constants.ALL_REPORTING()):
+        proceedToAllReporting(testFixture, market, makeReport, automatedDisputer, limitedDisputer)
+
+    reportingWindow = testFixture.applySignature('ReportingWindow', market.getReportingWindow())
+
+    reportingTokenNo = testFixture.getReportingToken(market, [2,0])
+    reportingTokenYes = testFixture.getReportingToken(market, [0,2])
```
MicahZoltu: Only the `no` token is used, recommend deleting this line for getting the yes token.

PR #288, opened 2017-09-10
Porting TradingEscapeHatch to Solidity

redsquirrel:
Had to extract method to `calculateFrozenShareValues` in order to avoid stack depth issues.


src/trading/TradingEscapeHatch.sol
```diff
@@ -0,0 +1,96 @@
...
+        ICash _marketCurrency = _market.getDenominationToken();
+
+        for (uint8 _outcome = 0; _outcome < _market.getNumberOfOutcomes(); ++_outcome) {
+            IShareToken _shareToken = _market.getShareToken(_outcome);
+            uint256 _sharesOwned = _shareToken.balanceOf(msg.sender);
+            if (_sharesOwned > 0) {
+                uint256 _frozenShareValue = getFrozenShareValue(_market, _outcome);
+                _shareToken.destroyShares(msg.sender, _sharesOwned);
+                uint256 _amountToTransfer = _sharesOwned.fxpMul(_frozenShareValue, 1 ether);
+                _marketCurrency.transferFrom(_market, msg.sender, _amountToTransfer);
```
redsquirrel: If I change `_amountToTransfer` to `0`, there is no exception. If I change it to `1`, there's an exception.

nuevoalex: In Market.sol the method 'approveSpenders' needs to update the capitalization for this contract.

src/trading/TradingEscapeHatch.sol
```diff
@@ -0,0 +1,94 @@
...
+import 'ROOT/trading/IOrders.sol';
+import 'ROOT/trading/IShareToken.sol';
+import 'ROOT/libraries/DelegationTarget.sol';
+import 'ROOT/libraries/math/SafeMathUint256.sol';
+
+
+contract TradingEscapeHatch is DelegationTarget, ITradingEscapeHatch {
+    using SafeMathUint256 for uint256;
+
+    //      market             outcome  value
```
MicahZoltu: style, I'm generally not a fan of vertical alignment of text like this.  I would personally prefer
```
// market => (outcome => value)
```
What do others think?

src/trading/TradingEscapeHatch.sol
```diff
@@ -0,0 +1,94 @@
...
+                _shareToken.destroyShares(msg.sender, _sharesOwned);
+                uint256 _amountToTransfer = _sharesOwned.fxpMul(_frozenShareValue, 1 ether);
+                require(_marketCurrency.transferFrom(_market, msg.sender, _amountToTransfer));
+            }
+        }
+        return true;
+    }
+
+    function getFrozenShareValue(IMarket _market, uint8 _numOutcomes, uint8 _outcome) internal returns(uint256) {
+        require(0 <= _outcome && _outcome < _numOutcomes);
```
MicahZoltu: This can be simplified to `require(outcome < _numOutcomes)`.  We needed the `0` check before because Serpent didn't have types so this could have been negative.

src/trading/TradingEscapeHatch.sol
```diff
@@ -0,0 +1,94 @@
...
+
+        if (frozenShareValues[_market][_outcome] != 0)
+            return frozenShareValues[_market][_outcome];
```
MicahZoltu: Style: Missing curly braces around if block.

src/trading/TradingEscapeHatch.sol
```diff
@@ -0,0 +1,94 @@
...
+
+        int256 _minValue = _market.getMinDisplayPrice();
+        int256 _maxValue = _market.getMaxDisplayPrice();
+        uint256 _range = uint256(_maxValue - _minValue);
+
+        require(calculateFrozenShareValues(_market, _numOutcomes, _minValue, _range));
```
MicahZoltu: This is an internal method that will succeed or throw, so I don't think the `require` is necessary.

src/trading/TradingEscapeHatch.sol
```diff
@@ -0,0 +1,94 @@
...
+
+        return frozenShareValues[_market][_outcome];
+    }
+
+    function calculateFrozenShareValues(IMarket _market, uint8 _numOutcomes, int256 _minValue, uint256 _range) internal returns(bool) {
```
MicahZoltu: I wonder if we should rename this to `memorizeFrozenShareValues` since it doesn't return the result and instead updates state.

redsquirrel:
I'm suddenly getting a bunch of linter/indentation errors in the Travis build. Did something change? Almost all of the them are unrelated to this PR.

MicahZoltu:
Travis fixed.

MicahZoltu:
What is the status of this?  Ready to merge after resolving merge conflicts?  Or is there more work to be done?

redsquirrel:
Thanks. I thought I'd already merged that.

PR #289, opened 2017-09-11
TakeOrder migration to solidity.

MicahZoltu:
There is definitely room for cleanup/improvement, especially with regards to the somewhat duplicated data structures.  However, I decided to get this PR out as-is rather than doing that cleanup now so we can finish up the Solidity migration (we are so close).

This PR effectively rewrites the TakeOrder code to follow a more OOP design, which alleviates the stack depth issues we were running into.  It also cleans up a number of unused variables in Orders along the way, though there is certainly room for more cleanup (so many functions take `orderId` and then a bunch of details about that order).

We have discussed rewriting the trading code to follow the patterns in reporting, which further discourages me from actually finishing the cleanup work.  :)


tests/trading/test_orders.py
```diff
@@ -208,881 +208,73 @@ def test_walkOrderList_asks(contractsFixture):
...
-                "betterOrderID": longTo32Bytes(0),
-                "worseOrderID": orderId18
-            }]
-        }
-    })
+    # validate the new order was inserted correctly
+    assert orders.getBetterOrderId(insertedOrder) == longTo32Bytes(0) if where == 'best' else bestOrderId
+    assert orders.getWorseOrderId(insertedOrder) == longTo32Bytes(0) if where == 'worst' else worstOrderId
+    assert orders.getBestOrderId(orderType, market.address, YES) == insertedOrder if where == 'best' else bestOrderId
+    assert orders.getWorstOrderId(orderType, market.address, YES) == insertedOrder if where == 'worst' else worstOrderId
```
adrake33: Thanks for cleaning this up!  Far fewer lines of code now. 👍

src/trading/TakeOrder.sol
```diff
@@ -0,0 +1,393 @@
+pragma solidity ^0.4.13;
+
+import 'ROOT/trading/ITakeOrder.sol';
+import 'ROOT/Controlled.sol';
+import 'ROOT/libraries/ReentrancyGuard.sol';
+import 'ROOT/libraries/math/SafeMathUint256.sol';
+import 'ROOT/libraries/math/SafeMathInt256.sol';
```
adrake33: Is this being used?  I think you're only using SafeMathUint256.

src/trading/TakeOrder.sol
```diff
@@ -0,0 +1,393 @@
...
+
+    function getDirections(IOrders _orders, bytes32 _orderId) private constant returns (Direction _makerDirection, Direction _takerDirection) {
+        return (_orders.getTradeType(_orderId) == Order.TradeTypes.Bid)
+            ? (Direction.Long, Direction.Short)
+            : (Direction.Short, Direction.Long)
+        ; // move semicolon up a line when https://github.com/duaraghav8/Solium/issues/110 is fixed
+    }
+
+    function getTakerSharesToSell(IShareToken _longShareToken, IShareToken[] memory _shortShareTokens, address _taker, Direction _takerDirection, uint256 _takerSize) private constant returns (uint256) {
+        uint256 _sharesAvailable = 2**256-1;
```
adrake33: Perhaps instead of hard-coding this value, it should be added as a constant to SafeMathUint256.sol?

src/trading/TakeOrder.sol
```diff
@@ -0,0 +1,393 @@
...
+                _sharesAvailable = _shortShareTokens[_outcome].balanceOf(_taker).min(_sharesAvailable);
+            }
+        }
+        return _sharesAvailable.min(_takerSize);
+    }
+}
+
+
+library DirectionExtensions {
+    function toTradeType(Trade.Direction _direction) internal constant returns (Order.TradeTypes) {
```
adrake33: Do we still need this since I added getOrderTradingTypeFromTakerDirection() to Order.sol?

src/trading/TakeOrder.sol
MicahZoltu: Fixed.

src/trading/TakeOrder.sol
MicahZoltu: Fixed.

src/trading/TakeOrder.sol
MicahZoltu: At some point we need to merge this file and your `Order.sol` file, which includes merging `Direction`.  However, I'm going to put that off for a future PR, probably after we decide what we are going to do with the trading contracts in general.

PR #290, opened 2017-09-11
Updating OpenZeppelin contracts to comply with Augur coding standards

adrake33:


src/libraries/token/ERC20Basic.sol
```diff
@@ -7,8 +7,14 @@ pragma solidity ^0.4.13;
...
  */
 contract ERC20Basic {
-    uint256 public totalSupply;
-    function balanceOf(address who) constant returns (uint256);
-    function transfer(address to, uint256 value) returns (bool);
+    uint256 internal totalSupply;
+
     event Transfer(address indexed from, address indexed to, uint256 value);
+
+    function getTotalSupply() public constant returns (uint256) {
```
MicahZoltu: This function needs to remain named `totalSupply` so it is compatible with ERC20.  To achieve this, I recommend removing the `totalSupply` variable from this contract and implementing `function totalSupply(...)` in `StandardToken` using a backing variable named `supply` (to avoid name conflict with the function).  Mark it as `internal`, and you'll need to also edit `VariableSupplyToken.sol` to use the new internal variable name.

src/macros/externs.sem
```diff
@@ -21,20 +21,20 @@ extern setExtern: [getTypeName:[]:bytes32, addSetItem:[address]:int256, contains
...
+extern disputeBondTokenExtern: [allowance:[address,address]:uint256, approve:[address,uint256]:bool, balanceOf:[address]:uint256, getBondHolder:[]:address, getBondRemainingToBePaidOut:[]:uint256, getMarket:[]:address, getDisputedPayoutDistributionHash:[]:bytes32, getReputationToken:[]:address, initialize:[address,address,uint256,bytes32]:bool, getTotalSupply:[]:uint256, transfer:[address,uint256]:bool, transferFrom:[address,address,uint256]:bool, withdraw:[]:bool]
+extern registrationTokenExtern: [allowance:[address,address]:int256, approve:[address,uint256]:int256, balanceOf:[address]:uint256, getBranch:[]:int256, getPeakSupply:[]:int256, getReportingWindow:[]:int256, getReputationToken:[]:int256, initialize:[address]:int256, redeem:[]:int256, register:[]:int256, getTotalSupply:[]:int256, transfer:[address,uint256]:int256, transferFrom:[address,address,uint256]:int256]
```
MicahZoltu: Note: We no longer need to update this file anymore as we don't really have any serpent of significance left.  Once the last few contracts are migrated from Serpent to Solidity we'll be deleting this file.

adrake33: Yeah, I figured it wasn't necessary, but I did it to keep things consistent until the file gets deleted.

src/libraries/token/ERC20Basic.sol
adrake33: Ok, sounds good.  I'll make this change.

src/libraries/token/BasicToken.sol
```diff
@@ -33,4 +34,12 @@ contract BasicToken is ERC20Basic {
     function balanceOf(address _owner) public constant returns (uint256 balance) {
         return balances[_owner];
     }
+
+    function getBalance(address _address) public constant returns (uint256) {
```
MicahZoltu: This is a duplicate of `balanceOf`, so is unnecessary.  Even though `balanceOf` doesn't comply with our naming convention, I don't think it is necessary to duplicate the method _just_ to conform and it can lead to user confusion when there are multiple methods that do the same thing.

src/reporting/DisputeBondToken.sol
```diff
@@ -19,14 +19,14 @@ contract DisputeBondToken is DelegationTarget, Typed, Initializable, ERC20Basic,
...
     uint256 private bondRemainingToBePaidOut;
+    uint256 internal supply;
 
     function initialize(IMarket _market, address _bondHolder, uint256 _bondAmount, bytes32 _payoutDistributionHash) public beforeInitialized returns (bool) {
         endInitialization();
         market = _market;
         bondHolder = _bondHolder;
         disputedPayoutDistributionHash = _payoutDistributionHash;
         bondRemainingToBePaidOut = _bondAmount * 2;
-        totalSupply = 1;
```
adrake33: Since this inherits from ERC20Basic, I removed the totalSupply.  I take it this is ok to do since we are not attempting to make this an actual ERC20 token?

MicahZoltu: Yeah, we'll be removing the ERC interface off this "soon".

src/reporting/DisputeBondToken.sol
```diff
@@ -101,4 +101,8 @@ contract DisputeBondToken is DelegationTarget, Typed, Initializable, ERC20Basic,
         bondHolder = _destinationAddress;
         return true;
     }
+
+    function totalSupply() public constant returns (uint256) {
+        return supply;
+    }
```
MicahZoltu: You probably will want to delete this as well, and potentially the storage variable.

MicahZoltu: Ah, I see.  This is necessary to get it to compile.  Keep this then and just have it return a hard-coded `1`, then remove the `supply` storage variable.

adrake33: Oops, I had forgotten I'd added this when I made my earlier comment about totalSupply.  Yeah, I'll just hard-code this to return 1.

PR #291, opened 2017-09-12
More fixmes

nuevoalex:
A few more FIXMES and a first cut implementation of the fallback market for when there are no markets to report on in a window.

Even though I think this satisfies the requirements it feels a little heavy to me. I wonder if instead of a special `IMarket` stored on the `ReportingWindow` we could instead just surface a special function on the `ReportingWindow` that allows people to "check in" and satisfy their reporting requirements when this case is active.


src/reporting/Branch.sol
```diff
@@ -117,7 +118,7 @@ contract Branch is DelegationTarget, Typed, Initializable, IBranch {
         return getReportingWindowByTimestamp(block.timestamp + getReportingPeriodDurationInSeconds());
     }
 
-    function getChildBranch(bytes32 _parentPayoutDistributionHash) public returns (IBranch) {
+    function publicGetChildBranch(bytes32 _parentPayoutDistributionHash) public returns (IBranch) {
```
MicahZoltu: Recommend a name that more clearly indicates what is happening under the hood.  Maybe `getOrCreateChildBranch`?

src/reporting/Branch.sol
```diff
@@ -78,8 +78,9 @@ contract Branch is DelegationTarget, Typed, Initializable, IBranch {
         return reportingWindows[_reportingWindowId];
     }
 
-    // TODO: this has a signature conflict with the public version below.  for consistency, we should have this be the signature for the `constant` version and rename the `public` one
-    //function getChildBranch(bytes32 _parentPayoutDistributionHash) constant returns (Branch);
+    function getChildBranch(bytes32 _parentPayoutDistributionHash) constant returns (IBranch) {
```
MicahZoltu: This method (and many others in this contract) needs a visibility modifier.  `public` seems reasonable.

src/trading/Orders.sol
```diff
@@ -176,13 +176,7 @@ contract Orders is DelegationTarget, IOrders {
...
     function removeOrder(bytes32 _orderId) public onlyWhitelistedCallers returns (bool) {
         removeOrderFromList(_orderId);
-        // TODO: Test replacing the lines below with `delete orders[_orderId]` to see if we can get a full gas refund.
-        orders[_orderId].id = bytes32(0);
-        orders[_orderId].fxpPrice = 0;
-        orders[_orderId].fxpAmount = 0;
-        orders[_orderId].maker = 0;
-        orders[_orderId].fxpMoneyEscrowed = 0;
-        orders[_orderId].fxpSharesEscrowed = 0;
+        delete orders[_orderId];
```
MicahZoltu: Should I take this to mean that `delete orders[_orderId]` gives the same size gas refund as setting all struct items to `0`?

src/trading/Orders.sol
```diff
@@ -198,7 +192,7 @@ contract Orders is DelegationTarget, IOrders {
         require(_order.fxpPrice >= _order.market.getMinDisplayPrice());
         require(_order.market.getMaxDisplayPrice() + _order.market.getMinDisplayPrice() <= 2**254);
         uint256 _fill = 0;
-        // FIXME: We need to investigate why the fxpDiv() lines below are passing in our test scripts.  We shouldn't need to do fxpDiv(x, 10^18) since fxpDiv() is already dividing by 10^18.
+        // The fxpDiv below is needed to make tokensFilled be in the appropriate base for division against the fpxPrice value. fxpDiv will multiply the initial value by the final provided value before doing division.
         if (_order.tradeType == Order.TradeTypes.Bid) {
             // We can't use safeSub here because it disallows subtracting negative numbers. Worst case here is an operation of 2**254 - 1 as required above, which won't overflow
```
MicahZoltu: Feel free to knock these out as well.  Now that we have `SafeMathInt256` we should be able to do these safely.

src/trading/Orders.sol
nuevoalex: It appears to be slightly cheaper to so so

tests/conftest.py
```diff
@@ -276,10 +276,10 @@ def getReportingToken(self, market, payoutDistribution):
         reportingToken = ABIContract(self.chain, ContractTranslator(ContractsFixture.signatures['ReportingToken']), reportingTokenAddress)
         return reportingToken
 
-    def getChildBranch(self, parentBranch, market, payoutDistribution):
+    def publicGetChildBranch(self, parentBranch, market, payoutDistribution):
```
MicahZoltu: I'm guessing this was a find/replace misstep?

tests/reporting/reporting_utils.py
```diff
@@ -119,10 +119,10 @@ def finalizeForkingMarket(reportingFixture, market, finalizeByMigration, yesMigr
     reputationToken = reportingFixture.applySignature('ReputationToken', branch.getReputationToken())
 
     # The universe forks and there is now a branch where NO and YES are the respective outcomes of each
-    noBranch = reportingFixture.getChildBranch(branch, market, secondReportOutcomes)
+    noBranch = reportingFixture.publicGetChildBranch(branch, market, secondReportOutcomes)
```
MicahZoltu: Same for these (and other calls throughout tests to `fixture.getChildBranch`).

tests/conftest.py
nuevoalex: I went ahead and changed this to match the function it calls actually.

tests/reporting/reporting_utils.py
nuevoalex: Same with these.

tests/trading/test_makeOrder.py
```diff
@@ -32,7 +29,7 @@ def test_publicMakeOrder_bid(contractsFixture):
     cash.depositEther(value = 10**17)
     cash.approve(makeOrder.address, 10**17)
 
-    orderID = makeOrder.publicMakeOrder(BID, 10**18, 10**17, market.address, 1, longTo32Bytes(0), longTo32Bytes(0), 7)
+    orderID = makeOrder.publicMakeOrder(contractsFixture.constants.BID(), 10**18, 10**17, market.address, 1, longTo32Bytes(0), longTo32Bytes(0), 7)
```
MicahZoltu: Hmm, I'm not sure about this pattern.  In particular, we reference BID/ASK a _lot_ in tests and having it be this many characters in each case is going to make the tests harder to read.  I don't mind centralizing the constant (rather than putting it at the top of every file) but I would much prefer to do something like `from constants import BID, ASK, YES, NO, LONG, SHORT` so that we can continue to use the pithy names rather than having to fetch.

I recognize that this means we won't be pulling the values out of the contracts, but these values should be stable and the tests _should_ fail if someone changes BID from 0 to 1 (or whatever) so I think there is marginal value in the tests having the data hard coded rather than fetched.

nuevoalex: I'll see if I can get this to fit within a pattern like that as I agree it would be nice to have shorter versions

nuevoalex:
Did you have any thoughts on my PR description of possible alternatives to the FallbackMarket as well? If you think this is fine then all good but I want to make sure we've explicitly rejected simpler alternatives.

MicahZoltu:
I agree, the fallback market is heavy handed and it will increase the cost of creating new reporting windows by a bit (200k+ gas).  What worries me about having a checkin function is that it could lead to us having to account for that in a lot of code paths.  If you believe that we can do the checkin route without having to branch in a bunch of places, I'm not outright against it.

A checkin function would mean that the UI would also need to branch on this, which means that knowledge of the whole concept of "reporting windows with no markets" would need to propagate up to the UI.  I think this is the primary thing making me resistant to a checkin function.  I would like to constrain the scope of impact of empty reporting windows as much as possible.

nuevoalex:
As implemented now the UI still has to branch since the fallback market must be retrieved through a specific function call. We could address that by having it return though whatever method we intend client consumer to get markets to report on though.

One note is that I think this case (Reporting windows with no markets to report on) is actually something we hope will happen very often since it implies automated reporting was copacetic in all markets. I originally thought it was edge case but it may not be.

MicahZoltu:
Correct, the hope is that this will be the hotpath.  Hmm, we should definitely try to make it so the UI doesn't need to fetch the fallback market.  I believe in order to achieve this we'll need to log a market creation event (which doesn't exist yet in normal markets) that looks just like the normal market creation event.  Then when the Augur node is scanning for markets it will find this one and can present it to the user.  🤔 Of course, the problem then is that we want it to _not_ display this market if there are other markets.  Bleh.  Everything is horrible, this solution is fine for now.  Feel free to change to `claim` pattern if it makes you happy.  😄

nuevoalex:
Haha _software_. Yea I don't know exactly why but the checkin method feels like the most correct thing to me so I'll see how that looks.

tests/constants.py
```diff
@@ -0,0 +1,4 @@
+BID = 0
+ASK = 1
+LONG = 0
+SHORT = 1
```
MicahZoltu: Can be done later (or never) but there may be value in throwing YES/NO in here as well since we use them a lot.

tests/trading/test_trade.py
```diff
@@ -559,7 +554,7 @@ def test_one_ask_on_books_buy_excess_order(contractsFixture):
             "_event_type": "MakeOrder",
             "market": market.address,
             "outcome": YES,
-            "orderType": BUY,
+            "orderType": LONG,
```
MicahZoltu: This was wrong before, it should have been BID for the order type, and still be BID.  Orders have type (BID/ASK) participants have direction (SHORT/LONG).

tests/trading/test_wcl.py
```diff
@@ -10,8 +10,9 @@
 YES = 1
 NO = 0
 
-BID = 1
-ASK = 2
+BID = 0
+ASK = 1
```
MicahZoltu: Any particular reason these aren't pulling from `constants` like the other files?

nuevoalex: nope, will fix!

PR #296, opened 2017-09-14
Remove display prices

nuevoalex:
Aside from constraining the number of outcomes and the price of Orders the tick Size isn't used much. This made sense to me since it's largely a UX value but I may have been missing some part of its utility.

In particular I haven't really changed much regarding the payoutDenominator other than simplifying it in scalar markets to be a constant. I get the sense I may be misinterpreting some part of the intent here but the changes so far do seem to simplify the code a good bit.


src/trading/IOrdersFetcher.sol
```diff
@@ -6,6 +6,6 @@ import 'ROOT/reporting/IMarket.sol';
 
 contract IOrdersFetcher {
     function getOrderIds(Order.TradeTypes _type, IMarket _market, uint8 _outcome, bytes32 _startingOrderId, uint256 _numOrdersToLoad) external constant returns (bytes32[]);
-    function getOrder(bytes32 _orderId) public constant returns (uint256 _attoshares, int256 _displayPrice, address _owner, uint256 _tokensEscrowed, uint256 _sharesEscrowed, bytes32 _betterOrderId, bytes32 _worseOrderId, uint256 _gasPrice);
-    function findBoundingOrders(Order.TradeTypes _type, int256 _fxpPrice, bytes32 _bestOrderId, bytes32 _worstOrderId, bytes32 _betterOrderId, bytes32 _worseOrderId) public constant returns (bytes32, bytes32);
+    function getOrder(bytes32 _orderId) public constant returns (uint256 _attoshares, uint256 _price, address _owner, uint256 _tokensEscrowed, uint256 _sharesEscrowed, bytes32 _betterOrderId, bytes32 _worseOrderId, uint256 _gasPrice);
+    function findBoundingOrders(Order.TradeTypes _type, uint256 _fxpPrice, bytes32 _bestOrderId, bytes32 _worstOrderId, bytes32 _betterOrderId, bytes32 _worseOrderId) public constant returns (bytes32, bytes32);
```
MicahZoltu: This shouldn't be `fxpPrice` anymore.  This number should be some number between 0 and `range`.

src/reporting/IMarket.sol
```diff
@@ -26,17 +26,15 @@ contract IMarket is Typed, IOwnable {
         FINALIZED
     }
 
-    function initialize(IReportingWindow _reportingWindow, uint256 _endTime, uint8 _numOutcomes, uint256 _payoutDenominator, uint256 _feePerEthInAttoeth, ICash _cash, address _creator, int256 _minDisplayPrice, int256 _maxDisplayPrice, address _automatedReporterAddress, bytes32 _topic) public payable returns (bool _success);
+    function initialize(IReportingWindow _reportingWindow, uint256 _endTime, uint8 _numOutcomes, uint256 _payoutDenominator, uint256 _tickSize, uint256 _feePerEthInAttoeth, ICash _cash, address _creator, address _automatedReporterAddress, bytes32 _topic) public payable returns (bool _success);
```
MicahZoltu: `tickSize` should be able to be derived from `payoutDenominator` (which needs to be renamed).  The idea here is that when a market is created, it is created with a single number (currently named `_payoutDenominator`).  This number is then used both for reporting and for trading.  If a binary market has a range of `[-10,30]` (user-facing) and the author wants to go out to "2 decimal places" then the `_payoutDenominator` would be `(max - min) * decimals` or `(30 - (-10)) * 100 == 40 * 100 == 4000`.  `4000` would be the only number stored with the market and everything would be relative to that.  When a user interacts with the UI and wants to sell a share for 25.5, the UI would first normalize that to 0, bringing the value to 35.5, and then scale it to 0-4000 (rather than 0-40) bringing the value to 3550.  The contracts would then receive a trade with a "price" of 3550.

This technique allows us to not have to worry about scaling/normalizing anywhere in the contracts and instead they can focus on simple trades that are always price/max (3550/4000 in this case).  All path would then be done using the ratio 3550/4000 which will constrain all trading to a "tick size" of 0.01 as far as the user is concerned, without the contracts ever knowing anything about tick sizes.

The big overarching goal of this is to get code complexity out of the contracts and into the UI.  I _believe_ it also addresses a couple miscellaneous rounding issues that could have occurred with the assumed/forced base of `10^18`.  Also, it means there is less data we need to store on contract, since we would be normalizing the reporting base and the trading base.  The UI of course would not surface any of this to the user.

PR #297, opened 2017-09-15
Contract compiler

adrake33:
This is a class for finding all Solidity contracts in a specified directory and compiling them to a single file.


source/contracts/Apple.sol
```diff
@@ -1,6 +1,6 @@
 pragma solidity ^0.4.13;
 
-import 'ROOT/Controlled.sol';
```
MicahZoltu: I have been thinking about this a lot lately.  Solidity supports any string as the "global name" for a file.  What this means is that whatever string we use when adding the file to the standard-json input to solidity is the string that we need to use to import he file.  This means we could do something like `trading/apple.sol` or just `apple.sol` or `./trading/apple.sol`.  It also supports remappings, which is something I did originally before I really understood how the system worked.

I think remappings like we had before (`ROOT`) is the worst option.  What you have here is definitely better, though I wonder if we could save ourselves some effort by not requiring the leading `./`?  In this case, the import would be simplified to `Controlled.sol`.  This would require a minor change in the script that builds the standard-json, but that should be relatively minor.

source/libraries/CompileSolidity.ts
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env node
+
+import * as fs from 'fs';
+import * as path from 'path';
+import * as solc from 'solc';
+
+class SolidityContractCompiler {
+    private compiler: any = solc;
```
MicahZoltu: No need to have a private member that is just an alias to the import.  Recommend deleting this line and just using `solc` below.  If desired, you could change the import line to `import * as compiler from 'solc'` if you find the terminology of `compiler` to be more clear.

Personally, I like `solc` as it is more explicit but I'll leave the naming decision here up to you.

source/libraries/CompileSolidity.ts
```diff
@@ -0,0 +1,88 @@
...
+                    contractFilePaths.push(filePath);
+                }
+            }
+        }
+
+        return contractFilePaths;
+    }
+
+    private generateCompilerInput() {
+        let inputJSON: any = {
```
MicahZoltu: We should create a typing file for solc that defines the structure of this so we get type checking, static analysis and auto-complete on this object in a TypeScript aware editor and compiler errors the rest of the time.  Writing type definition files is a bit annoying because you have to figure out the right kind of definition file to author based on the library, but once you have the structure down writing the definition is pretty easy.  Let me know when you want to work on this and I can help out.

source/libraries/CompileSolidity.ts
adrake33: Ok, `solc` is fine with me since it's more explicit.

source/contracts/Apple.sol
adrake33: Ok, so would this have to be done using the `remappings` JSON parameter in my script or a different parameter?  I see in conftest.py we have:

```         
compilerParameter = {
            'language': 'Solidity',
            'sources': {
                absoluteFilePath: {
                    'urls': [ absoluteFilePath ]
                }
            },
            'settings': {
                'remappings': [ 'ROOT=%s' % resolveRelativePath("../src") ],
                'outputSelection': {
                    '*': [ 'metadata', 'evm.bytecode', 'evm.sourceMap' ]
                }
            }
        }
```
Also, I just realized I'm outputting only the bytecode to the output file, but I see here we are outputting:
```
[ 'metadata', 'evm.bytecode', 'evm.sourceMap' ]
```
I assume I'll need to change this in my script?

source/libraries/CompileSolidity.ts
adrake33: Ok, I can work on this today.  I suppose we might as well do this before moving on to creating a unit test?

source/libraries/CompileSolidity.ts
```diff
@@ -0,0 +1,88 @@
...
+        let wstream: any = fs.createWriteStream(contractOutputFilePath);
+        for (let contract in compilerOutput.contracts) {
+            wstream.write(compilerOutput.contracts[contract].bytecode);
+        }
+
+        console.log("Saved " + contractOutputFilePath);
+    }
+
+    private getContractFilePaths(directoryPath){
+        let contractFilePaths: string[] = [];
```
MicahZoltu: Style: I recommend `const` over `let` in almost all cases.  `let` should only be used if you have a variable that you _intend_ to change later on (e.g., `let i = 0` where you will later do `++i`).

source/libraries/CompileSolidity.ts
```diff
@@ -0,0 +1,88 @@
...
+        if (!fs.existsSync(directoryPath)) {
+            console.log("Invalid directory path: ", directoryPath);
+        } else {
+            let filesInCurrentDirectory: any = fs.readdirSync(directoryPath);
+            for (let file of filesInCurrentDirectory) {
+                let filePath: string = path.join(directoryPath, file);
```
MicahZoltu: The type here can be inferred from the right side of the equal sign, so it isn't necessary.  Same comment in a number of other places.  I'm generally a fan of type inference, particularly in code where auditability requirements are low.

source/libraries/CompileSolidity.ts
adrake33: Agreed, I'll change these to `const` where applicable.

source/libraries/CompileSolidity.ts
```diff
@@ -0,0 +1,88 @@
...
+        let contractOutputFilePath: string = this.contractOutputDirectoryPath + "/" + this.contractOutputFileName;
+        let wstream: any = fs.createWriteStream(contractOutputFilePath);
+        for (let contract in compilerOutput.contracts) {
+            wstream.write(compilerOutput.contracts[contract].bytecode);
+        }
+
+        console.log("Saved " + contractOutputFilePath);
+    }
+
+    private getContractFilePaths(directoryPath){
```
MicahZoltu: Consider using an NPM module rather than rolling your own directory listing function.  I haven't used it, but https://www.npmjs.com/package/recursive-readdir looks like it does what we want.  It is also async friendly, which is nice.  However, it appears that it returns absolute file pahts, so there may be value in looking to see if there is a better package that returns relative file paths so we don't have to strip off the base.

source/libraries/CompileSolidity.ts
MicahZoltu: Same comment pretty much anywhere `any` is used for a type.  It takes a bit of up-front work to add custom type definitions, but in the long run I believe it is well worth it.

source/libraries/CompileSolidity.ts
```diff
@@ -0,0 +1,88 @@
...
+    public compileContracts() {
+        // Compile all contracts in this.inputDirectoryPath
+        let inputJSON: any = this.generateCompilerInput();
+        let compilerOutput: any = this.compiler.compile(inputJSON);
+        if (compilerOutput.errors.length > 0) {
+            console.log("The compiler encountered the following warnings/errors:");
+            for (let error of compilerOutput.errors) {
+                console.log(error);
+            }
+            // Do not return automatically in case there are only warnings and no errors
```
MicahZoltu: I think compilation should probably fail on both warnings and errors, so I'm OK with just throwing here.  Also, we may want to consider using a custom Error type that includes the compiler errors and then letting the caller decide how they want to deal with compiler errors.  It _feels_ like this is the wrong place to print the errors (what if the caller wants to write errors to a file, or wants to redirect errors to stderr instead of stdout).

source/libraries/CompileSolidity.ts
```diff
@@ -0,0 +1,88 @@
...
+    public constructor(contractInputDirectoryPath: string, contractOutputDirectoryPath: string, contractOutputFileName: string) {
+        this.contractInputDirectoryPath = contractInputDirectoryPath;
+        this.contractOutputDirectoryPath = contractOutputDirectoryPath;
+        this.contractOutputFileName = contractOutputFileName;
+    }
+
+    public compileContracts() {
+        // Compile all contracts in this.inputDirectoryPath
+        let inputJSON: any = this.generateCompilerInput();
+        let compilerOutput: any = this.compiler.compile(inputJSON);
```
MicahZoltu: I believe the recommendation is to use `compiler.compileStandardWrapper(inputJson)`.  `compile` is a legacy mechanism for compiling and not recommended going forward.

source/libraries/CompileSolidity.ts
```diff
@@ -0,0 +1,88 @@
...
+        if (compilerOutput.errors.length > 0) {
+            console.log("The compiler encountered the following warnings/errors:");
+            for (let error of compilerOutput.errors) {
+                console.log(error);
+            }
+            // Do not return automatically in case there are only warnings and no errors
+        }
+
+        // Create output directory (if it doesn't exist) and save contract bytecodes to single file
```
MicahZoltu: The output will include more than just bytecode, it will also include ABI, metadata, potentially the AST, etc.  You can configure all of this via the input JSON, but for now lets go with defaults and see what we get.

source/libraries/CompileSolidity.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    public constructor(contractInputDirectoryPath: string, contractOutputDirectoryPath: string, contractOutputFileName: string) {
+        this.contractInputDirectoryPath = contractInputDirectoryPath;
+        this.contractOutputDirectoryPath = contractOutputDirectoryPath;
+        this.contractOutputFileName = contractOutputFileName;
+    }
+
+    public compileContracts() {
+        // Compile all contracts in this.inputDirectoryPath
+        let inputJSON: any = this.generateCompilerInput();
```
MicahZoltu: Style: recommend camelCase here which means `inputJson` (lower last 3 letters)

source/libraries/CompileSolidity.ts
```diff
@@ -0,0 +1,88 @@
...
+            sources: {}
+        };
+        let filePaths: string[] = this.getContractFilePaths(this.contractInputDirectoryPath);
+        for (let filePath of filePaths) {
+            inputJSON.sources[filePath] = fs.readFileSync(filePath, 'utf8');
+        }
+        return inputJSON;
+    }
+
+    private createOutputDirectory() {
```
MicahZoltu: Same comment as above, I have to believe there is an NPM module that will do directory creation in one-line rather than having to roll our own.

source/contracts/Apple.sol
MicahZoltu: I don't think we shold use remappings anymore.  I did that because I didn't understand how file lookups in solc worked and it fixed my problem at the time.  Now that I understand better, I think it isn't the right choice.

Lets go with just the default outputs for now.  I don't know what they are, but I don't think it is bytecode only by default?

source/libraries/CompileSolidity.ts
MicahZoltu: Yeah, writing tests will be easier when we have strong types.

source/libraries/CompileSolidity.ts
adrake33: Yeah, I agree.  Error handling is something I figured we'd need to discuss further, so I hadn't spent a lot of time on it.

source/libraries/CompileSolidity.ts
adrake33: I originally tried using `compiler.compileStandardWrapper()`, but I was getting the following error from solc:

```
{"errors":[{"component":"general","formattedMessage":"* Line 1, Column 1\n  Syntax error: value, object or array expected.\n","message":"* Line 1, Column 1\n  Syntax error: value, object or array expected.\n","severity":"error","type":"JSONError"}]}
```

I'll investigate further and try to see what's causing it.

source/libraries/CompileSolidity.ts
adrake33: Ok, I wasn't sure if I should try to minimize dependencies on other modules.  I'll look into incorporating recursive-readdir.

source/contracts/Apple.sol
adrake33: No, I think by default it outputs everything.  The fields I see are: `assembly`, `bytecode`, `functionHashes`, `gasEstimates`, `interface`, `metadata`, `opcodes`, `runtimeBytecode`, `srcmap`, & `srcmapRuntime`.

adrake33:
I've made the requested changes and made a few changes/additions to the type definitions, so please review these.  Once everything is merged, I will work on revising the script so that it reads in all of the files before starting the compile (since we found out that solc doesn't support async read callbacks).

package.json
```diff
@@ -1,5 +1,14 @@
...
+        "@types/mkdirp": "^0.5.1",
+        "@types/node": "^8.0.28",
+        "solc": "^0.4.16",
         "solium": "github:AugurProject/solium"
+    },
+    "dependencies": {
+        "fs": "0.0.1-security",
+        "mkdir-recursive": "^0.3.0",
+        "mkdirp": "^0.5.1",
+        "recursive-readdir": "^2.2.1"
```
MicahZoltu: We should pick whether to do all `dependencies` or all `devDependencies`.  For this project, where nothing is "deployed" like a web server or website, I don't believe it matters which we use.

source/libraries/CompileSolidity.ts
```diff
@@ -0,0 +1,88 @@
...
+            // Output contract data to single file
+            const contractOutputFilePath = this.contractOutputDirectoryPath + "/" + this.contractOutputFileName;
+            let wstream: any = fs.createWriteStream(contractOutputFilePath);
+            for (let contract in compileOutputJson.contracts) {
+                wstream.write(JSON.stringify(compileOutputJson.contracts[contract]));
+            }
+
+            return { output: "Contracts in " + this.contractInputDirectoryPath + " were successfully compiled by solc and saved to " + contractOutputFilePath};
+        } catch (error) {
+            return { error: error.message };
```
MicahZoltu: I personally find it better to let the exception bubble than to convert it to a return value.  This allows the caller to easily handle failures separately from handling successes.  It also allows you to return a simpler object (e.g., just a string in this case).

source/libraries/CompileSolidity.ts
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+
+    public readCallback(path: string): { contents?: string, error?: string } {
+        try {
+            const result = fs.readFileSync(path, 'utf8');
+            return { contents: result };
+        } catch (error) {
+            console.log(error);
```
MicahZoltu: I believe this error will get double logged?  Once here and once somewhere up the call stack?

adrake33: Yeah, I meant to take that out.  Will remove it.

source/libraries/CompileSolidity.ts
```diff
@@ -0,0 +1,88 @@
...
+                contractInputDirectoryPath += path.sep;
+            }
+
+            const files: any = await recursiveReadDir(this.contractInputDirectoryPath, [this.ignoreFile]);
+            for (let index in files) {
+                inputJson.sources[files[index].replace(contractInputDirectoryPath, "")] = {"urls": [files[index]]};
+            }
+        } catch (error) {
+            inputJson.error = error.message;
+        }
```
MicahZoltu: Same comment here as above, by returning the error rather than letting the error bubble, the caller can't easily handle failure vs success separately.  In the case of the caller of this method, they won't want to call `compile` if this function fails, and they probably will want to bail completely.

My general strategy for error handling is to let thrown errors bubble as high up as necessary until they can be cleanly handled.  Often this means letting the errors bubble all the way up the stack and crash the program, because there is no obvious mechanism for recovery.  This is OK because Node will show the error to the user and in the case of scripts like this, a developer will generally be the one running them and can take action to remedy the error and try again.

source/libraries/CompileSolidity.ts
adrake33: Ok, I'll throw it from here instead.

typings/mkdirp.d.ts
```diff
@@ -0,0 +1,9 @@
+import * as fs from "fs";
+declare namespace MkDirp {
+    interface mkdirp {
+        (path: string): Promise<string[]>;
+    }
+}
+
+declare var mkdirp: MkDirp.mkdirp;
+export = mkdirp;
```
MicahZoltu: You have `"@types/mkdirp": "^0.5.1",` in `pagkace.json`, which means you shouldn't need this file?  In fact, I suspect it conflicts and only one of the two is being used (I'm not sure which).

typings/compileSolidity.d.ts
```diff
@@ -0,0 +1,6 @@
+declare module 'compileSolidity' {
+    interface CompileContractsOutput {
+        output?: string;
+        error?: string[];
+    }
+}
```
MicahZoltu: Generally speaking, `.d.ts` files are only used for defining types of JavaScript modules.  In this case, you can just put the `interface CompileContractsOutput { ... }` at the top of your script's source file, or in another file in `source/libraries` if that makes more sense to you.  Regular `.ts` files are generally much easier to work with.

If you decide to go with a separate file for this (note: I'm not implying that is the right solution, I leave that to you) then the file would look something like:
```
interface CompileContractsOutput {
    output?: string;
    error?: string[];
}
```
And then in any other file in your project you would do:
```
import { CompileContractsOutput } from './CompileContractsOutput.ts'; // path is relative to directory this file is in
...
```

package.json
adrake33: I suppose we can just stick with devDependencies since it doesn't seem like it will make a difference.  I also noticed I'm including `mkdir-recursive`, which I don't use, so I'll remove it.

I haven't used node much before (my JS experience is primarily just on the front-end) -- do we actually need to be including `fs` here?  Isn't it a default module like `path`?  Also, can I just use 
```
"mkdirp": "^0.5.1",
```
and get rid of:
```
"@types/mkdirp": "^0.5.1",
```
What is the difference?

source/libraries/CompileSolidity.ts
adrake33: Sounds good.  That's how I've typically done error handling, but I saw you were returning it within a JSON object in `readCallback()`.  I guess this is what solc is expecting from the callback function, and there's no other way to do it?

typings/compileSolidity.d.ts
adrake33: Ok, good to know.  Thanks for the clarification.  I'll move the interface to the `.ts` file, since that's probably the only place it will be used.

typings/mkdirp.d.ts
adrake33: Alright, I'll remove `"@types/mkdirp": "^0.5.1"`

package.json
MicahZoltu: `"mkdirp": "^0.5.1",` is Node's way of saying the project depends on the `mkdirp` package at runtime.

`"@types/mkdirp": "^0.5.1",` is TypeScript's way of pulling in definitions for the `mkdirp` package so that the compiler doesn't error when you try to use it at compile time.  When working with a JavaScript package like `mkdirp`, you either need the package + one or more `.d.ts` files _OR_ you need the package + `@types/<package>`.  Basically, `@types/<package>` just pulls in `.d.ts` files that someone else created for some JavaScript package.

If the package is natively TypeScript or it includes its own `.d.ts` files then you need neither the @types nor a `.d.ts` file.  Most packages on NPM are JS without d.ts files, so you'll _usually_ need one of the two.

----

I don't believe you need to depend on `fs`.  I think that is actually just a placeholder package on NPM to prevent attackers from replacing node's `fs` library with something malicious and pulling it down as a dependency doesn't do anything.  You can safely remove that line.

source/libraries/CompileSolidity.ts
MicahZoltu: Exactly correct.  The only reason we are returning that `{contents;errors}` object from the `readCallback` is because that is what `solc` needs.  In all other cases I'm a fan of throwing.  Of course, follow whatever pattern you are most comfortable with, even if I include some different pattern in a code snippet or comment, you will be the owner of this project so it should be written in your image not mine.

In general, I'll often give feedback on a bunch of things but in most cases you can consider it just "friendly feedback" and not a requirement.  If something is a hard requirement I'll usually flag the PR as needs changes and mention which points of feedback in particular need to be addressed.  This is usually limited to things that are simply wrong/incorrect.

package.json
adrake33: Ok, thanks for the clarification.  It sounds like I need `"mkdirp": "^0.5.1",` and `"@types/mkdirp": "^0.5.1",`.

PR #298, opened 2017-09-17
Deletes collection libraries, topics and remaining Serpent code.

MicahZoltu:
With the decision to require an Augur node, Topics can be deleted.  This has a cascading effect of allowing us to delete some other stuff, including some collection libraries.  In the process, I deleted any collection libraries that were unused and migrated the single remaining used collection library over to Solidity (in the form of a struct with extensions).

Migrates the placeholder ReportingEscapeHatch to Solidity.

Removes the last of the solidity code (some macros and miscellaneous tests).


src/libraries/ComparorUint256.sol
```diff
@@ -0,0 +1,16 @@
+pragma solidity ^0.4.13;
+import "ROOT/libraries/DelegationTarget.sol";
+
+
+contract ComparorUint256 is DelegationTarget {
+
```
nuevoalex: Style nit: Empty lines after the contract definition and before it closes

README.md
```diff
@@ -65,104 +60,38 @@ py.test -s tests/trading_tests.py
...
+
+The reentrancy guard (libraries/ReentrancyGuard.sol) and controller (Controller.sol) contracts also help uphold security. One can also attack Ethereum contracts by creating a chain of wanted contract calls through a malicious user. The mutex contract prevents against that by preventing someone from re-entering the contracts before they return. The controller contract serves to update Augur and stores a registry of whitelisted contract addresses. The controller also can perform administrative actions, such as transferring ownership and performing emergency stop functions. It can be operated in either the developer mode, where the developer updates contracts, or the decentralized mode, when people must upload contracts to make changes and therefore, the market creators/traders must cancel orders and shift to the new contract. For bugs found involving money or REP, contracts are locked / an escape hatch is enabled which allows withdrawal of shares [and cancelling orders] for a value half-way between the bid-ask spread at the time of the lock. In addition, another escape hatch is enabled, which allows withdrawal of any locked up REP.
```
adrake33: Small nit, this sentence sounds a bit repetitive: 
```
The mutex contract prevents against that by preventing someone from re-entering the contracts before they return.```
Recommend changing it to something like: 
```The mutex contract protects against that by preventing someone from re-entering the contracts before they return.```

File: src/libraries/ComparorUint256.sol (hunk 1 above)
Comment (MicahZoltu) [reply src/libraries/ComparorUint256.sol]: Hmm... this file should have been deleted...

Comment (MicahZoltu) [reply src/libraries/ComparorUint256.sol]: Fixed via deletion of file.

File: README.md (hunk 2 above)
Comment (MicahZoltu) [reply README.md]: Fixed.

File: tests/solidity_test_helpers/ControllerUser.sol [hunk 3]
```diff
@@ -0,0 +1,17 @@
+pragma solidity ^0.4.13;
+
+import 'ROOT/IControlled.sol';
+
+
+contract ControllerUser is IControlled {
+    address public suicideFundsDestination;
+    address public updatedController;
+
+    function setController(IController _controller) public returns(bool) {
```
adrake33: Shouldn't these be explicitly returning true since the return type is set to bool?  (Also, small nit: there should be a space after `returns`.)

MicahZoltu: Fixed.

PR #299, opened 2017-09-18
Remove display prices 2

nuevoalex:
I currently have the market creation extension taking in `numTicks` and in the case of scalars taking in `priceRange`. They could both just take in `marketDenominator` alternatively but that seemed like a less clear API.

After discussing it further it turns out these changes do involve changing shares, at least in terms of the quantities dealt with.

Many tests had to change since we fundamentally end up changing the price and payout of a share with this change. 1 ETH in our binary market example used to buy 10**18 complete sets, whereas now it will buy a single indivisible complete set. With this in mind I also change the ShareToken to no longer have decimal precision. This is (I think) just a semantic but we're really not allowing partial shares anymore so this seemed appropriate.


src/macros/externs.sem
```diff
@@ -0,0 +1,51 @@
+extern branchFactoryExtern: [createBranch:[address,address,bytes32]:int256, getLastBranch:[]:int256]
```
MicahZoltu: This file was deleted, I'm guessing it got brought back during a merge/rebase?

src/trading/CompleteSets.sol
```diff
@@ -48,11 +49,12 @@ contract CompleteSets is Controlled, ReentrancyGuard, ICompleteSets {
         uint8 _numOutcomes = _market.getNumberOfOutcomes();
         ERC20 _denominationToken = _market.getDenominationToken();
         uint256 _marketCreatorFeeRate = _market.getMarketCreatorSettlementFeeInAttoethPerEth();
-        uint256 _marketCreatorFee = _amount.mul(_marketCreatorFeeRate).div(1 ether);
+        uint256 _payout = _amount.mul(_market.getMarketDenominator());
```
MicahZoltu: If we can spare the local variable, we should read `_market.getMarketDenominator()` once and reuse it.  Reading from storage is ~200 gas and calling a cross-contract function is 700 gas or something.

src/trading/IOrders.sol
```diff
@@ -5,31 +5,31 @@ import 'ROOT/reporting/IMarket.sol';
 
 
 contract IOrders {
-    function saveOrder(Order.TradeTypes _type, IMarket _market, uint256 _fxpAmount, int256 _fxpPrice, address _sender, uint8 _outcome, uint256 _fxpMoneyEscrowed, uint256 _fxpSharesEscrowed, bytes32 _betterOrderId, bytes32 _worseOrderId, uint256 _tradeGroupId) public returns (bytes32 _orderId);
+    function saveOrder(Order.TradeTypes _type, IMarket _market, uint256 _fxpAmount, uint256 _price, address _sender, uint8 _outcome, uint256 _moneyEscrowed, uint256 _sharesEscrowed, bytes32 _betterOrderId, bytes32 _worseOrderId, uint256 _tradeGroupId) public returns (bytes32 _orderId);
```
MicahZoltu: I wonder if we should come up with a better name for the `price` variable?  In the reporting system we called the denominator `payoutDenominator` and the numerator `payoutNumerator` and the two combined to be a payout multiplier (as a rational number).  Similarly here, the denominator (`marketDenominator`) combined with `<TBD>` results in a price multiplier (as a rational number).  I'm worried that someone looking at the name `price` won't have much of a chance of guessing what they need to put here.  What they need is a number between `0` and `marketDenominatory` that represents the distribution of share value between maker and taker.

src/trading/Order.sol
```diff
@@ -102,20 +102,20 @@ library Order {
...
+        return _orderData.orders.saveOrder(_orderData.tradeType, _orderData.market, _orderData.amount, _orderData.price, _orderData.maker, _orderData.outcome, _orderData.moneyEscrowed, _orderData.sharesEscrowed, _orderData.betterOrderId, _orderData.worseOrderId, _tradeGroupId);
     }
 
     //
     // Private functions
     //
 
     function escrowFundsForBid(Order.Data _orderData) private returns (bool) {
-        uint256 _orderValueInAttotokens = SafeMathUint256.fxpMul(_orderData.fxpAmount, SafeMathUint256.sub(_orderData.market.getCompleteSetCostInAttotokens(), uint256(_orderData.fxpPrice)), 1 ether);
+        uint256 _orderValueInAttotokens = _orderData.amount.mul(SafeMathUint256.sub(_orderData.market.getMarketDenominator(), _orderData.price));
```
MicahZoltu: Recommend using the extension method syntax here for the subtraction, and also recommend chaining over nesting:
```
_orderData.market.getMarketDenominator().sub(_orderData.price).mul(_orderData.amount)
```
Alternatively, without chaining (but still with extension calling convention):
```
_orderData.amount.mul(_orderData.market.getMarketDenominator().sub(_orderData.price))
```

src/trading/Order.sol
```diff
@@ -129,46 +129,46 @@ library Order {
...
+            _orderData.sharesEscrowed = SafeMathUint256.min(_attosharesHeld, _attosharesToCover);
+            _attosharesToCover -= _orderData.sharesEscrowed;
+            _shareToken.transferFrom(_orderData.maker, _orderData.market, _orderData.sharesEscrowed);
         }
 
         // If not able to cover entire order with shares alone, then cover remaining with tokens
         if (_attosharesToCover > 0) {
-            _orderData.fxpMoneyEscrowed = SafeMathUint256.fxpMul(_attosharesToCover, SafeMathUint256.sub(_orderData.market.getCompleteSetCostInAttotokens(), uint256(_orderData.fxpPrice)), 1 ether);
-            require(_orderData.market.getDenominationToken().transferFrom(_orderData.maker, _orderData.market, _orderData.fxpMoneyEscrowed));
+            _orderData.moneyEscrowed = _attosharesToCover.mul(SafeMathUint256.sub(_orderData.market.getMarketDenominator(), _orderData.price));
```
MicahZoltu: Same as comment above about extension calling convention and chaining over nesting.

src/trading/ShareToken.sol
```diff
@@ -49,4 +49,11 @@ contract ShareToken is DelegationTarget, Typed, Initializable, VariableSupplyTok
     function isShareToken() constant public returns(bool) {
         return true;
     }
+
+    function transfer(address _to, uint256 _value) public returns(bool) {
```
MicahZoltu: This implementation looks identical to the one on `BasicToken` (which is a base contract for `ShareToken`).  What is the reason for the override here?

src/trading/TakeOrder.sol
```diff
@@ -326,11 +323,10 @@ library Trade {
...
     function getSharePriceDetails(IMarket _market, IOrders _orders, bytes32 _orderId) private constant returns (uint256 _sharePriceRange, uint256 _sharePriceLong, uint256 _sharePriceShort) {
-        int256 _maxDisplayPrice = _market.getMaxDisplayPrice();
-        int256 _minDisplayPrice = _market.getMinDisplayPrice();
-        int256 _orderDisplayPrice = _orders.getPrice(_orderId);
-        _sharePriceRange = uint256(_maxDisplayPrice - _minDisplayPrice);
-        _sharePriceLong = uint256(_orderDisplayPrice - _minDisplayPrice);
+        uint256 _maxDisplayPrice = _market.getMarketDenominator();
+        uint256 _orderDisplayPrice = _orders.getPrice(_orderId);
+        _sharePriceRange = _maxDisplayPrice;
+        _sharePriceLong = _orderDisplayPrice;
```
MicahZoltu: These variable names no longer seem correct.  The first is `marketDenominator` and the second is `orderPrice`.  Also, no need for the intermediate variables here anymore I think, we can just directly assign to `_sharePriceRange` and `_sharePriceLong`.

src/trading/TradingEscapeHatch.sol
```diff
@@ -87,7 +82,7 @@ contract TradingEscapeHatch is DelegationTarget, ITradingEscapeHatch {
             // FIXME: Think about this math, can shiftedPrice * range be greater than 2*254?
             //        Can shiftedPrice / _sumOfBids lead to rounding errors?  Should we * then / or / then *?
             //        Yes it can be greater, shifted / _sumOfBids will lead to rounding errors
-            frozenShareValues[_market][_tempOutcome] = _shiftedPrices[_tempOutcome].fxpMul(_range, 1 ether).fxpDiv(_sumOfBids, 1 ether);
+            frozenShareValues[_market][_tempOutcome] = _shiftedPrices[_tempOutcome].fxpMul(_market.getMarketDenominator(), 1 ether).fxpDiv(_sumOfBids, 1 ether);
```
MicahZoltu: Is it intentional that this still does a `fxpMul` with a base of `1 ether`?  I thought with the changes in this PR we didn't have any more 1 ether base fixed point math being done?

tests/conftest.py
```diff
@@ -279,19 +279,19 @@ def getOrCreateChildBranch(self, parentBranch, market, payoutDistribution):
...
         assert marketAddress
         market = ABIContract(self.chain, ContractTranslator(ContractsFixture.signatures['Market']), marketAddress)
         return market
 
-    def createScalarMarket(self, branch, endTime, feePerEthInWei, denominationToken, minDisplayPrice, maxDisplayPrice, automatedReporterAddress, topic):
+    def createScalarMarket(self, branch, endTime, feePerEthInWei, denominationToken, priceRange, automatedReporterAddress, topic):
         marketCreationFee = self.contracts['MarketFeeCalculator'].getValidityBond(branch.getCurrentReportingWindow()) + self.contracts['MarketFeeCalculator'].getTargetReporterGasCosts()
-        marketAddress = self.contracts['MarketCreation'].createScalarMarket(branch.address, endTime, feePerEthInWei, denominationToken.address, minDisplayPrice, maxDisplayPrice, automatedReporterAddress, topic, value = marketCreationFee)
+        marketAddress = self.contracts['MarketCreation'].createScalarMarket(branch.address, endTime, feePerEthInWei, denominationToken.address, priceRange, 10 ** 18, automatedReporterAddress, topic, value = marketCreationFee)
```
MicahZoltu: I'm curious, if we change this to `10^6` do all of the tests break?  If so, do they break for complex reasons or just because we use `fix(2)` and such everywhere?

At some point we should pick a more reasonable base to test against, though I recognize that it will require touching a _lot_ of tests and is worthy of a separate PR and possibly put off for a bit.

tests/libraries/math/test_safeMathUint256.py
```diff
@@ -126,3 +126,24 @@ def test_fxpDiv(a, b, base, expectedResult, testerContractsFixture):
...
+])
+def test_isMultipleOf(a, b, expectedResult, testerContractsFixture):
+    testerContractsFixture.uploadAndAddToController("../tests/solidity_test_helpers/SafeMathUint256Tester.sol")
+    safeMathUint256Tester = testerContractsFixture.contracts['SafeMathUint256Tester']
+    if (expectedResult == "TransactionFailed"):
+        with raises(TransactionFailed):
+            safeMathUint256Tester.isMultipleOf(a, b)
+    else:
+        assert safeMathUint256Tester.isMultipleOf(a, b) == expectedResult
+
```
MicahZoltu: Major nit: There are a bunch of spaces on the last line of the file.  😄

tests/reporting/test_dispute_bond_token.py
```diff
@@ -20,9 +20,9 @@
 MARKET_TYPE_CATEGORICAL = 0
 MARKET_TYPE_SCALAR = 1
 
-CATEGORICAL_OUTCOME_A = [3, 0, 0]
-CATEGORICAL_OUTCOME_B = [0, 3, 0]
-CATEGORICAL_OUTCOME_C = [0, 0, 3]
+CATEGORICAL_OUTCOME_A = [3*10**17, 0, 0]
```
MicahZoltu: This shouldn't be 3... in fact, 10^18 isn't divisible by 3 so a 3-outcome categorical market cannot have 10^18 as a base (or 10^anything).  This was the thing Joey mentioned about how tick sizes for odd numbered markets will be "odd".

src/macros/externs.sem
nuevoalex: must've. Will re-delete

src/trading/IOrders.sol
nuevoalex: `marketPrice` maybe?

src/trading/ShareToken.sol
nuevoalex: Woops. This was left in from debugging. Will remove.

src/trading/TradingEscapeHatch.sol
nuevoalex: Yea I missed this. Will change

tests/reporting/test_dispute_bond_token.py
nuevoalex: The payout numerator here is 3*10**17, which is what the marketDenominator now is for the categorical market. I'm not sure why 10^18 is involved?

tests/conftest.py
nuevoalex: It would break many tests that use the scalar market yea. It's not very difficult to change but it is time consuming since it's slightly more complicated than a find replace operation.

PR #301, opened 2017-09-19
removing some unused imports from tests

redsquirrel:
I was working on getting `test_wcl_fuzzy.py` to work, when I noticed that it was the last place we're using `bytesToLong`, yet we still import it in many tests. So I went through and removed a bunch of unused imports in tests.


MicahZoltu:
Needs a rebase.

redsquirrel:
Yep, I'll get to it.

PR #302, opened 2017-09-19
Read in Solidity contract contents to input JSON 

adrake33:


source/libraries/CompileSolidity.ts
```diff
@@ -24,12 +25,12 @@ class SolidityContractCompiler {
     public async compileContracts(): Promise<CompileContractsOutput> {
         try {
             // Compile all contracts in the specified input directory
-            const inputJson: CompilerInput = await this.generateCompilerInput();
-            const compilerOutput: any = compileStandardWrapper(JSON.stringify(inputJson), this.readCallback);
-            const compileOutputJson = JSON.parse(compilerOutput);
-            if (compileOutputJson.errors) {
+            const compilerInputJson: CompilerInput = await this.generateCompilerInput();
+            const compilerOutput: any = compileStandardWrapper(JSON.stringify(compilerInputJson), this.readCallback);
```
MicahZoltu: Recommend removing the `any` here. `compileStandarWrapper` should be returning a non-any type (`CompilerOutput`?).  If it isn't, we should add that.  In general avoiding `any` types is a good idea as it gives you compile time errors (which are easy to fix) rather than runtime exceptions.

source/libraries/CompileSolidity.ts
```diff
@@ -24,12 +25,12 @@ class SolidityContractCompiler {
     public async compileContracts(): Promise<CompileContractsOutput> {
         try {
             // Compile all contracts in the specified input directory
-            const inputJson: CompilerInput = await this.generateCompilerInput();
-            const compilerOutput: any = compileStandardWrapper(JSON.stringify(inputJson), this.readCallback);
-            const compileOutputJson = JSON.parse(compilerOutput);
-            if (compileOutputJson.errors) {
+            const compilerInputJson: CompilerInput = await this.generateCompilerInput();
```
MicahZoltu: No need for the type here, it should be inferred from the return value of `generateCompilerInput`.

source/libraries/CompileSolidity.ts
```diff
@@ -24,12 +25,12 @@ class SolidityContractCompiler {
...
+            const compilerOutput: any = compileStandardWrapper(JSON.stringify(compilerInputJson), this.readCallback);
+            const compilerOutputJson = JSON.parse(compilerOutput);
```
MicahZoltu: The names here should probably be swapped.  `compileStandardWrapper` returns a JSON string, `JSON.parse(...)` turns that into a native JS object.

typings/solc.d.ts
```diff
@@ -1,7 +1,7 @@
 declare module 'solc' {
     interface CompilerInputSourceFile {
         keccak256?: string;
-        urls: string[];
+        content: string;
```
MicahZoltu: It looks like I got the definition here wrong, _either_ `urls` or `content` is allowed so the definition should be:
```
interface CompilerInputSourceFile {
    keccak256?: string;
    urls: string[];
}
interface CompilerInputSourceCode {
    keccak256?: string;
    content: string;
}
interface CompilerInput {
    language: "Solidity" | "serpent" | "lll" | "assembly";
    sources: {
        [globalName: string]: CompilerInputSourceFile|CompilerInputSourceCode,
    }
}
```

The key to this working is the union type on the `sources` object values.  It is saying that for any key in `sources`, the value can be _either_ a `CompilerInputSourceFile` or a `CompilerInputSourceCode` object.

source/libraries/CompileSolidity.ts
adrake33: Ok, or maybe the names `compilerOutput` for the first one and `compilerOutputObj` for the second would be clearest?

typings/solc.d.ts
adrake33: Oops, yeah I knew it was the case that either urls or content is allowed, but I didn't catch this either.  Will update this.

source/libraries/CompileSolidity.ts
MicahZoltu: That works.  I would personally go with `compilerOutputJson` for the first and `compilerOutput` for the second though.  In general, anything that is a native JSON object I refer to by the base name, and the stringified JSON form of that same object I refer to with a Json suffix.

source/libraries/CompileSolidity.ts
adrake33: Yeah, we don't have CompilerOutput defined, but I'll work on adding that and let you know if I have questions.

adrake33:
@micah Please look over the new CompilerOutput type definitions I've added and let me know if you see anything that ought to be changed.

source/libraries/CompileSolidity.ts
```diff
@@ -1,10 +1,11 @@
 #!/usr/bin/env node
 
 import * as fs from 'fs';
+import * as readFile from 'fs-readfile-promise';
 import * as mkdirp from 'mkdirp';
 import * as path from 'path';
 import * as recursiveReadDir from 'recursive-readdir';
-import { CompilerInput, CompilerInputSourceCode, CompilerInputSourceFile, compileStandardWrapper } from 'solc';
+import { CompilerInput, CompilerInputSourceCode, CompilerInputSourceFile, CompilerOutput, CompilerOutputError, CompilerOutputContractAbi, CompilerOutputEvmBytecode, compileStandardWrapper } from 'solc';
```
MicahZoltu: You only need to import classes/interfaces that you will be using directly.  In this case, I think only `CompilerOutput` is referenced so that is the only thing you need to import.

adrake33: Good to know.  I only included them because we were already including CompilerInputSourceCode and CompilerInputSourceFile, which aren't directly referenced anywhere.  I'll remove all the unused ones.

MicahZoltu:
I didn't validate it against the docs (I trust you on that), but in general I'm quite happy with the additional type definitions.

micah:
wrong micah...

adrake33:
Sorry about that.  I'm used to using that handle on Slack.

PR #303, opened 2017-09-20
initial commit with generic trusted transfer function for StandardTokens

nuevoalex:
It's pretty simple initially. I figure if we choose to migrate logging here at some point that can come separately.


nuevoalex:
My only though is having specific functions with more specific validation techniques but Im not sure it would actually reduce auditor burden. Something to think about going forward though. I'll note the additional risk there in an audit comment.

PR #304, opened 2017-09-21
Eth & Cash auto conversion

nuevoalex:
There were a few instances where a modifier wasn't appropriate since we actually disperse Cash to users based on another party initiating a transaction. Those cases are commented as such in the code and all tested with one exception which I'm intending to give as a first task to the new intern Lauren (@qqqube in Slack)


src/libraries/CashWrapper.sol
```diff
@@ -0,0 +1,45 @@
...
+ */
+contract CashWrapper is Controlled {
+    /**
+     * @dev Convert any ETH provided in the transaction into Cash before the function executes
+     */
+    modifier convertToCash() {
+        ethToCash();
+        _;
+        cashToETH();
+    }
```
MicahZoltu: I wonder if we could get away with just a single modifier that works like this one and delete `convertFromCash`?  Is there any situation where converting from ETH at the start would be a problem but we need to convert to ETH at the end?

src/libraries/CashWrapper.sol
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * @dev Withdraw any Cash held by the sender into ETH and transfer it to them after the function executes
+     */
+    modifier convertFromCash() {
+        _;
+        cashToETH();
+    }
+
+    function ethToCash() payable public returns (bool) {
```
MicahZoltu: Is there a compelling reason for this being public?  Auditing is simpler if it is private (though only marginally).

src/libraries/CashWrapper.sol
```diff
@@ -0,0 +1,45 @@
...
+
+import 'ROOT/trading/ICash.sol';
+import 'ROOT/Controlled.sol';
+import 'ROOT/Augur.sol';
+
+
+/**
+ * @title Provides modifiers which take care of Cash/ETH conversion
+ */
+contract CashWrapper is Controlled {
```
MicahZoltu: Naming: This contract doesn't actually wrap cash (which the name implies), it makes modifiers available that allow derived contracts to auto-wrap ETH with cash.  I wonder if we can come up with a name that makes that more clear?

Maybe `CashAutoConvert`?  `Apple is CashAutoConverter`?

src/trading/CancelOrder.sol
```diff
@@ -17,12 +18,12 @@ import 'ROOT/trading/IOrders.sol';
  * @title CancelOrder
  * @dev This allows you to cancel orders on the book.
  */
-contract CancelOrder is Controlled, ReentrancyGuard, ICancelOrder {
+contract CancelOrder is Controlled, CashWrapper, ReentrancyGuard, ICancelOrder {
```
MicahZoltu: `CashWrapper is Controlled`, which makes the `Controlled` reference here redundant.  I believe diamond inheritance is supported, but there is something that needs to be considered about the order you inherit from.  It may be easier to just remove `Controlled` here.

src/libraries/CashWrapper.sol
nuevoalex: I'd like that better. The Cash -> ETH modifier felt a little weird to me anyway.

src/libraries/CashWrapper.sol
nuevoalex: Originally I was going to test these outside of modifier use but it's not worth the tradeoff so I'll make em private

src/libraries/CashWrapper.sol
nuevoalex: `CashAutoConverter` sounds good to me

PR #305, opened 2017-09-22
More fixmes & todos

nuevoalex:
Seemed like a good place to do a PR

 - Migration of market when no reports occur during limited or all reporting
 - Clarify FIXME comment in TradingEscapeHatch
 - Split out functions in dispute bond tests

### DIFF
--- a/src/reporting/Market.sol
+++ b/src/reporting/Market.sol
@@ -188,10 +188,14 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
     }
 
     function tryFinalize() public returns (bool) {
-        require(tentativeWinningPayoutDistributionHash != bytes32(0));
-
         if (getReportingState() != ReportingState.AWAITING_FINALIZATION) {
             return false;
+        }
+
+        // If we're waiting for finalization and no reports have been recieved we move to the next reporting window
+        if (!hasReceivedAReport()) {
+            IReportingWindow _newReportingWindow = getBranch().getNextReportingWindow();
+            return migrateReportingWindow(_newReportingWindow);
         }
 
         if (getBranch().getForkingMarket() == this) {
@@ -209,7 +213,20 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
         // FIXME: if finalPayoutDistributionHash == getIdentityDistributionId(), transfer validity bond to reportingWindow (reporter fee pot)
         // FIXME: if automated report is wrong, transfer automated report bond to reportingWindow
         // FIXME: if automated report is right, transfer automated report bond to market creator
-        // FIXME: handle markets that get 0 reports during their scheduled reporting window. We should move to the next reporting window in this case.
+    }
+
+    // This function should not use the ReportingState since we may be in AWAITING_FINALIZATION which would make the method of determining reporting ambiguous 
+    function hasReceivedAReport() private constant returns (bool) {
+        if (getBranch().getForkingMarket() == this) {
+            return true;
+        }
+        if (address(limitedReportersDisputeBondToken) != NULL_ADDRESS) {
+            return reportingWindow.hasReceivedAllReport(this);
+        }
+        if (automatedReportReceivedTime == 0 || address(automatedReporterDisputeBondToken) != NULL_ADDRESS) {
+            return reportingWindow.getNumberOfReportsByMarket(this) > 0;
+        }
+        return true;
     }
 
     function migrateThroughAllForks() public returns (bool) {
@@ -444,24 +461,29 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
             return ReportingState.FORKING;
         }
 
+        if (block.timestamp > reportingWindow.getEndTime()) {
+            return ReportingState.AWAITING_FINALIZATION;
+        }
+
         // If a limited dispute bond has been posted we are in some phase of all reporting depending on time
         if (_limitedReportDisputed) {
-            if (block.timestamp > reportingWindow.getEndTime()) {
-                return ReportingState.AWAITING_FINALIZATION;
-            }
             if (reportingWindow.isDisputeActive()) {
-                return ReportingState.ALL_DISPUTE;
+                if (reportingWindow.hasReceivedAllReport(this)) {
+                    return ReportingState.ALL_DISPUTE;
+                } else {
+                    return ReportingState.AWAITING_FINALIZATION;
+                }
             }
             return ReportingState.ALL_REPORTING;
         }
 
         // Either no automated report was made or the automated report was disputed so we are in some phase of limited reporting
-        if (block.timestamp > reportingWindow.getEndTime()) {
-            return ReportingState.AWAITING_FINALIZATION;
-        }
-
         if (reportingWindow.isDisputeActive()) {
-            return ReportingState.LIMITED_DISPUTE;
+            if (reportingWindow.getNumberOfReportsByMarket(this) > 0) {
+                return ReportingState.LIMITED_DISPUTE;
+            } else {
+                return ReportingState.AWAITING_FINALIZATION;
+            }
         }
 
         return ReportingState.LIMITED_REPORTING;
--- a/src/reporting/Market.sol
+++ b/src/reporting/Market.sol
@@ -188,14 +188,20 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
     }
 
     function tryFinalize() public returns (bool) {
-        if (getReportingState() != ReportingState.AWAITING_FINALIZATION) {
+        ReportingState _state = getReportingState();
+        if (_state < ReportingState.AWAITING_FINALIZATION) {
+            return false;
+        }
+
+        if (_state == ReportingState.FINALIZED) {
             return false;
         }
 
         // If we're waiting for finalization and no reports have been recieved we move to the next reporting window
-        if (!hasReceivedAReport()) {
+        if (_state == ReportingState.AWAITING_FINALIZATION_MIGRATION) {
             IReportingWindow _newReportingWindow = getBranch().getNextReportingWindow();
-            return migrateReportingWindow(_newReportingWindow);
+            migrateReportingWindow(_newReportingWindow);
+            return false;
         }
 
         if (getBranch().getForkingMarket() == this) {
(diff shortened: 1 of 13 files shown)


src/reporting/Market.sol
```diff
@@ -188,12 +188,16 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
...
-
         if (getReportingState() != ReportingState.AWAITING_FINALIZATION) {
             return false;
         }
 
+        // If we're waiting for finalization and no reports have been recieved we move to the next reporting window
+        if (!hasReceivedAReport()) {
+            IReportingWindow _newReportingWindow = getBranch().getNextReportingWindow();
+            return migrateReportingWindow(_newReportingWindow);
+        }
```
MicahZoltu: This code _feels_ like it should go along with the market migration stuff?  Though, I believe currently the market migration stuff is all related to forking, so perhaps not.

MicahZoltu: I _believe_ that returning false from this method is meant to indicate that the market was not finalized and returning true means the market was finalized.  Based on that, I think this if block should _always_ return false (regardless of whether the market migrated successfully or not).

src/reporting/Market.sol
```diff
@@ -209,7 +213,20 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
...
+        if (getBranch().getForkingMarket() == this) {
+            return true;
+        }
+        if (address(limitedReportersDisputeBondToken) != NULL_ADDRESS) {
+            return reportingWindow.hasReceivedAllReport(this);
+        }
+        if (automatedReportReceivedTime == 0 || address(automatedReporterDisputeBondToken) != NULL_ADDRESS) {
+            return reportingWindow.getNumberOfReportsByMarket(this) > 0;
+        }
+        return true;
```
MicahZoltu: Currently, this will return true if the market hasn't ended yet or is still in designated reporting phase.  I recommend ensuring this method returns the correct value in _all_ situations, or throws in situations it can't handle, or the method name is changed to make it very clear when it is appropriate to call this method.  I worry that even if its current usage pattern is such that it is only called under fairly tight constraints, a future update may relax those and the code will _appear_ correct when it no longer is.

src/reporting/Market.sol
```diff
@@ -444,24 +461,29 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
             return ReportingState.FORKING;
         }
 
+        if (block.timestamp > reportingWindow.getEndTime()) {
+            return ReportingState.AWAITING_FINALIZATION;
```
MicahZoltu: I wonder if we should rename this state since now it means (if I am following the code right) `EITHER_AWAITING_FINALIZATION_OR_MIGRATION_TO_NEW_WINDOW` (note: I don't actually recommend that as an enum name).

MicahZoltu: Alternatively, can we differentiate between the two easily and have two different states?

src/reporting/Market.sol
```diff
@@ -444,24 +461,29 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
...
         if (_limitedReportDisputed) {
-            if (block.timestamp > reportingWindow.getEndTime()) {
-                return ReportingState.AWAITING_FINALIZATION;
-            }
             if (reportingWindow.isDisputeActive()) {
-                return ReportingState.ALL_DISPUTE;
+                if (reportingWindow.hasReceivedAllReport(this)) {
+                    return ReportingState.ALL_DISPUTE;
+                } else {
+                    return ReportingState.AWAITING_FINALIZATION;
```
MicahZoltu: Shouldn't this be, `ReportingState.AWAITING_ALL_REPORTING` (or similar)?  Perhaps we are overloading `AWAITING_FINALIZATION` more than I realized and it really is just a generic `PENDING` state.

src/reporting/Market.sol
nuevoalex: Will do. And yea the current migration stuff is pretty fork specific.

src/reporting/Market.sol
nuevoalex: If we add another state as suggested below I think we can actually remove this function entirely.

src/reporting/Market.sol
nuevoalex: In this case we're in the DISPUTE phase but since no reporting was received we are awaiting finalization so that we will migrate to a new window

src/reporting/Market.sol
nuevoalex: I would really like to have a distinct phase for this case (No reporters reported on the market during reporting). It does mean that there will be two phases which indicate the market needs to have `tryFinalize` called, but that's probably fine as long as the name reflects that.

Maybe something like `AWAITING_FINALIZATION_MIGRATION`?

src/reporting/ReportingWindow.sol
```diff
@@ -224,6 +225,10 @@ contract ReportingWindow is DelegationTarget, Typed, Initializable, IReportingWi
         return numberOfReportsByMarket[_market];
     }
 
+    function hasReceivedAllReport(IMarket _market) public afterInitialized constant returns (bool) {
+        return receivedAllReport[_market];
```
MicahZoltu: Hmm, I think I misunderstood your question in #dev earlier but get it now.  Since any market in all reporting phase necessarily has reports (since it received some in limited reporting phase) we need to decide if we will blindly accept the limited report outcome should no one report during all reporting.  I think this situation further cements my desire to count the dispute bond as negative reports.

I suppose even if we did have the dispute bond be negative, we can't change the tentative outcome to the "second best" because we don't know who the second best was...

We should probably have a more involved discussion on how we want to move forward with this code (dealing with dispute bonds as negative stake and no reporters during all reporting).

nuevoalex: I suspected this would provoke more discussion :]

I'll address the easy bits here and modify it to use a distinct phase for this case in the mean time still.

src/reporting/Market.sol
MicahZoltu: Isn't this code path in the "all we know is that a limited dispute exists, but we still may be in limited dispute phase"?

nuevoalex: We know for certain at this point we're in the `ALL DISPUTE` phase and that we've received no reporting during the `ALL REPORT` phase.

nuevoalex:
Just noting here we also decided to remove the code that makes designated/automated reporter markets not get migrated & to remove Topics related code.

nuevoalex:
Other issues/changes to pull in:

- Rename `marketDenominator` to `numTicks` and have the market creation extensions take just that argument
- Keep the top 2 payout hashes stored in the market to accommodate negative stake from dispute bonds
- Turn designated/automated reporting into normal reporting rather than special such that they stake the same way as limited/all reporting but that phase is limited to the designated reporter only.
- No longer require a report during limited and all reporting. Simply require a tentativePayoutHash exist, and if not migrate to the next reporting window.

src/reporting/Market.sol
```diff
@@ -455,13 +456,17 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
...
-        if (block.timestamp > reportingWindow.getEndTime()) {
+        if (_reportingWindowOver) {
             return ReportingState.AWAITING_FINALIZATION;
         }
 
+        // Either no automated report was made or the automated report was disputed so we are in some phase of limited reporting
         if (reportingWindow.isDisputeActive()) {
-            return ReportingState.LIMITED_DISPUTE;
+            if (tentativeWinningPayoutDistributionHash == bytes32(0)) {
+                return ReportingState.AWAITING_FINALIZATION;
```
MicahZoltu: I believe we discussed this previously in this PR, but I don't remember the outcome.  Shouldn't this be `AWAITING_MIGRATION` or something as this market needs to move to a future reporting window?

nuevoalex: Since we no longer need to check if any reports were made I was able to remove the explicit status for finalization migration in favor of just checking the tentative payout hash. Recall before we had a new status for this `AWAITING_FINALIZATION_MIGRATION`

MicahZoltu: But this market isn't actually awaiting finalization right?  It is awaiting being moved to a new reporting window?

nuevoalex: Technically it will be migrated but the awaiting migration status is actually pretty specific to forking. We could rename that status to something more specific `AWAITING_FORK_MIGRATION` and then add back `AWAITING_FINALIZATION_MIGRATION`

MicahZoltu: It isn't awaiting finalization at all though.  It is awaiting migration to a new reporting window.  My contention is primarily around the `FINALIZATION` bit, since that implies the market is ready to finalize (which it most definitely isn't).

`AWAITING_NO_REPORT_MIGRATION` perhaps?  This would align with a change to `AWAITING_FORK_MIGRATION` where the name describes the type of migration we are waiting on.

Alternatively, we could include the migration destination instead:
`AWAITING_UNIVERSE_MIGRATION` vs `AWAITING_REPORTING_WINDOW_MIGRATION`.

nuevoalex: I like the `AWAITING_NO_REPORT_MIGRATION` and `AWAITING_FORK_MIGRATION`.

Do we want to make a new function to trigger the `NO REPORT` migration or continue to do it when `tryFinalize` is called?

MicahZoltu: Can we make the move happen automatically when someone tries to report on the market?

nuevoalex: Time-wise it will always be after reporting for the market is finished so that seems like an odd way to trigger it to me unless I'm missing something.

MicahZoltu: 1. Market is in window 5, receives no reports.
2. Time passes and window 6 begins.
3. UI presents market to user as part of window 6.
4. User attempts to report on market (whose state is still AWAITING_NO_REPORT_MIGRATION`).
5. As part of the report transaction, the market is moved to window 6 (which will change its state to `LIMITED_REPORTING`).

nuevoalex: I get the goal though, to have there be no interruption and seamlessly be able to be reported on. We could do this as long as we think it makes sense to consumers (API mostly)

MicahZoltu: I think this is a solvable problem at the UI level, and from the contracts this is the best option as it doesn't require a user signing a one-off transaction just to trigger a move so they can then report on it.  Things do get weird if a fork starts when a market is in this state, so we'll need to make sure to handle that.

PR #306, opened 2017-09-23
Deployment

adrake33:
The adds a library for contract deployment and adds code to allow the contract compiler to work with the contract deployer.  Also created a simple unit test.


source/deployment/deployContracts.ts
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env node
```
MicahZoltu: TS files can't be directly executed by `node` so this line isn't correct and should be left out.

source/deployment/deployContracts.ts
```diff
@@ -0,0 +1,61 @@
...
+
+const CONTRACT_INPUT_DIR_PATH = path.join(__dirname, "../../source/contracts");
+const CONTRACT_OUTPUT_DIR_PATH = path.join(__dirname, "../contracts");
+const COMPILED_CONTRACT_OUTPUT_FILE_NAME = "augurCore";
+const GAS_AMOUNT = 3000000;
+const DEFAULT_ETHEREUM_PORT = 8545;
+
+// Set these variables to use node settings other than TestRPC on port 8545
+const ETHEREUM_HOST = "";
+const ETHEREUM_PORT = "";
```
MicahZoltu: No need to have file variables for these, you can just use `process.env.ETHEREUM_HOST` down on line 30/31 instead (and similar for port).

source/libraries/CompileSolidity.ts
```diff
@@ -11,58 +11,66 @@ interface CompileContractsOutput {
...
-            const compilerOutputJson: string = compileStandardWrapper(JSON.stringify(compilerInputJson), this.readCallback);
-            const compilerOutput: CompilerOutput = JSON.parse(compilerOutputJson);
-            if (compilerOutput.errors) {
-                let errors = "";
-                for (let error of compilerOutput.errors) {
-                    errors += error.formattedMessage + "\n";
-                }
-                throw new Error("The following errors/warnings were returned by solc:\n\n" + errors);
-            }
+            const result = fs.readFileSync(path, 'utf8');
```
MicahZoltu: We should have read in all files by the time this is called so we should be able to just reference the previously-read file here.  Alternatively, maybe we should just remove the `readCallback` and let things crash if some file is attempted to be read that doesn't exist?

source/deployment/deployContracts.ts
adrake33: Ok, I wasn't sure if we wanted the JS file generated from the TS file to have the shebang line at the top.

source/libraries/CompileSolidity.ts
```diff
@@ -11,58 +11,66 @@ interface CompileContractsOutput {
...
+        }
+    }
 
-            // Create output directory (if it doesn't exist)
-            mkdirp(this.contractOutputDirectoryPath, this.mkdirpCallback);
+    public async compileContracts(): Promise<CompileContractsOutput> {
+        // Check if all contracts are cached (and thus do not need to be compiled)
+        const stats = fs.statSync(this.contractOutputDirectoryPath + this.contractOutputFileName);
+        const lastCompiledTimestamp = stats.mtime;
+        const uncachedFiles = await recursiveReadDir(this.contractInputDirectoryPath, [function(file: string, stats: fs.Stats): boolean {return stats.isDirectory() || (stats.isFile() && path.extname(file) != ".sol") || (stats.isFile() && path.extname(file) == ".sol" && stats.mtime < lastCompiledTimestamp);}]);
```
MicahZoltu: Consider extracting the ignore function out into a separate function to improve readability.

source/libraries/CompileSolidity.ts
```diff
@@ -11,58 +11,66 @@ interface CompileContractsOutput {
...
-        } catch (error) {
-            throw error;
+            throw new Error("The following errors/warnings were returned by solc:\n\n" + errors);
         }
-    }
 
-    private ignoreFile(file: string, stats: fs.Stats): boolean {
-        return stats.isFile() && path.extname(file) != ".sol";
+        // Create output directory (if it doesn't exist)
+        mkdirp(this.contractOutputDirectoryPath, this.mkdirpCallback);
```
MicahZoltu: Is there an async version of this function?

source/libraries/CompileSolidity.ts
```diff
@@ -11,58 +11,66 @@ interface CompileContractsOutput {
...
+
+        // Output contract data to single file
+        const contractOutputFilePath = this.contractOutputDirectoryPath + this.contractOutputFileName;
+        let wstream: any = fs.createWriteStream(contractOutputFilePath);
```
MicahZoltu: Is there an async version of this function we could use instead?  Also, why `:any` here?  Doesn't `createWriteStream` return a strongly typed object?

source/deployment/deployContracts.ts
adrake33: So that logic should look like the following?

```
    const httpProviderport = (typeof process.env.ETHEREUM_PORT == 'undefined') ? DEFAULT_ETHEREUM_PORT : process.env.ETHEREUM_PORT;
    if (typeof process.env.ETHEREUM_HOST == 'undefined') {
        const rpcClient = new RpcClient();
        await rpcClient.listen(httpProviderport);
    }
```

And we would set these values in a .env file?  (Sorry, I haven't worked with Node much before.  I did a search on how to set Node environment variables, and it sounds like this is how we would set them?)

source/deployment/deployContracts.ts
MicahZoltu: Oh, will this land in the generated JS files?  If so I guess this is fine.  I tend to not execute scripts directly (always `node ...`) but I know a lot of other people do and if this makes it to the JS file then that seems valuable to those people.

source/deployment/deployContracts.ts
MicahZoltu: Yeah, I believe that is correct.

You would just export them before running the script:
```
export ETHEREUM_HOST = 'localhost'
export ETHEREUM_PORT = '8545'
node output/deployment/deployContracts.js
```

source/libraries/RpcClient.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export class RpcClient {
+    public async listen(port): Promise<boolean> {
+        try {
+            const server = TestRpc.server();
+            server.listen(port);
+            return true;
+        } catch (error) {
+            throw error;
+        }
```
MicahZoltu: No need for this try/catch as it doesn't do anything.

typings/contract-deployment.d.ts
```diff
@@ -0,0 +1,22 @@
+declare module 'contract-deployment' {
+    interface ContractBlockchainData {
+        abi: any;
+        query: any;
+        address: string;
+        bytecode: string;
+        defaultTxObject: any;
+        filters: any;
+        // TODO: Figure out how to specify in TypeScript that ContractBlockchainData can have any number of functions
```
MicahZoltu: Ping me in dev with the details/example of what else goes on this interface and I can help figure out how to structure it.

source/libraries/ContractDeployer.ts
```diff
@@ -0,0 +1,33 @@
...
+        return contractBuilder.at(receipt.contractAddress);
+    }
+
+    public async deployContracts(ethjs: Eth, contractJson: string, from: string, gas: number): Promise<ContractBlockchainData[]> {
+        const contractFiles = JSON.parse(contractJson);
+        let deployedContracts = [];
+        for (let contractFileName in contractFiles) {
+            for (let contractName in contractFiles[contractFileName]) {
+                // Filter out interface contracts, as they do not need to be deployed
+                if (contractFiles[contractFileName][contractName].evm.bytecode.object != '') {
```
MicahZoltu: Recommend inverting this if statement to minimize nesting:
```
if (!contractFiles[contractFileName][contractName].evm.bytecode.object) {
    continue;
}
```

source/deployment/deployContracts.ts
adrake33: Yeah, I checked the generated JS file, and it does show up there.  I'll go ahead and leave it in.

source/libraries/CompileSolidity.ts
adrake33: Yeah, it looks like there's `async-mkdirp`.  I'll use that instead.

source/libraries/RpcClient.ts
adrake33: Sorry, I thought I'd removed this earlier.  Will take it out.

source/libraries/CompileSolidity.ts
```diff
@@ -72,26 +84,26 @@ class SolidityContractCompiler {
     }
 
     private async generateCompilerInput(): Promise<CompilerInput> {
+        const filePaths: any = await recursiveReadDir(this.contractInputDirectoryPath, [this.ignoreFile]);
```
MicahZoltu: You can delete this `: any`.  `recursiveReadDir()` correctly returns `Promise<string[]>` and when I tested everything works correctly with that.

source/deployment/deployContracts.ts
```diff
@@ -0,0 +1,57 @@
...
+import { ContractBlockchainData } from 'contract-deployment';
+import { SolidityContractCompiler } from "../libraries/CompileSolidity";
+import { ContractDeployer } from "../libraries/ContractDeployer";
+import { RpcClient } from "../libraries/RpcClient";
+
+const CONTRACT_INPUT_DIR_PATH = path.join(__dirname, "../../source/contracts");
+const CONTRACT_OUTPUT_DIR_PATH = path.join(__dirname, "../../output/contracts");
+const COMPILED_CONTRACT_OUTPUT_FILE_NAME = "augurCore";
+const GAS_AMOUNT = 3000000;
+const DEFAULT_ETHEREUM_PORT = 8545;
```
MicahZoltu: This should be some other port since 8545 is _likely_ to be in-use by a developer.  Consider using something like `https://www.npmjs.com/package/get-port` which will get a random unused port.  It supports promises (async/await) and you can provide it with a preferred port that will be used if available.  Then if someone already has something listening on 8545 a different port will be used.

adrake33: Yeah, good point.  I'll try using `get-port` instead.

source/deployment/deployContracts.ts
```diff
@@ -0,0 +1,81 @@
...
+    if (typeof process.env.ETHEREUM_HOST == "undefined") {
+        const privateKey = "Augur";
+        const hexlifiedPrivateKey = await padAndHexlify(privateKey, 64);
+
+        // This sets the account balance & private key for test accounts
+        const accountArray = [{balance: DEFAULT_TEST_ACCOUNT_BALANCE, secretKey: hexlifiedPrivateKey}];
+        const options = {gasLimit: GAS_BLOCK_AMOUNT, accounts: accountArray};
+
+        const rpcClient = new RpcClient();
+        await rpcClient.listen(httpProviderport, options);
```
MicahZoltu: The first parameter needs to be a number, but `httpProviderport` is a `number|undefined|string`.  I believe the issue is that `process.env.ETHEREUM_PORT` returns a `string|undefined` so you'll need to parse that into a number with `parseInt(...)` or similar.

source/deployment/deployContracts.ts
```diff
@@ -0,0 +1,81 @@
...
+    return "0x" + binascii.hexlify(paddedStringToEncode);
+}
+
+export async function compileAndDeployContracts(): Promise<ContractBlockchainData[]> {
+    // Compile contracts to a single output file
+    const solidityContractCompiler = new SolidityContractCompiler(CONTRACT_INPUT_DIR_PATH, CONTRACT_OUTPUT_DIR_PATH, COMPILED_CONTRACT_OUTPUT_FILE_NAME);
+    const compilerResult = await solidityContractCompiler.compileContracts();
+
+    // Initialize Ethereum node details.  (If no host is specified, TestRPC will be used.)
+    const httpProviderport = (typeof process.env.ETHEREUM_PORT == "undefined") ? DEFAULT_ETHEREUM_PORT : process.env.ETHEREUM_PORT;
```
MicahZoltu: Nitpick: `P` in `port` should be capitalized in variable name.

source/libraries/ContractDeployer.ts
```diff
@@ -0,0 +1,156 @@
...
+        await this.approveCentralAuthority();
+
+        return true;
+    }
+
+    public async uploadAndAddToController(relativeFilePath: string, lookupKey: string = "", signatureKey: string = "", constructorArgs: any = []): Promise<ContractBlockchainData> {
+        lookupKey = (lookupKey == "") ? path.basename(relativeFilePath).split(".")[0] : lookupKey;
+        const contract = await this.upload(relativeFilePath, lookupKey, signatureKey, constructorArgs);
+        if (typeof contract == "undefined") {
+            return {};
```
MicahZoltu: This isn't a valid return type for this function.  Recommend `throw` instead if this is a failure scenario (it looks like it is).  Same comment below.

adrake33: Ok, sounds good.  I'll change this to a throw.

source/libraries/ContractDeployer.ts
```diff
@@ -0,0 +1,156 @@
...
+        await this.uploadAllContracts();
+        await this.whitelistTradingContracts();
+        await this.initializeAllContracts();
+        await this.approveCentralAuthority();
+
+        return true;
+    }
+
+    public async uploadAndAddToController(relativeFilePath: string, lookupKey: string = "", signatureKey: string = "", constructorArgs: any = []): Promise<ContractBlockchainData> {
+        lookupKey = (lookupKey == "") ? path.basename(relativeFilePath).split(".")[0] : lookupKey;
```
MicahZoltu: In JavaScript (and by extension TypeScript) always use `===` for equality comparison and `!==` for inequality.  `==` in JavaScript is fraught with peril.

MicahZoltu: (same comment all over)

package.json
```diff
@@ -1,11 +1,25 @@
...
+        "ethereumjs-testrpc": "^4.1.3",
+        "ethjs-contract": "^0.1.9",
+        "ethjs-provider-http": "^0.1.6",
+        "ethjs-query": "^0.2.9",
         "fs-readfile-promise": "^3.0.0",
         "mkdirp": "^0.5.1",
+        "mocha": "^3.5.3",
         "path": "^0.12.7",
         "recursive-readdir": "^2.2.1",
         "solc": "^0.4.16",
```
MicahZoltu: The contracts don't currently compile with solc `0.4.16`.  We should change this to `"solc": "0.4.15"` until we can take the time to fix all of the warnings.

source/libraries/ContractDeployer.ts
adrake33: Yep, better safe than sorry.  Will change this.

source/libraries/CompileSolidity.ts
adrake33: Yeah, I suppose we don't really need `readCallback` here.  I'll just remove it.

source/libraries/CompileSolidity.ts
adrake33: Yes, I'll change this to `fs.writeFile()` and remove the `:any`.

typings/contract-deployment.d.ts
adrake33: Since I have a TODO note here, let's revisit this later.  I have all of the other recommendations for this PR finished now, and I feel like this PR is about at a point where it's useable and can probably be merged.

adrake33:
@MicahZoltu Please take one more look over everything and let me know if you have any other comments/suggestions.  If not, I'll go ahead and merge what I have so far.

PR #307, opened 2017-09-25
marketDenominator -> numTicks

nuevoalex:
rename `marketDenominator` to `numTicks`

### DIFF
--- a/src/extensions/MarketCreation.sol
+++ b/src/extensions/MarketCreation.sol
@@ -7,9 +7,9 @@ import 'ROOT/reporting/IReportingWindow.sol';
 
 
 contract MarketCreation {
-    function createScalarMarket(IBranch _branch, uint256 _endTime, uint256 _feePerEthInWei, ICash _denominationToken, uint256 _priceRange, uint256 _numTicks, address _automatedReporterAddress, bytes32 _topic) payable public returns (IMarket) {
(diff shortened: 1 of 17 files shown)


src/extensions/MarketCreation.sol
```diff
@@ -7,9 +7,9 @@ import 'ROOT/reporting/IReportingWindow.sol';
 
 
 contract MarketCreation {
-    function createScalarMarket(IBranch _branch, uint256 _endTime, uint256 _feePerEthInWei, ICash _denominationToken, uint256 _priceRange, uint256 _numTicks, address _automatedReporterAddress, bytes32 _topic) payable public returns (IMarket) {
+    function createScalarMarket(IBranch _branch, uint256 _endTime, uint256 _feePerEthInWei, ICash _denominationToken, uint256 _numTicks, address _automatedReporterAddress, bytes32 _topic) payable public returns (IMarket) {
         IReportingWindow _reportingWindow = _branch.getReportingWindowByMarketEndTime(_endTime, (_automatedReporterAddress != 0));
-        return _reportingWindow.createNewMarket.value(msg.value)(_endTime, 2, _priceRange * _numTicks, _feePerEthInWei, _denominationToken, msg.sender, _automatedReporterAddress, _topic);
+        return _reportingWindow.createNewMarket.value(msg.value)(_endTime, 2, _numTicks, _feePerEthInWei, _denominationToken, msg.sender, _automatedReporterAddress, _topic);
```
MicahZoltu: Consider: Now may be the time to collapse these two functions into one since they now only have a single difference which is `numOutcomes`.

nuevoalex: Good point. I'll do so

PR #308, opened 2017-09-25
remove post-automated requirement on fork migration

nuevoalex:
As it turns out this didn't simplify things much except possibly in terms of groking how the system at large functions. I'm still in favor of this change for that reason.

### DIFF
--- a/src/reporting/Market.sol
+++ b/src/reporting/Market.sol
@@ -406,20 +406,20 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
     }
 
     function getReportingState() public constant returns (ReportingState) {
-        // Before trading in the market is finished
-        if (block.timestamp < endTime) {
-            return ReportingState.PRE_REPORTING;
-        }
-
         // This market has been finalized
         if (finalPayoutDistributionHash != bytes32(0)) {
             return ReportingState.FINALIZED;
         }
 
-        // If there is an active fork and we're past PRE-REPORTING we need to migrate
+        // If there is an active fork we need to migrate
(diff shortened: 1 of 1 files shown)


src/reporting/Market.sol
```diff
@@ -416,6 +416,12 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
             return ReportingState.FINALIZED;
         }
 
+        // If there is an active fork and we're past PRE-REPORTING we need to migrate
```
MicahZoltu: Even `PRE_REPORTING` markets need to be migrated.  It doesn't appear that we currently use `PRE_REPORTING` (though I think it is a valuable state), so I think for clarity we should change this comment to just be,
> If there is an active fork and we aren't finalized, we need to migrate.

Whether we label the market as `PRE_REPORTING` or `AWAITING_MIGRATION` doesn't really matter to me, but technically the market is `PRE_REPORTING & AWAITING_MIGRATION`.

nuevoalex: We'll need to move this up actually if that's the case. See above: 

> if (block.timestamp < endTime) {
>     return ReportingState.PRE_REPORTING;
>  }

MicahZoltu: We can return _either_ `PRE_REPORTING` or `AWAITING_MIGRATION` for markets in this state and the system will be correct.  The `PRE_REPORTING` check is cheaper, so my preference is to leave it up there and this one down here.  My concern is just that the comment tells the reader that being past `PRE_REPORTING` is a requirement.

PR #310, opened 2017-09-26
Added Rep to Suicide to close #241

### LINKED ISSUE #241: Make augur contract self destructs work w/ tokens too
﻿# Issue Template
### Brief Description

See https://github.com/OpenZeppelin/zeppelin-solidity/blob/master/contracts/lifecycle/TokenDestructible.sol

joeykrug:


source/contracts/Controlled.sol
```diff
@@ -37,6 +37,12 @@ contract Controlled is IControlled {
     }
 
     function suicideFunds(address _target) public onlyControllerCaller returns(bool) {
+        // Transfer REP tokens to target
+        ERC20Basic repToken = ERC20Basic(controller.lookup("Reputation"));
```
MicahZoltu: There are many `ReputationToken` contracts and they aren't registered with controller (because we would have no way to sanely name them).  To get the the reputation token you need a branch to start with and then you can call `getReputationToken()` on it to get the address of the reputation token for that branch.  The only way to start from a global state and get to a branch is by starting with the genesis branch and walking the forks.  This is (technically) an unbounded operation so should not be done here.

I think if you want to be able to suicide owned REP then this function will need a new parameter for `Branch`.

source/contracts/Controlled.sol
```diff
@@ -36,7 +37,15 @@ contract Controlled is IControlled {
         return true;
     }
 
-    function suicideFunds(address _target) public onlyControllerCaller returns(bool) {
+    function suicideFunds(address _target, IUniverse _universe) public onlyControllerCaller returns(bool) {
+        // Transfer REP tokens to target
+        if(_universe) {
```
MicahZoltu: This line will fail to compile, you need to compare `_universe` against `address(0)` to have it be a valid boolean expression.

source/contracts/Controlled.sol
```diff
@@ -36,7 +37,15 @@ contract Controlled is IControlled {
         return true;
     }
 
-    function suicideFunds(address _target) public onlyControllerCaller returns(bool) {
+    function suicideFunds(address _target, IUniverse _universe) public onlyControllerCaller returns(bool) {
```
MicahZoltu: I wonder if we would be better off having this take in an array of ERC20 tokens instead?  That would allow the caller to specify any tokens that the contract somehow got a hold of, including REP from multiple universes, share tokens, Cash, or any other random token.

source/contracts/Controller.sol
```diff
@@ -87,8 +88,8 @@ contract Controller is IController {
      * Contract Administration [dev mode can use it]
      */
 
-    function suicide(IControlled _target, address _destination) public devModeOwnerOnly returns(bool) {
-        _target.suicideFunds(_destination);
+    function suicide(IControlled _target, address _destination, IUniverse _universe) public devModeOwnerOnly returns(bool) {
```
MicahZoltu: Needs rebase on/merge of develop, `suicide` was renamed to address Solidity compiler warning about function name conflicting with name of reserved word.

source/contracts/Controlled.sol
joeykrug: Yeah good call

source/contracts/Controller.sol
joeykrug: What was it renamed to? I don't see it renamed anywhere

MicahZoltu: @adrake33 You merged your change that renamed `suicide` right?  Or is that still pending merge today?

source/contracts/Controlled.sol
```diff
@@ -36,7 +37,16 @@ contract Controlled is IControlled {
         return true;
     }
 
-    function suicideFunds(address _target) public onlyControllerCaller returns(bool) {
+    function suicideFunds(address _target, ERC20Basic[] _tokens) public onlyControllerCaller returns(bool) {
+        // Transfer tokens to target
+        // TODO Uncomment when Market.sol size is fixed
```
MicahZoltu: Add a `:` after `TODO` for `// TODO: Uncomment...`.  This will make it so it shows up when we are searching for TODOs as that is how the rest are formatted.

MicahZoltu: (assuming your plan is to merge with the TODO included)

tests/test_controller.py
```diff
@@ -33,10 +33,10 @@ def test_registry(controller, decentralizedController):
     with raises(TransactionFailed): controller.assertOnlySpecifiedCaller(tester.a2, key2, sender = tester.k2)
 
 def test_suicide(controller, decentralizedController, controllerUser):
-    with raises(TransactionFailed): controller.suicide(controllerUser.address, tester.a0, sender = tester.k2)
+    with raises(TransactionFailed): controller.suicide(controllerUser.address, tester.a0, [], sender = tester.k2)
```
MicahZoltu: Recommend adding a test that includes a token (the test legacy rep token is probably easiest as you can just give yourself some REP, transfer it to the controlled contract, then suicide with that address in the array.

joeykrug:
Would recommended way to add a test for it be to use `contractsFixture.upload` and upload a variableSupplyToken, mint some supply, and then send it to the contract we're testing?

MicahZoltu:
I would use `source/contracts/LegacyRepContract.sol` (which I think is already uploaded in `contractsFixture` as that has a `.faucet` on it so you can give yourself some, transfer them to the contract under test, then suicide it with `LegacyRepContract`.  Alternatively, you could upload a new `VariableSupplyToken` but that seems like unnecessary work.

tests/test_controller.py
```diff
@@ -32,12 +32,16 @@ def test_registry(controller, decentralizedController):
...
+def test_suicideFunds(contractsFixture, controller, decentralizedController, controllerUser):
+    legacyRepContract = contractsFixture.contracts['LegacyRepContract']
+    legacyRepContract.faucet(0)
+    legacyRepContract.transfer(controllerUser.address, 40 * 10**18)
+    with raises(TransactionFailed): controller.suicideFunds(controllerUser.address, tester.a0, [legacyRepContract.address], sender = tester.k2)
     assert decentralizedController.owner() == bytesToHexString(tester.a0)
-    with raises(TransactionFailed): decentralizedController.suicideFunds(controllerUser.address, tester.a0, sender = tester.k0)
-    assert controller.suicideFunds(controllerUser.address, tester.a0, sender = tester.k0)
+    with raises(TransactionFailed): decentralizedController.suicideFunds(controllerUser.address, tester.a0, [legacyRepContract.address], sender = tester.k0)
+    assert controller.suicideFunds(controllerUser.address, tester.a0, [legacyRepContract.address], sender = tester.k0)
```
MicahZoltu: Recommend creating a new test like `test_suicide_tokens` that does the token stuff and asserts that the tokens were correctly suicided.  Such a test will fail now since a block of the code is commented out, but we can mark just that test as skipped and give the reason for the skip so that once we uncomment the code we just have to unmark it.

joeykrug: How do you mark a test as skipped

MicahZoltu: https://docs.pytest.org/en/latest/skipping.html#skipping-test-functions

```
from pytest import mark
@mark.skip(reason="This test currently fails for unknown reasons, need to investigate before release.")
def test_stuff():
    assert False
```

PR #311, opened 2017-09-26
src -> source/contracts

nuevoalex:


.travis/ci.sh
```diff
@@ -3,6 +3,6 @@
 set -e
 
 echo "Starting Solidity Linting..."
-docker container run --entrypoint=npx augur/core-test:latest solium --dir src --reporter=gcc
+docker container run --entrypoint=npx augur/core-test:latest solium --dir source --reporter=gcc
```
MicahZoltu: This should be `source/contracts`

PR #312, opened 2017-09-27
Remove ROOT remappings

adrake33:
Removed "ROOT" prefix from import lines in Solidity contracts.  Also updated conftest.py so that the Python tests will still work.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -102,7 +102,7 @@ class ContractsFixture:
                 }
             },
             'settings': {
-                'remappings': [ "=" + resolveRelativePath("../source/contracts") + "/" ],
+                'remappings': [ '=%s/' % resolveRelativePath("../source/contracts") ],
                 'outputSelection': {
                     '*': [ 'metadata', 'evm.bytecode', 'evm.sourceMap' ]
                 }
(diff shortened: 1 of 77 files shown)


tests/conftest.py
```diff
@@ -102,7 +102,7 @@ def compileSolidity(relativeFilePath):
                 }
             },
             'settings': {
-                'remappings': [ 'ROOT=%s' % resolveRelativePath("../source/contracts") ],
+                'remappings': [ '=%s/' % resolveRelativePath("../source/contracts") ],
```
MicahZoltu: We should be able to remove this line and edit line 100 above to change the key to be the relative file path (no prefix) just like you did in the new compile stuff.  Something like `stripPrefix(resolveRelativePath("../source/contracts/"), absoluteFilePath)` (made up some of those functions but that is the idea).

If that proves to be too difficult and this works, we can leave it like this for now and make that change later.

adrake33: Yeah, I figured this line should be taken out altogether, but removing it didn't work, and I wasn't sure exactly what other change I needed to make.  I'll update the path in the `sources` array like you mentioned.

adrake33: I tried:
```
resolveRelativePath(relativeFilePath).replace(resolveRelativePath("../source/contracts/") + "/", "")
```
which should remove the prefix and just leave you with "Controller.sol".  However, that didn't work.  Since it's getting late, I think I'll just merge this for now and fix it in the `deployment` branch.

PR #314, opened 2017-09-27
dispute bonds count as negative stake

nuevoalex:


source/contracts/reporting/Market.sol
```diff
@@ -171,20 +169,44 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
...
+        int256 _payoutStake = getPayoutDistributionHashStake(_payoutDistributionHash, _reportingToken);
+
+        if (_payoutStake > _tentativeWinningStake) {
+            secondPlaceTentativeWinningPayoutDistributionHash = tentativeWinningPayoutDistributionHash;
             tentativeWinningPayoutDistributionHash = _payoutDistributionHash;
         }
         return true;
     }
 
+    function getPayoutDistributionHashStake(bytes32 _payoutDistributionHash, IReportingToken _reportingToken) private returns (int256) {
```
MicahZoltu: I believe this should be `constant`, and there is probably value in having it be `public` as well.

source/contracts/reporting/Market.sol
```diff
@@ -171,20 +169,44 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
...
+        if (address(automatedReporterDisputeBondToken) != NULL_ADDRESS) {
+            if (automatedReporterDisputeBondToken.getDisputedPayoutDistributionHash() == _payoutDistributionHash) {
+                _payoutStake -= int256(Reporting.automatedReporterDisputeBondAmount());
+            }
+        }
+        if (address(limitedReportersDisputeBondToken) != NULL_ADDRESS) {
+            if (limitedReportersDisputeBondToken.getDisputedPayoutDistributionHash() == _payoutDistributionHash) {
+                _payoutStake -= int256(Reporting.limitedReportersDisputeBondAmount());
+            }
+        }
```
MicahZoltu: For completeness, it feels like we should include `allReportersDisputeBond` even though we don't use this function directly during a fork at the moment.  The UI however may want to present the various stakes during a fork and they would use this function and the all-dispute should be included in that case.

source/contracts/reporting/Market.sol
MicahZoltu: Consider getting `_payoutDistributionHash` from `_reportingToken` rather than passing it in.  Will increase gas costs slightly, but we can address that later if it turns out that this is a hot path that needs gas optimization.  I feel like having a single parameter will be more clear for reading/auditing.

source/contracts/reporting/Market.sol
```diff
@@ -171,20 +169,44 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
...
-            _tentativeWinningReportingToken = _reportingToken;
+            return true;
         }
-        if (_reportingToken.totalSupply() > _tentativeWinningReportingToken.totalSupply()) {
+
+        int256 _tentativeWinningStake = getPayoutDistributionHashStake(tentativeWinningPayoutDistributionHash, _tentativeWinningReportingToken);
+        int256 _payoutStake = getPayoutDistributionHashStake(_payoutDistributionHash, _reportingToken);
+
+        if (_payoutStake > _tentativeWinningStake) {
+            secondPlaceTentativeWinningPayoutDistributionHash = tentativeWinningPayoutDistributionHash;
```
MicahZoltu: I don't believe this logic is correct.  It is possible that a dispute bond placement may drop a winning outcome to 3rd place or worse, so we can't assume that just because `_payoutStake > _firstPlaceStake` that `_firstPlaceStake > _secondPlaceStake`.

source/contracts/reporting/Market.sol
```diff
@@ -136,27 +132,29 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
...
         require(getReportingState() == ReportingState.LIMITED_DISPUTE);
-        limitedReportersDisputeBondToken = DisputeBondTokenFactory(controller.lookup("DisputeBondTokenFactory")).createDisputeBondToken(controller, this, msg.sender, LIMITED_REPORTERS_DISPUTE_BOND_AMOUNT, tentativeWinningPayoutDistributionHash);
-        reportingWindow.getReputationToken().trustedTransfer(msg.sender, limitedReportersDisputeBondToken, LIMITED_REPORTERS_DISPUTE_BOND_AMOUNT);
+        limitedReportersDisputeBondToken = DisputeBondTokenFactory(controller.lookup("DisputeBondTokenFactory")).createDisputeBondToken(controller, this, msg.sender, Reporting.limitedReportersDisputeBondAmount(), tentativeWinningPayoutDistributionHash);
+        reportingWindow.getReputationToken().trustedTransfer(msg.sender, limitedReportersDisputeBondToken, Reporting.limitedReportersDisputeBondAmount());
         IReportingWindow _newReportingWindow = getBranch().getNextReportingWindow();
+        if (secondPlaceTentativeWinningPayoutDistributionHash != bytes32(0)) {
+            updateTentativeWinningPayoutDistributionHash(secondPlaceTentativeWinningPayoutDistributionHash);
+        }
```
MicahZoltu: Shouldn't this same block be added to the other two dispute functions?

nuevoalex: For `automated` there can't be a second place and for `all` the payoutHash is no longer relevant, but we could do this anyway in both with no negative side effects except gas cost.

MicahZoltu: Actually, I think no matter `updateTentativeWinningPayoutDistributionHash` should be called.  It should then contain all of the logic necessary to figure out the top 2 winners based on provided hash and existing first and second place (which can be fetched from storage in that function).  So I would get rid of this if block and just call `updateTentativeWinningPayoutDistributionHash(disputedHash)`.

source/contracts/reporting/Market.sol
nuevoalex: The only case where second place matters is during promotion during a `limited` dispute. When a dispute occurs if second place is promoted it is correct that it will be promoted to second place even if it is not technically, but this won't actually cause any errors. second place may simply be a bad name for the variable. It is simply the current best guess second place but should be guaranteed accurate in the one case we need to promote it.

MicahZoltu:
Do one or more of the tests you updated verify the situation:
1. Stake on outcome A with 100 REP
2. Stake on outcome B with 50 REP
3. Dispute (-10,000 REP)
4. Finalize market
5. Validate B won

We should also probably have a test for various iterations of that like A staked with 10,051, same with two disputes, etc.

nuevoalex:
I'll add some to the dispute bond tests that specifically check that sort of situation

source/contracts/reporting/Market.sol
MicahZoltu: I would prefer to just always call `updateTentativeWinningPayoutDistributionHash` for clarity, even if unnecessary.  We should be able to throw some early-exit checks at the top that make it so most of the function isn't run if the three values are already correctly "sorted".  I believe I see what you are saying with being able to avoid the code path in some situations, but I'm generally hesitant to rely on functions being called in certain contexts in order for them to behave correctly.

source/contracts/reporting/Market.sol
MicahZoltu: I wonder if we should rename the variable to `maybeSecondPlace` or something, to make it clear to the reader that they need to understand there are some caveats to the variable?

source/contracts/reporting/Market.sol
nuevoalex: Will do then! If we have to optimize in some way later I don't think it'll be hard to figure out.

source/contracts/reporting/Market.sol
nuevoalex: `bestGuessSecondPlace...` ?

MicahZoltu: `bestGuessSecondPlace` works for me.  Would be nice to have a comment above the variable declaration that gives a brief description of why it isn't always correct.

source/contracts/reporting/Market.sol
```diff
@@ -177,14 +176,53 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
...
+            IReportingToken _secondPlaceWinningReportingToken = reportingTokens[_secondPlacePayoutDistributionHash];
+            int256 _secondPlaceTentativeWinningStake = getPayoutDistributionHashStake(_secondPlaceWinningReportingToken);
+            if (_secondPlaceTentativeWinningStake > _tentativeWinningStake) {
+                bestGuessSecondPlaceTentativeWinningPayoutDistributionHash = tentativeWinningPayoutDistributionHash;
+                tentativeWinningPayoutDistributionHash = _secondPlacePayoutDistributionHash;
+            } else if (_secondPlacePayoutDistributionHash > bestGuessSecondPlaceTentativeWinningPayoutDistributionHash) {
+                bestGuessSecondPlaceTentativeWinningPayoutDistributionHash = _secondPlacePayoutDistributionHash;
+            }
+        }
+
```
MicahZoltu: This logic _feels_ more complicated than necessary.  Fundamentally, we are just trying to find the first and second place hashes given the stake behind each of the potentially 3 hashes.  If we tread non-existant hashes as 0 stake and null hashes as 0 stake, I believe the following if-else brick will come to the correct solution in a way that I _think_ is easier to read (open to feedback on the truth of that statement).  It is effectively a really naive sorting algorithm (since building a real sorting algorithm is a pain in Solidity when dealing with complex datatypes (hash + value).

```
if (stakeA > stakeB && stakeB > stakeC) {
  winningHash = stakeA;
  secondPlaceHash = stakeB;
} else if (stakeA > stakeC && stakeC > stakeB) {
  winningHash = stakeA;
  secondPlaceHash = stakeC;
} else if (stakeB > stakeA && stakeA > stakeC) {
  winningHash = stakeB;
  secondPlaceHash = stakeA;
} else if (stakeB > stakeC && stakeC > stakeA) {
  winningHash = stakeB;
  secondPlaceHash = stakeC;
} else if (stakeC > stakeA && stakeA > stakeB) {
  winningHash = stakeC;
  secondPlaceHash = stakeA;
} else if (stakeC > stakeB && stakeB > stakeA) {
  winningHash = stakeC;
  secondPlaceHash = stakeB;
}
```

nuevoalex: Yours has less indentation and is probably more intuitive so I'll go with it!

nuevoalex: Well, there is the issue that treating non existent hashes and null hashes at 0 stake will incorrectly promote them in some scenarios where we have negative values. I'll think about how we could still organize this better.

MicahZoltu: Perhaps it is reasonable to consider anything with `0` stake to be `0` hash?  `secondPlaceHash = (stakeA != 0) ? hashA : bytes32(0);` for each of the assignment lines?

nuevoalex: That worries me for a case where somehow legitimate staked REP cancels out a bond to result at 0. I know it's unlikely but it'd be possible.

Would it be too inelegant to have the `getPayoutDistributionHashStake` function return the smallest possible value based on negative stake from bonds if the has or token doesn't exist? That would make the above work quite nicely I think.

MicahZoltu: If a bond cancels out stake and brings it to 0 I think that should count as "not reported on" for the sake of punting the market to the next window.  That is, I don't think any outcome should be able to win with <= 0 stake.

MicahZoltu: Could we have `getPayoutDistributionHashStake` return -2^255 (INT_MIN) if the payout hash doesn't exist?  Would that help simplify things in the way you suggested?  If we require at least positive stake (> 0), do we still need to have that function return a anything other than `0` for null hashes?

MicahZoltu:
Updating my previous comment to:
```
winningHash = (stakeX > 0) ? hashX : bytes32(0);
secondPlaceHash = (stakeY > 0) ? hashY : bytes32(0);
```

source/contracts/reporting/Market.sol
nuevoalex: If we want to enforce negative hashes being unable to win we'll need to add some tests and _may_ need to have additional code written as currently a negatively staked outcome can win (if the only two staked outcomes are negative and one is greater for example).

If we go with that rule then it should be fine to have the function return 0 for those, although as I said I _think_ we'll need to add additional code to validate the tentative winning hash as > 0 in some places

MicahZoltu: In that case I think my vote would be to add the necessary code to disallow winning with <= 0 stake.  An outcome that has negative stake definitely feels like it shouldn't win because reporters have effectively asserted that "some outcome _other_ than this one is correct".  If we have to build out code to deal with that scenario, then extending that to include `0` seems pretty trivial and make it so we don't have to deal with the `0` boundary condition.

source/contracts/reporting/Market.sol
```diff
@@ -170,41 +170,40 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
...
-        IReportingToken _tentativeWinningReportingToken = reportingTokens[tentativeWinningPayoutDistributionHash];
-        if (address(_tentativeWinningReportingToken) == NULL_ADDRESS) {
-            tentativeWinningPayoutDistributionHash = _payoutDistributionHash;
-            return true;
+        int256 _tentativeWinningStake = getPayoutDistributionHashStake(tentativeWinningPayoutDistributionHash);
+        int256 _secondPlaceStake = getPayoutDistributionHashStake(bestGuessSecondPlaceTentativeWinningPayoutDistributionHash);
+        int256 _payoutStake = getPayoutDistributionHashStake(_payoutDistributionHash);
+
+        if (_tentativeWinningStake >= _payoutStake && _payoutStake >= _secondPlaceStake) {
+            bestGuessSecondPlaceTentativeWinningPayoutDistributionHash = (_payoutStake > 0) ? _payoutDistributionHash : bytes32(0);
```
MicahZoltu: I recognize why you left out the assignment to `tentativeWinningPayoutDistributionHash`, but does this create a situation where it is possible that we'll keep a `tentativeWinningPayoutDistributionHash` with a negative stake when we should be zeroing it at?

Same question (possibly different answer) for the last else and the one missing one.

nuevoalex: Yes it does (to both) ! I removed these before I added the ternary bit so it seemed redundant. Good catch!

source/contracts/reporting/Market.sol
```diff
@@ -171,30 +170,72 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
...
+        } else if (_secondPlaceStake >= _payoutStake && _payoutStake >= _tentativeWinningStake) {
+            tentativeWinningPayoutDistributionHash = (_secondPlaceStake > 0) ? bestGuessSecondPlaceTentativeWinningPayoutDistributionHash: bytes32(0);
+            bestGuessSecondPlaceTentativeWinningPayoutDistributionHash = (_payoutStake > 0) ? _payoutDistributionHash: bytes32(0);
+        } else if (_payoutStake >= _tentativeWinningStake && _tentativeWinningStake >= _secondPlaceStake) {
+            tentativeWinningPayoutDistributionHash = (_payoutStake > 0) ? _payoutDistributionHash: bytes32(0);
+            bestGuessSecondPlaceTentativeWinningPayoutDistributionHash = (_tentativeWinningStake > 0) ? tentativeWinningPayoutDistributionHash: bytes32(0);
+        } else if (_payoutStake >= _secondPlaceStake && _secondPlaceStake >= _tentativeWinningStake) {
+            tentativeWinningPayoutDistributionHash = (_payoutStake > 0) ? _payoutDistributionHash: bytes32(0);
+            bestGuessSecondPlaceTentativeWinningPayoutDistributionHash = (_secondPlaceStake > 0) ? bestGuessSecondPlaceTentativeWinningPayoutDistributionHash: bytes32(0);
+        }
```
MicahZoltu: There is one possible outcome set missing here... not sure which one it is but there should be 6 blocks but only 5 exist:
```
A > B > C
A > C > B
B > A > C
B > C > A
C > A > B
C > B > A
```

nuevoalex: Ah, I deleted another set that appeared redundant at the time (A > B > C). Will add back.

PR #318, opened 2017-09-28
make -> create & take -> fill

nuevoalex:


nuevoalex:
Theres a commit missing that does this I forgot to push apparently. Will do!

PR #319, opened 2017-09-28
Magic number eternal approval

nuevoalex:


MicahZoltu:
I'll admit, I thought this change would be harder than that.  😛

PR #321, opened 2017-09-29
Created unit test to test reentrancy_guard. tests method marked with …

bthaile:
created py tester for reentrancy guard and helper to use nonReentrant marking

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -158,11 +158,8 @@ class ContractsFixture:
         print 'Gas Used: %s' % (self.chain.head_state.gas_used - startingGas)
         self.scalarMarket = self.createReasonableScalarMarket(self.universe, 40, self.cash)
         self.constants = self.uploadAndAddToController("solidity_test_helpers/Constants.sol")
-        self.chain.mine(1)
-        self.originalHead = self.chain.head_state
-        self.originalBlock = self.chain.block
-        self.originalContracts = deepcopy(self.contracts)
-        self.snapshot = self.chain.snapshot()
+        self.captured = {}
+        self.createSnapshot
 
     def uploadAndAddToController(self, relativeFilePath, lookupKey = None, signatureKey = None, constructorArgs=[]):
         lookupKey = lookupKey if lookupKey else path.splitext(path.basename(relativeFilePath))[0]
@@ -202,11 +199,21 @@ class ContractsFixture:
         return contract
 
     def resetSnapshot(self):
-        self.chain.block = self.originalBlock
-        self.chain.revert(self.snapshot)
-        self.chain.head_state = self.originalHead
-        self.contracts = deepcopy(self.originalContracts)
+        self.resetToSnapshot(self.captured)
 
+    def createSnapshot(self):
+        self.chain.mine(1)
+        self.captured = {'block': self.chain.block, 'head_state': self.chain.head_state, 'contracts' : deepcopy(self.contracts), 'snapshot': self.chain.snapshot()}
+        return self.captured
+
+    def resetToSnapshot(self, captured):
+        if len(captured) > 0:
+            self.chain.block = captured['block']
+            self.chain.revert(captured['snapshot'])
+            self.chain.head_state = captured['head_state']
+            self.contracts = deepcopy(captured['contracts'])
+        else:
+            self.createSnapshot()
     ####
     #### Bulk Operations
     ####
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -205,13 +205,13 @@ class ContractsFixture:
         return {'block': self.chain.block, 'head_state': self.chain.head_state, 'contracts' : deepcopy(self.contracts), 'snapshot': self.chain.snapshot()}
 
     def resetToSnapshot(self, captured):
-        if len(captured) > 0:
-            self.chain.block = captured['block']
-            self.chain.revert(captured['snapshot'])
-            self.chain.head_state = captured['head_state']
-            self.contracts = deepcopy(captured['contracts'])
-        else:
-            self.createSnapshot()
+        if len(captured) < 4:
+            raise "captured snapshot doesn't have all parameters in dictionary, need to call createSnapshot"
+        self.chain.block = captured['block']
+        self.chain.revert(captured['snapshot'])
+        self.chain.head_state = captured['head_state']
+        self.contracts = deepcopy(captured['contracts'])
+
     ####
     #### Bulk Operations
     ####
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -158,8 +158,7 @@ class ContractsFixture:
         print 'Gas Used: %s' % (self.chain.head_state.gas_used - startingGas)
         self.scalarMarket = self.createReasonableScalarMarket(self.universe, 40, self.cash)
         self.constants = self.uploadAndAddToController("solidity_test_helpers/Constants.sol")
-        self.captured = {}
-        self.createSnapshot
+        self.captured = self.createSnapshot()
 
     def uploadAndAddToController(self, relativeFilePath, lookupKey = None, signatureKey = None, constructorArgs=[]):
         lookupKey = lookupKey if lookupKey else path.splitext(path.basename(relativeFilePath))[0]
@@ -203,8 +202,7 @@ class ContractsFixture:
 
     def createSnapshot(self):
         self.chain.mine(1)
-        self.captured = {'block': self.chain.block, 'head_state': self.chain.head_state, 'contracts' : deepcopy(self.contracts), 'snapshot': self.chain.snapshot()}
-        return self.captured
+        return {'block': self.chain.block, 'head_state': self.chain.head_state, 'contracts' : deepcopy(self.contracts), 'snapshot': self.chain.snapshot()}
 
     def resetToSnapshot(self, captured):
         if len(captured) > 0:
--- a/tests/libraries/test_reentrancy_guard.py
+++ b/tests/libraries/test_reentrancy_guard.py
@@ -21,8 +21,6 @@ def testerContractsFixture(sessionFixture, testerSnapshot):
 
 def test_nonReentrant(testerContractsFixture): 
     ReentrancyGuardHelper = testerContractsFixture.contracts['ReentrancyGuardHelper']
-    # Whitelist the contract
-    testerContractsFixture.controller.addToWhitelist(ReentrancyGuardHelper.address)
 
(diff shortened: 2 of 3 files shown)


tests/libraries/test_reentrancy_guard.py
```diff
@@ -0,0 +1,33 @@
...
+
+from ethereum.tools import tester
+from ethereum.tools.tester import TransactionFailed
+from pytest import fixture, mark, lazy_fixture, raises
+
+
+@fixture(scope='session')
+def testerSnapshot(sessionFixture):
+    sessionFixture.resetSnapshot()
+    sessionFixture.uploadAndAddToController(
```
nuevoalex: No need to put this on another line

tests/libraries/test_reentrancy_guard.py
```diff
@@ -0,0 +1,33 @@
...
+
+@fixture
+def testerContractsFixture(sessionFixture, testerSnapshot):
+    sessionFixture.chain.revert(testerSnapshot)
+    return sessionFixture
+
+
+def test_nonReentrant(testerContractsFixture):
+    ReentrancyGuardHelper = testerContractsFixture.contracts[
+        'ReentrancyGuardHelper']
```
nuevoalex: same here w/ regard to the extra line

tests/libraries/test_reentrancy_guard.py
bthaile: damn! the formatter

tests/libraries/test_reentrancy_guard.py
```diff
@@ -0,0 +1,33 @@
...
+    # Whitelist the contract
+    testerContractsFixture.controller.addToWhitelist(
+        ReentrancyGuardHelper.address)
+
+    assert ReentrancyGuardHelper.testerCanReentrant() == True
```
nuevoalex: We usually just leave out the `== True` for test asserts

bthaile: ok, I can make that change.

tests/libraries/test_reentrancy_guard.py
bthaile: formatter did that as well, using

tests/solidity_test_helpers/ReentrancyGuardHelper.sol
```diff
@@ -0,0 +1,21 @@
+pragma solidity ^0.4.13;
+
+import 'libraries/ReentrancyGuard.sol';
+
```
nuevoalex: 2 Lines before the contract definition. You can also test this style enforcement by using solium locally with this command: `solium --dir=source/contracts`

bthaile:
Want me to make changes and re-submit?

tests/libraries/test_reentrancy_guard.py
nuevoalex: This ones still here ;p

bthaile: DOH!

tests/libraries/test_reentrancy_guard.py
```diff
@@ -0,0 +1,30 @@
...
+def testerSnapshot(sessionFixture):
+    sessionFixture.resetSnapshot()
+    sessionFixture.uploadAndAddToController('solidity_test_helpers/ReentrancyGuardHelper.sol')
+    ReentrancyGuardHelper = sessionFixture.contracts['ReentrancyGuardHelper']
+    return sessionFixture.chain.snapshot()
+
+
+@fixture
+def testerContractsFixture(sessionFixture, testerSnapshot):
+    sessionFixture.chain.revert(testerSnapshot)
```
MicahZoltu: There are several other things that need to happen along with reverting the snaptshot.  The following is what happens in `sessionFixture.resetSnapshot()`.
```
        self.chain.block = self.originalBlock
        self.chain.revert(self.snapshot)
        self.chain.head_state = self.originalHead
        self.contracts = deepcopy(self.originalContracts)
```

Also, we do the following when creating snapshots, which we should also do:
```
        self.originalHead = self.chain.head_state
        self.originalBlock = self.chain.block
        self.originalContracts = deepcopy(self.contracts)
        self.snapshot = self.chain.snapshot()
```

It definitely seems like there would be value in having the `ContractsFixture` have a method for create snapshot that returns an object containing all of those pieces of data and another method for resetting a snapshot that takes in the same object type.  This way we don't have to keep re-implementing everything correctly anytime we want a new derived fixture.

tests/libraries/test_reentrancy_guard.py
```diff
@@ -0,0 +1,30 @@
...
+    return sessionFixture
+
+
+def test_nonReentrant(testerContractsFixture): 
+    ReentrancyGuardHelper = testerContractsFixture.contracts['ReentrancyGuardHelper']
+    # Whitelist the contract
+    testerContractsFixture.controller.addToWhitelist(ReentrancyGuardHelper.address)
```
MicahZoltu: This is unnecessary for this test as the reentrancy guard doesn't depend on being whitelisted.  Recommend removing it so the test is more clear as what is required for it to pass.

tests/solidity_test_helpers/ReentrancyGuardHelper.sol
bthaile: what is meant by 'test this style enforcement' also I don't have an executable solium. I can see if I can download it on macos.

tests/libraries/test_reentrancy_guard.py
bthaile: I'll remove it.

tests/solidity_test_helpers/ReentrancyGuardHelper.sol
nuevoalex: Solium is a command line tool used to test style guidelines for Solidity. The easiest way to install it would be to use npm.

bthaile: command
>npx solium --dir=source/contracts
No issues found.

Is that what I'm suppose to see?

MicahZoltu:
```
npm install
npx solium --dir=source/contracts
```
(assumes Node 7+ I believe)

MicahZoltu: Yes, that means everything is good. If you have only a single blank line between `import` and `contracts`  and you still see "No issues found" then something is wrong though.

nuevoalex: I'm realizing now our CI and the default command we use `dir=source/contracts` won't analyze the contracts in `tests`. That's why CI is passing and this doesn't come up.

MicahZoltu: Ah.  If someone wants to update the docs and Dockerfile and maybe add an `npm lint` command I'll approve the PR, but I don't particularly care that much so I probably won't do this myself.  😄

tests/conftest.py
```diff
@@ -158,11 +158,8 @@ def __init__(self):
         print 'Gas Used: %s' % (self.chain.head_state.gas_used - startingGas)
         self.scalarMarket = self.createReasonableScalarMarket(self.universe, 40, self.cash)
         self.constants = self.uploadAndAddToController("solidity_test_helpers/Constants.sol")
-        self.chain.mine(1)
-        self.originalHead = self.chain.head_state
-        self.originalBlock = self.chain.block
-        self.originalContracts = deepcopy(self.contracts)
-        self.snapshot = self.chain.snapshot()
+        self.captured = {}
+        self.createSnapshot
```
MicahZoltu: Style: Add `()` at the end of this line so it is clear that this is a function call.

tests/conftest.py
```diff
@@ -202,11 +199,21 @@ def applySignature(self, signatureName, address):
...
     def resetSnapshot(self):
-        self.chain.block = self.originalBlock
-        self.chain.revert(self.snapshot)
-        self.chain.head_state = self.originalHead
-        self.contracts = deepcopy(self.originalContracts)
+        self.resetToSnapshot(self.captured)
 
+    def createSnapshot(self):
+        self.chain.mine(1)
+        self.captured = {'block': self.chain.block, 'head_state': self.chain.head_state, 'contracts' : deepcopy(self.contracts), 'snapshot': self.chain.snapshot()}
```
MicahZoltu: Don't assign to `self.captured` here.  Instead just return this dictionary and then up on line 162 do: `self.captured = self.createSnapshot()`.  This also means you don't need to do line 161 (`self.captured = {}`).

bthaile: doh! makes sense.

tests/conftest.py
```diff
@@ -202,11 +199,21 @@ def applySignature(self, signatureName, address):
...
+        return self.captured
+
+    def resetToSnapshot(self, captured):
+        if len(captured) > 0:
+            self.chain.block = captured['block']
+            self.chain.revert(captured['snapshot'])
+            self.chain.head_state = captured['head_state']
+            self.contracts = deepcopy(captured['contracts'])
+        else:
+            self.createSnapshot()
```
MicahZoltu: Recommend removing this `if/else` block as creating a snapshot if someone tries to reset is the wrong behavior.  It may be good to do
`if len(captured) <= 4: raise Exception("invalid snapshot")`
(or however you throw in Python)

tests/conftest.py
MicahZoltu: The reason for this is that various fixtures will create different snapshots and so there may be many snapshots floating around in the wild at a time, not just one.

tests/libraries/test_reentrancy_guard.py
```diff
@@ -0,0 +1,24 @@
...
+from ethereum.tools import tester
+from ethereum.tools.tester import TransactionFailed
+from pytest import fixture, mark, lazy_fixture, raises
+
+@fixture(scope='session')
+def testerSnapshot(sessionFixture):
+    sessionFixture.resetSnapshot()
+    sessionFixture.uploadAndAddToController('solidity_test_helpers/ReentrancyGuardHelper.sol')
+    ReentrancyGuardHelper = sessionFixture.contracts['ReentrancyGuardHelper']
+    return sessionFixture.chain.snapshot()
```
MicahZoltu: This should use the new snapshot stuff:
```
return sessionFixture.createSnapshot()
```
And then down below we would do:
```
sessionFixture.resetToSnapshot(testerSnapshot)
```

PR #322, opened 2017-09-30
added more tests to standard token. tests to attempt to get negative …

bthaile:
working with what was initially there, added more tests to standard token. tests to attempt to get negative balance. test to ping pong transfers.

### DIFF
--- a/tests/libraries/token/test_standard_token.py
+++ b/tests/libraries/token/test_standard_token.py
@@ -117,3 +117,21 @@ def test_ping_pong_transfers(testStandardTokenFixture):
     assert standardToken.balanceOf(tester.a0) == 100
     assert standardToken.balanceOf(tester.a1) == 100
 
+def test_transfrom_without_approve(testStandardTokenFixture):
+    standardToken = testStandardTokenFixture.contracts['StandardTokenHelper']
+    assert standardToken.totalSupply() == 0
+    # We get some tokens for tester 0
+    assert standardToken.faucet(10)
+    with raises(TransactionFailed):
+        standardToken.transferFrom(tester.a0,tester.a1, 10)
+
+def test_transfrom_more_than_total_supply(testStandardTokenFixture):
+    standardToken = testStandardTokenFixture.contracts['StandardTokenHelper']
+    assert standardToken.totalSupply() == 0
+    assert standardToken.faucet(10)
+    assert standardToken.totalSupply() == 10
+    assert standardToken.faucet(10)
+    assert standardToken.totalSupply() == 20
+    with raises(TransactionFailed):
+        standardToken.transfer(tester.a1, 21)
+    
--- a/tests/libraries/token/test_standard_token.py
+++ b/tests/libraries/token/test_standard_token.py
@@ -37,3 +37,85 @@ def test_eternal_approval_magic(contractsFixture):
     assert standardToken.balanceOf(tester.a0) == 4
     assert standardToken.balanceOf(tester.a1) == 96
     
+def test_create_negative_balance(contractsFixture):
+    standardToken = contractsFixture.uploadAndAddToController("solidity_test_helpers/StandardTokenHelper.sol")
+    assert standardToken.totalSupply() == 0
+    # We get some tokens for tester 0
+    assert standardToken.faucet(100)
+    assert standardToken.balanceOf(tester.a0) == 100
+    # Tester 1 tries to send tokens from tester 0 to himself and fails
+    with raises(TransactionFailed):
+        standardToken.transferFrom(tester.a0, tester.a1, 101)
+
+def test_transfer_then_create_negative_balance(contractsFixture):
+    standardToken = contractsFixture.uploadAndAddToController("solidity_test_helpers/StandardTokenHelper.sol")
+    assert standardToken.totalSupply() == 0
+    # We get some tokens for tester 0
+    assert standardToken.faucet(10)
+    assert standardToken.balanceOf(tester.a0) == 10
+    # approve and transfer 10 to other tester
+    assert standardToken.approve(tester.a1, 10)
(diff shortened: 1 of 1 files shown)


tests/libraries/token/test_standard_token.py
```diff
@@ -36,4 +36,86 @@ def test_eternal_approval_magic(contractsFixture):
...
+    
+def test_create_negative_balance(contractsFixture):
+    standardToken = contractsFixture.uploadAndAddToController("solidity_test_helpers/StandardTokenHelper.sol")
+    assert standardToken.totalSupply() == 0
+    # We get some tokens for tester 0
+    assert standardToken.faucet(100)
+    assert standardToken.balanceOf(tester.a0) == 100
+    # Tester 1 tries to send tokens from tester 0 to himself and fails
+    with raises(TransactionFailed):
+        standardToken.transferFrom(tester.a0, tester.a1, 101)
```
MicahZoltu: `transferFrom` in general, is only used if you are trying to send someone else's tokens.  If you want to send your own tokens you would use `transfer`.  If you want to specify who the sender is you can do `standardToken.transferFrom(tester.a1, 101, sender=tester.k0)`.  In this particular case, `tester.k0` is the default so you could leave it out.

tests/libraries/token/test_standard_token.py
```diff
@@ -36,4 +36,86 @@ def test_eternal_approval_magic(contractsFixture):
...
+
+def test_transfer_then_create_negative_balance(contractsFixture):
+    standardToken = contractsFixture.uploadAndAddToController("solidity_test_helpers/StandardTokenHelper.sol")
```
MicahZoltu: Uploading contracts can be a marginally time consuming operation.  Since you have several tests that all use this, consider creating a fixture with a snapshot like you did previously so you can just reset back to that state rather than having to re-upload each test.

tests/libraries/token/test_standard_token.py
bthaile: ok, removed the unnecessary tester.k0 and changed the transferFrom()  to transfer()

tests/libraries/token/test_standard_token.py
bthaile: add test fixture to minimize the thrashing of uploading contract

tests/libraries/token/test_standard_token.py
```diff
@@ -2,12 +2,25 @@
...
+def testerSnapshot(sessionFixture):
+    sessionFixture.resetSnapshot()
+    sessionFixture.uploadAndAddToController("solidity_test_helpers/StandardTokenHelper.sol")
+    standardToken = sessionFixture.contracts['StandardTokenHelper']
+    return sessionFixture.chain.snapshot()
 
-def test_eternal_approval_magic(contractsFixture):
+
+@fixture
+def testerContractsFixture(sessionFixture, testerSnapshot):
```
MicahZoltu: Consider coming up with a more pithy name for this fixture.  `testerContractsFixture` doesn't really imply what it is used for.  Also, since this is a local fixture feel free to use something shorter than normal like `fixture` (unfortunately that name is taken though).

MicahZoltu: Maybe `tokenFixture`?

tests/libraries/token/test_standard_token.py
```diff
@@ -2,12 +2,25 @@
...
+    sessionFixture.chain.revert(testerSnapshot)
+    return sessionFixture
+
+
```
MicahZoltu: Style: Only one newline between tests.

tests/libraries/token/test_standard_token.py
```diff
@@ -36,4 +49,80 @@ def test_eternal_approval_magic(contractsFixture):
...
+    # approve and transfer 10 to other tester
+    assert standardToken.approve(tester.a1, 10)
+    assert standardToken.allowance(tester.a0, tester.a1) == 10
+    assert standardToken.transfer(tester.a1, 10)
+    assert standardToken.balanceOf(tester.a1) == 10
+    # approve and allow transfer of more tokens than tester has
+    assert standardToken.approve(tester.a1, 11)
+    # try to transfer more than tester has  
+    with raises(TransactionFailed):
+        standardToken.transfer(tester.a1, 11)
```
MicahZoltu: I don't believe this test is doing what you want.  You start out by giving 10 tokens to `a0`.  You then approve `a1` to spend your 10 tokens.  Then you transfer 10 tokens to `a1`, but the approval isn't used at all because the approval was to allow `a1` to transfer the tokens but you transferred the tokens as `a0` (the token owner).  If you want `a1` to transfer on behalf of `a0` then you need to use `transferFrom(tester.a0, tester.a2, 10, sender=tester.k1)`.  This will have `a1` send 10 of `a0`s tokens to `a2`.

MicahZoltu: Alternatively, just remove the approvals if your intent is to test the `transfer` function rather than `approvals`.

MicahZoltu: Same issue in a couple other tests where you are setting an approval and then not leveraging it at all in the test.

tests/libraries/token/test_standard_token.py
```diff
@@ -2,12 +2,23 @@
...
 from ethereum.tools.tester import TransactionFailed
-from pytest import raises
+from pytest import fixture, mark, lazy_fixture, raises
 
+@fixture(scope='session')
+def testerSnapshot(sessionFixture):
+    sessionFixture.resetSnapshot()
+    sessionFixture.uploadAndAddToController("solidity_test_helpers/StandardTokenHelper.sol")
+    standardToken = sessionFixture.contracts['StandardTokenHelper']
+    return sessionFixture.chain.snapshot()
```
MicahZoltu: Recommend updating this to follow the new pattern.

PR #330, opened 2017-10-01
Fee calc todos

nuevoalex:
Addresses all but one of the TODOs in the `MarketFeeCalculator`. Namely the one about getting the market value of REP in ETH:

> get this from an auto-generated market or some other special contract

I figure that one will require its own story and a significant amount of work to implement depending on how exactly we choose to do it.

One decision I made here which may be bad is that the default gas price to report and the default average reports per market are factored into the real value calculations. This was done to avoid branching code since the methods will be called relatively frequently and these are just estimates anyway. It'd be easy to make these instead just branch on `numReports < 0` and return "true" values.

### DIFF
--- a/source/contracts/extensions/MarketFeeCalculator.sol
+++ b/source/contracts/extensions/MarketFeeCalculator.sol
@@ -43,20 +43,25 @@ contract MarketFeeCalculator {
         
         uint256 _targetIndeterminateMarkets = _totalMarkets.div(_targetIndeterminateMarketsDivisor);
 
+        // Modify the validity bond based on the previous amount and the number of indeterminate markets. We want the bond to be somewhere in the range of 0.5 to 2 times its previous value where ALL markets being indeterminate results in 2x and 0 indeterminate results in 0.5x.
         if (_indeterminateMarkets <= _targetIndeterminateMarkets) {
+            // FXP formula: previous_bond * percent_invalid / (2 * target_percent_invalid) + 0.5;
             return _indeterminateMarkets
                 .mul(_previousValidityBondInAttoeth)
                 .mul(_targetIndeterminateMarketsDivisor)
                 .div(_totalMarkets)
                 .div(2)
-            .add(_previousValidityBondInAttoeth.div(2));
+                .add(_previousValidityBondInAttoeth.div(2))
+            ; // FIXME: This is here due to a solium bug
         } else {
+            // FXP formula: previous_bond * (1/(1 - target_percent_invalid)) * (percent_invalid - target_percent_invalid) + 1;
             return _targetIndeterminateMarketsDivisor
                 .mul(_previousValidityBondInAttoeth.mul(_indeterminateMarkets)
                 .div(_totalMarkets)
                 .sub(_previousValidityBondInAttoeth.div(_targetIndeterminateMarketsDivisor)))
                 .div(_targetIndeterminateMarketsDivisor - 1)
-            .add(_previousValidityBondInAttoeth);
+                .add(_previousValidityBondInAttoeth)
+            ; // FIXME: This is here due to a solium bug
         }
     }
 
--- a/source/contracts/extensions/MarketFeeCalculator.sol
+++ b/source/contracts/extensions/MarketFeeCalculator.sol
@@ -14,8 +14,8 @@ contract MarketFeeCalculator {
     mapping (address => uint256) private validityBondInAttoeth;
     mapping (address => uint256) private targetReporterGasCosts;
 
-    uint256 private constant DEFAULT_VALIDITY_BOND = 1 ether;
-    uint256 private constant FXP_TARGET_INDETERMINATE_MARKETS = 10 ** 16; // 1% of markets
+    uint256 private constant DEFAULT_VALIDITY_BOND = 1 ether / 100;
+    uint256 private constant FXP_TARGET_INDETERMINATE_MARKETS = 1 ether / 100; // 1% of markets
     uint256 private constant TARGET_REP_MARKET_CAP_MULTIPLIER = 5;
 
     function getValidityBond(IReportingWindow _reportingWindow) public returns (uint256) {
@@ -23,7 +23,7 @@ contract MarketFeeCalculator {
         if (_currentValidityBondInAttoeth != 0) {
             return _currentValidityBondInAttoeth;
         }
-        IReportingWindow _previousReportingWindow = getPreviousReportingWindow(_reportingWindow);
+        IReportingWindow _previousReportingWindow = _reportingWindow.getPreviousReportingWindow();
         uint256 _totalMarketsInPreviousWindow = _reportingWindow.getNumMarkets();
         uint256 _indeterminateMarketsInPreviousWindow = _reportingWindow.getNumIndeterminateMarkets();
         uint256 _previousValidityBondInAttoeth = validityBondInAttoeth[_previousReportingWindow];
@@ -60,7 +60,7 @@ contract MarketFeeCalculator {
             return _gasToReport;
         }
 
-        IReportingWindow _previousReportingWindow = getPreviousReportingWindow(_reportingWindow);
+        IReportingWindow _previousReportingWindow = _reportingWindow.getPreviousReportingWindow();
         uint256 _estimatedReportsPerMarket = _previousReportingWindow.getAvgReportsPerMarket();
         uint256 _avgGasCost = _previousReportingWindow.getAvgReportingGasCost();
         _gasToReport = Reporting.gasToReport();
@@ -78,7 +78,7 @@ contract MarketFeeCalculator {
         IUniverse _universe = _reportingWindow.getUniverse();
         uint256 _repMarketCapInAttoeth = getRepMarketCapInAttoeth(_universe);
         uint256 _targetRepMarketCapInAttoeth = getTargetRepMarketCapInAttoeth(_reportingWindow);
-        IReportingWindow _previousReportingWindow = getPreviousReportingWindow(_reportingWindow);
+        IReportingWindow _previousReportingWindow = _reportingWindow.getPreviousReportingWindow();
         uint256 _previousPerEthFee = shareSettlementPerEthFee[_previousReportingWindow];
         if (_previousPerEthFee == 0) {
             _previousPerEthFee = 1 * 10 ** 16;
@@ -101,12 +101,6 @@ contract MarketFeeCalculator {
     function getTargetRepMarketCapInAttoeth(IReportingWindow _reportingWindow) constant public returns (uint256) {
         IUniverse _universe = _reportingWindow.getUniverse();
         return _universe.getOpenInterestInAttoEth() * TARGET_REP_MARKET_CAP_MULTIPLIER;
-    }
-
-    function getPreviousReportingWindow(IReportingWindow _reportingWindow) constant private returns (IReportingWindow) {
-        uint256 _previousTimestamp = _reportingWindow.getStartTime() - 1;
-        IUniverse _universe = _reportingWindow.getUniverse();
-        return _universe.getReportingWindowByTimestamp(_previousTimestamp);
     }
 
     function getMarketCreationCost(IReportingWindow _reportingWindow) constant public returns (uint256) {
--- a/source/contracts/extensions/MarketFeeCalculator.sol
+++ b/source/contracts/extensions/MarketFeeCalculator.sol
@@ -3,32 +3,54 @@ pragma solidity ^0.4.13;
 import 'reporting/IReputationToken.sol';
 import 'reporting/IUniverse.sol';
 import 'reporting/IReportingWindow.sol';
+import 'libraries/math/SafeMathUint256.sol';
 
 
 contract MarketFeeCalculator {
+    using SafeMathUint256 for uint256;
+
     mapping (address => uint256) private shareSettlementPerEthFee;
     mapping (address => uint256) private validityBondInAttoeth;
+
+    uint256 private constant DEFAULT_VALIDITY_BOND = 1 ether;
+    uint256 private constant FXP_TARGET_INDETERMINATE_MARKETS = 10 ** 16; // 1% of markets
 
     function getValidityBond(IReportingWindow _reportingWindow) public returns (uint256) {
         uint256 _currentValidityBondInAttoeth = validityBondInAttoeth[_reportingWindow];
         if (_currentValidityBondInAttoeth != 0) {
             return _currentValidityBondInAttoeth;
         }
-        // TODO: get the real data for this
-        uint256 _indeterminateMarketsInPreviousWindow = 10;
-        // TODO: get the real data for this
-        uint256 _totalMarketsInPreviousWindow = 1000;
         uint256 _previousTimestamp = _reportingWindow.getStartTime() - 1;
         IUniverse _universe = _reportingWindow.getUniverse();
         IReportingWindow _previousReportingWindow = _universe.getReportingWindowByTimestamp(_previousTimestamp);
+        uint256 _totalMarketsInPreviousWindow = _reportingWindow.getNumMarkets();
+        uint256 _indeterminateMarketsInPreviousWindow = _reportingWindow.getNumIndeterminateMarkets();
         uint256 _previousValidityBondInAttoeth = validityBondInAttoeth[_previousReportingWindow];
-        if (_previousValidityBondInAttoeth == 0) {
-            _previousValidityBondInAttoeth = 1 * 10 ** 16;
-        }
-        uint256 _targetIndeterminateMarketsPerHundred = 1;
-        _currentValidityBondInAttoeth = _previousValidityBondInAttoeth * _targetIndeterminateMarketsPerHundred * _totalMarketsInPreviousWindow / _indeterminateMarketsInPreviousWindow / 100;
+
+        _currentValidityBondInAttoeth = calculateValidityBond(_indeterminateMarketsInPreviousWindow, _totalMarketsInPreviousWindow, FXP_TARGET_INDETERMINATE_MARKETS, _previousValidityBondInAttoeth);
         validityBondInAttoeth[_reportingWindow] = _currentValidityBondInAttoeth;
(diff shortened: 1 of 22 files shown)


source/contracts/extensions/MarketFeeCalculator.sol
```diff
@@ -3,45 +3,70 @@ pragma solidity ^0.4.13;
...
 
 contract MarketFeeCalculator {
+    using SafeMathUint256 for uint256;
+
     mapping (address => uint256) private shareSettlementPerEthFee;
     mapping (address => uint256) private validityBondInAttoeth;
+    mapping (address => uint256) private targetReporterGasCosts;
+
+    uint256 private constant DEFAULT_VALIDITY_BOND = 1 ether;
+    uint256 private constant FXP_TARGET_INDETERMINATE_MARKETS = 10 ** 16; // 1% of markets
```
MicahZoltu: Recommend a divisor instead of fixed point.  e.g., `20` to be used as `num_markets/20`.  Naming is tricky, but the math ends up being easier and you don't need to worry about correctly doing `fxpMul`.

source/contracts/extensions/MarketFeeCalculator.sol
```diff
@@ -3,45 +3,70 @@ pragma solidity ^0.4.13;
...
 
 
 contract MarketFeeCalculator {
+    using SafeMathUint256 for uint256;
+
     mapping (address => uint256) private shareSettlementPerEthFee;
     mapping (address => uint256) private validityBondInAttoeth;
+    mapping (address => uint256) private targetReporterGasCosts;
+
+    uint256 private constant DEFAULT_VALIDITY_BOND = 1 ether;
```
MicahZoltu: I think Joey was right that this is probably too high ($300).  Maybe `1 ether / 100` instead?

source/contracts/extensions/MarketFeeCalculator.sol
```diff
@@ -68,27 +92,24 @@ contract MarketFeeCalculator {
...
+        return _universe.getOpenInterestInAttoEth() * TARGET_REP_MARKET_CAP_MULTIPLIER;
     }
 
-    function getOutstandingSharesInAttoeth(IReportingWindow) constant public returns (uint256) {
-        // TODO: start tracking the real number and store it somewhere
-        // NOTE: make sure we are getting the share value in attoeth, since complete set fees are not normalized across markets
-        // NOTE: eventually we will need to support shares in multiple denominations
-        uint256 _outstandingSharesInAttoeth = 100 * 10 ** 18;
-        return _outstandingSharesInAttoeth;
+    function getPreviousReportingWindow(IReportingWindow _reportingWindow) constant private returns (IReportingWindow) {
```
MicahZoltu: I think there is some other place we are doing this as well.  I wonder if we should just put a function on `ReportingWindow` called `getPreviousReportingWindow` and `getNextReportingWindow` so we don't have to keep re-implementing this logic?

source/contracts/reporting/ReportingToken.sol
```diff
@@ -163,4 +163,13 @@ contract ReportingToken is DelegationTarget, Typed, Initializable, VariableSuppl
         require(index < market.getNumberOfOutcomes());
         return payoutNumerators[index];
     }
+
+    function isIndeterminate() public constant returns (bool) {
```
MicahZoltu: Would be nice to see a unit test of this method, mainly because I don't trust `.length` in Solidity (because its availability differs based on storage/memory).

source/contracts/reporting/Market.sol
```diff
@@ -440,6 +440,12 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
         return false;
     }
 
+    function isIndeterminate() public constant returns (bool) {
```
MicahZoltu: Totally unrelated to this PR, but I wonder if we should have a modifier for market that is `onlyAfterFinalized`?  I suspect that `getFinalWinningReportingToken` does this check, but it would be nice to have the methods annotated to indicate when they can be called.  Similarly, I wonder if there is value in some other modifiers in `Market` like `onlyBeforeFinalized`?

source/contracts/reporting/ReportingWindow.sol
```diff
@@ -283,6 +318,9 @@ contract ReportingWindow is DelegationTarget, Typed, Initializable, IReportingWi
             reporterStatus[_reporter].finishedReporting = true;
         }
         numberOfReportsByMarket[_market] += 1;
+        // Record gas price
+        reportingGasPriceSum = reportingGasPriceSum.add(tx.gasprice);
+        numReports++;
```
MicahZoltu: Since these two variables are inextricably linked, I wonder if there would be readability value in creating a library for `RunningAverage` like:
```
library RunningAverage {
    struct Data {
        uint256 sum;
        uint256 count;
    }
    function currentAverage(Data self) internal returns (uint256) {
        return self.sum / self.count;
    }
    function record(Data self, uint256 input) internal returns (bool) {
        self.count++;
        self.sum += input;
        return bool;
    }
}
```
I'm not sure if we'll use this construct elsewhere or not, but I find it nice to be able to keep these two pieces of data together in a construct that makes it clear to the reader how they are related.

MicahZoltu: Side-note: If both of these can be fit into `uint128` we can save on storage costs.

MicahZoltu: Actually, it appears we already have two constructs like this putting up a minor argument for DRY.

source/contracts/extensions/MarketFeeCalculator.sol
nuevoalex: That seems like a good idea. Will do

source/contracts/reporting/ReportingToken.sol
nuevoalex: Will do!

source/contracts/reporting/Market.sol
nuevoalex: I'll put in a consider comment rather than implementing this only because we're battling the contract size currently :[

source/contracts/reporting/ReportingWindow.sol
nuevoalex: Seems like a good idea, will do!

source/contracts/extensions/MarketFeeCalculator.sol
nuevoalex: I'm not sure how we would employ that value without assuming it is actually being used as `1 ether / DIVISOR`. The example `num_markets / 20` doesn't make sense to me based on the formula we decided to use.

nuevoalex: nevermind I get it, Will approach this next

source/contracts/extensions/MarketFeeCalculator.sol
```diff
@@ -3,45 +3,76 @@ pragma solidity ^0.4.13;
...
+        
+        uint256 _targetIndeterminateMarkets = _totalMarkets.div(_targetIndeterminateMarketsDivisor);
+
+        if (_indeterminateMarkets <= _targetIndeterminateMarkets) {
+            return _indeterminateMarkets
+                .mul(_previousValidityBondInAttoeth)
+                .mul(_targetIndeterminateMarketsDivisor)
+                .div(_totalMarkets)
+                .div(2)
+            .add(_previousValidityBondInAttoeth.div(2));
```
MicahZoltu: Indentation on this line is off.  If you did this because Solium complained, move the `;` to the next line.  I have a comment somewhere else with a link to a GitHub issue as a reminder that we should put the `;` back on the previous line once the Solium issue is resolved.

source/contracts/extensions/MarketFeeCalculator.sol
```diff
@@ -3,45 +3,76 @@ pragma solidity ^0.4.13;
...
+        } else {
+            return _targetIndeterminateMarketsDivisor
+                .mul(_previousValidityBondInAttoeth.mul(_indeterminateMarkets)
```
MicahZoltu: Recommend each operation on its own line, or all operations on the same line.  In this case, I think all on own line is more readable.

source/contracts/extensions/MarketFeeCalculator.sol
```diff
@@ -3,45 +3,76 @@ pragma solidity ^0.4.13;
...
+                .div(_totalMarkets)
+                .sub(_previousValidityBondInAttoeth.div(_targetIndeterminateMarketsDivisor)))
```
MicahZoltu: Same comment as above re: operations per line.

source/contracts/extensions/MarketFeeCalculator.sol
```diff
@@ -3,45 +3,76 @@ pragma solidity ^0.4.13;
...
+            return DEFAULT_VALIDITY_BOND;
+        }
+        if (_previousValidityBondInAttoeth == 0) {
+            _previousValidityBondInAttoeth = DEFAULT_VALIDITY_BOND;
+        }
+        
+        uint256 _targetIndeterminateMarkets = _totalMarkets.div(_targetIndeterminateMarketsDivisor);
+
+        if (_indeterminateMarkets <= _targetIndeterminateMarkets) {
+            return _indeterminateMarkets
```
MicahZoltu: There is probably value in a comment indicating what the original function was and what it does.  Something like, "This is an integer-only implementation of `...` that provides a linear change in validity bond each reporting window based on how far off it is from the target.  The range is 0.5x => 1x => 2x."  Feel free to wordsmith that as you see fit.

source/contracts/extensions/MarketFeeCalculator.sol
```diff
@@ -3,45 +3,76 @@ pragma solidity ^0.4.13;
...
-        return _estimatedReportingGas;
+    function calculateValidityBond(uint256 _indeterminateMarkets, uint256 _totalMarkets, uint256 _targetIndeterminateMarketsDivisor, uint256 _previousValidityBondInAttoeth) constant public returns (uint256) {
+        if (_totalMarkets == 0) {
+            return DEFAULT_VALIDITY_BOND;
+        }
+        if (_previousValidityBondInAttoeth == 0) {
+            _previousValidityBondInAttoeth = DEFAULT_VALIDITY_BOND;
+        }
+        
+        uint256 _targetIndeterminateMarkets = _totalMarkets.div(_targetIndeterminateMarketsDivisor);
```
MicahZoltu: If `_totalMarkets < 100` then this will evaluate to `0`.  Is that OK?  What if there are are 10 markets and 9 are indeterminate (90%), it feels like we shouldn't treat that like 0 indeterminate markets.

[Wolfram Alpha](http://www.wolframalpha.com/input/?i=a+%3C%3D+b%2Fc,+solve+for+b) to the rescue again (since I can' never remember inequality math):
```
if (_indeterminateMarkets.mul(_targetIndeterminateMarketDivisor) <= _targetIndeterminateMarkets) { ... }
```

Probably worth a comment indicating what is happening, since the multiplication of the divisor is a bit weird.

nuevoalex: `_totalMarkets` can be less than 100 without it necessarily evaluating to `0` since it's being divided by the target divisor. If the divisor is greater than the number of markets it is essentially the equivalent of saying we expect to get 0 indeterminate markets for that amount. (10 markets with a divisor of 20 means we expect 0.5 indeterminate markets, so 0 seems appropriate)

source/contracts/extensions/MarketFeeCalculator.sol
nuevoalex: solium doesn't seem to like that as the previous line doesn't end in a paren when this is done. Any solium tricks for this one?

source/contracts/reporting/ReportingWindow.sol
```diff
@@ -17,11 +17,13 @@ import 'factories/MarketFactory.sol';
 import 'factories/RegistrationTokenFactory.sol';
 import 'reporting/Reporting.sol';
 import 'libraries/math/SafeMathUint256.sol';
+import 'libraries/math/RunningAverage.sol';
```
MicahZoltu: I wonder if we could use one of the [other import formats](https://solidity.readthedocs.io/en/develop/layout-of-source-files.html#syntax-and-semantics) for libraries like this so we don't have to do `RunningAverage.Data`?  I haven't really thought it through or experimented, just came to mind since I learned about alternative import formats and aliasing today.

source/contracts/reporting/ReportingWindow.sol
```diff
@@ -34,16 +36,22 @@ contract ReportingWindow is DelegationTarget, Typed, Initializable, IReportingWi
...
 
     function initialize(IUniverse _universe, uint256 _reportingWindowId) public beforeInitialized returns (bool) {
         endInitialization();
         universe = _universe;
         startTime = _reportingWindowId * universe.getReportingPeriodDurationInSeconds();
         RegistrationTokenFactory _registrationTokenFactory = RegistrationTokenFactory(controller.lookup("RegistrationTokenFactory"));
         registrationToken = _registrationTokenFactory.createRegistrationToken(controller, this);
+        // Initialize these to some reasonable value to handle the first market ever created without branching code 
+        reportingGasPrice.record(Reporting.defaultReportingGasPrice());
+        marketReports.record(Reporting.defaultReportsPerMarket());
```
MicahZoltu: Are we OK with these weighing into the final averages?  Is the assumption that we should have enough markets to drown out any amount of outlying them may trigger?

source/contracts/trading/ClaimProceeds.sol
```diff
@@ -28,7 +28,11 @@ contract ClaimProceeds is CashAutoConverter, ReentrancyGuard, IClaimProceeds {
         for (uint8 _outcome = 0; _outcome < _market.getNumberOfOutcomes(); ++_outcome) {
             IShareToken _shareToken = _market.getShareToken(_outcome);
             uint256 _numberOfShares = _shareToken.balanceOf(msg.sender);
-            var (_shareHolderShare, _creatorShare, _reporterShare) = divideUpWinnings(_market, _winningReportingToken, _outcome, _numberOfShares);
+            var (_proceeds, _shareHolderShare, _creatorShare, _reporterShare) = divideUpWinnings(_market, _winningReportingToken, _outcome, _numberOfShares);
+
+            if (_proceeds > 0) {
+                _market.getUniverse().decrementOpenInterest(_proceeds);
+            }
```
MicahZoltu: 🤔 It would be more correct (game theory) to decrease open interest for all outstanding shares in a market when it is finalized.  This is likely not feasible, and we have to be _really_ careful to then make it so closing a complete set after a market is finalized doesn't decrease open interest (as that would double-count open interest decreases).

In general, the open interest calculations are _super_ scary to me since they are critical to the security model and they don't reset each reporting period/market like a lot of values.  This means that a subtle bug that slowly bleeds off open interest or slowly accumulates open interest could result in a long-term problem.  This may even be a problem with rounding errors (superman 2 attack).

So, unless you think the above thought is easy to implement and provide strong guarantees on, I'm OK with this resulting in a bit of an over-estimate of open interest.  It does introduce a different problem where over time, open interest will constantly grow (indefinitely) as money gets forgotten/lost in markets.

source/contracts/reporting/ReportingWindow.sol
nuevoalex: My thoughts on this are that since its just an estimate and it lets us avoid branching code in oft used getters its alright. I could be easily convinced otherwise though

source/contracts/reporting/ReportingWindow.sol
nuevoalex: I'll play around and see if i can make it something more palatable

source/contracts/trading/ClaimProceeds.sol
nuevoalex: I think it's probably doable. I'll attempt it and if it feels fragile we can go back to this

source/contracts/extensions/MarketFeeCalculator.sol
MicahZoltu: Oh, bleh.  Didn't notice the open without close.  Does it complain if we indent the following lines?  I'm now more concerned about the fact that it was so easy to miss the open here followed by a close several lines down.  Ideally I would prefer something like:
```
_targetIndeterminateMarketsDivisor
    .mul(_previousValidityBondInAttoeth
        .mul(_indeterminateMarkets)
        .div(_totalMarkets)
        .sub(_previousValidityBondInAttoeth
            .div(_targetIndeterminateMarketsDivisor)
        )
    )
    .div(_targetIndeterminateDivisor - 1)
    .add(_previousValidityBondInAttoeth);
```
Though, after typing that all out I have to wonder if there is some value in some intermediate variables for the nested section?  Problem of course is that we don't have a good name for them as they are just transient pieces of the formula... The advantage of the above, while awkward to type and read, is that it makes it clear when the nested math is happening so bad readers like me don't fail to miss the missing close parens.

nuevoalex: Yea solium dislikes extra indentation as well. As a bonus this function is close to stack depth problems so transient variables probably cant be an option either

source/contracts/extensions/MarketFeeCalculator.sol
MicahZoltu: Ah, so if we have a divisor of 100 (1% target) and 99 markets and 1 of those markets is indeterminate then we'll correctly hit the `else` branch because `(1 > 98/99)`.  So even though we rounded to 0, its fine because we'll still ultimately hit the correct code path.  👍

source/contracts/trading/ClaimProceeds.sol
MicahZoltu: Be incredibly careful with it.  If your gut tells you that it is at all risky I would say fall back to this.  Tracking open interest accurately over an indefinite period of time is a big fear of mine.

nuevoalex: I have a pattern in mind for doing this but implementing even a portion of it puts the `Market` contract over the size limit :/ I added a CONSIDER comment so maybe if/when we address Market in some larger capacity we can add this.

PR #331, opened 2017-10-01
had to move setting chain mine call to init method, I think it was ge…

bthaile:
Had to move setting chain mine call to init method, I think it was getting called multiple times. Changed new test_standard_token to use new fixture create snapshot and reset snapshot
Changed existing test cash auto wrapper to use new fixture methods as a verification all is working.


tests/conftest.py
```diff
@@ -158,6 +158,7 @@ def __init__(self):
         print 'Gas Used: %s' % (self.chain.head_state.gas_used - startingGas)
         self.scalarMarket = self.createReasonableScalarMarket(self.universe, 40, self.cash)
         self.constants = self.uploadAndAddToController("solidity_test_helpers/Constants.sol")
+        self.chain.mine(1)
```
MicahZoltu: `createSnapshot()` should be the correct place for the `chain.mine(1)` call.  Can you go into detail on why you believe this move is necessary?

Anytime someone wants to create a snapshot, they _must_ mine a block first.  By putting the call to mine inside the `createSnapshot` function we ensure that a block will always be mined before creating a snapshot.  By putting it here, snapshots created in other code paths will have to mine a block themselves.

MicahZoltu: Also, extra calls to `mine` shouldn't cause any problems.

tests/libraries/token/test_standard_token.py
```diff
@@ -9,11 +9,11 @@ def testerSnapshot(sessionFixture):
     sessionFixture.resetSnapshot()
     sessionFixture.uploadAndAddToController("solidity_test_helpers/StandardTokenHelper.sol")
     standardToken = sessionFixture.contracts['StandardTokenHelper']
-    return sessionFixture.chain.snapshot()
+    return sessionFixture.createSnapshot()
 
 @fixture
 def testStandardTokenFixture(sessionFixture, testerSnapshot):
-    sessionFixture.chain.revert(testerSnapshot)
+    sessionFixture.resetToSnapshot(testerSnapshot)
```
MicahZoltu: There are a handful of other places we follow this pattern.  Feel free to hunt them down and update them as well, possibly in another PR though.

tests/conftest.py
bthaile: Issue comes up when multiple tests use createSnapshot() the second test suit gets errors: 

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/conftest.py:205: in createSnapshot
    self.chain.mine(1)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <ethereum.tools.tester.Chain object at 0x10e7c9290>, number_of_blocks = 1
coinbase = '\x82\xa9x\xb3\xf5\x96*[\tW\xd9\xee\x9e\xefG.\xe5[B\xf1'

    def mine(self, number_of_blocks=1, coinbase=a0):
        self.cs.finalize(self.head_state, self.block)
        set_execution_results(self.head_state, self.block)
        self.block = Miner(self.block).mine(rounds=100, start_nonce=0)
>       assert self.chain.add_block(self.block)
E       AssertionError

/usr/local/lib/python2.7/site-packages/ethereum/tools/tester.py:167: AssertionError

What I don't know is why.

tests/libraries/token/test_standard_token.py
bthaile: Once I understand the .mine() issue I can go hunt and change tests to use new pattern.

tests/conftest.py
MicahZoltu: Hmm, I'll checkout your branch and take a look to see if I can help figure out what the root cause is.  That is odd.

MicahZoltu:
Looks like the issue is that we are doing timestamps wrong pretty much everywhere in the tests.  In `test_registration_token_redemption.py` in the `registrationTokenRedemptionSnapshot` function it sets the [`chain.head_state.timestamp`](https://github.com/AugurProject/augur-core/pull/331/files#diff-cf2a44686a6620623a17967cec6cc35cR213).  On the following line if you do `sessionFixture.chain.mine()` the problem is resolved.  You should then be able to leave the `.mine` call in `createSnapshot` and everything will work correctly.

tests/libraries/token/test_standard_token.py
bthaile: updated to use new conftest.py helper methods to change the chain.

bthaile:
This PR can be removed, migrated code changes to origin/ut_updates_for_new_version_pyethereum branch. No useful changes here.

PR #332, opened 2017-10-01
include tests in solium linting for CI.

bthaile:
Need to run solium on whole project not just on source/contracts this will pick up tests. I didn't see option to pass in multiple directories to solium --dir source/contracts test or --dir source/contracts;tests wasn't acceptable.

### DIFF
--- a/.travis/ci.sh
+++ b/.travis/ci.sh
@@ -3,6 +3,6 @@
 set -e
 
 echo "Starting Solidity Linting..."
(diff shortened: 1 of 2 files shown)


bthaile:
merging to include tests directory. To exclude node_modules could create **code** directory to put source and tests in and point solium to **code** dir.

MicahZoltu:
Actually, that is a good point.  At some point we'll be moving the `tests` directory into `source`.  When that happens we don't need to worry about ignoring directories.  (not sure yet when we'll actually make that move)

PR #333, opened 2017-10-01
Dev environment improvements.

MicahZoltu:
Contracts don't currently compile against solc 0.4.16, so rolling back to 0.4.15.

Disables support for JS intermixed with TS.

Fixes debugger breakpoints not lining up correctly between VSCode and node runtime.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -8,10 +8,14 @@
             "program": "${workspaceRoot}/node_modules/mocha/bin/_mocha",
             "cwd": "${workspaceRoot}",
             "args": [
-                "-r",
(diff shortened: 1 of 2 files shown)


joeykrug:
> Contracts don't currently compile against solc 0.4.16, so rolling back to 0.4.15.

Why don't they compile against .4.16?

MicahZoltu:
@joeykrug Not sure why I _just_ received GitHub notice of your comment from 2 days ago, but the static analyzer got some new features and throws a bunch of warnings on us.  We just need to go through and fix them, but don't want to block the dev pipeline while someone does that.

PR #334, opened 2017-10-03
Todos fixmes

nuevoalex:
The biggest portion of this is implementing distribution for market creation fees on market finalization.

I had to mess with some stuff in `Market` that may not seem purposeful in order to get contract size within the limit. I have some rough ideas for how we could possibly split up Market too that I will write up soon.

### DIFF
--- a/source/contracts/reporting/Market.sol
+++ b/source/contracts/reporting/Market.sol
@@ -315,7 +315,7 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
 
     function doFeePayout(bool _toOwner, uint256 _amount) private returns (bool) {
         if (_toOwner) {
-            getOwner().transfer(_amount);
+            getOwner().call.value(_amount)();
         } else {
             getReportingWindow().receiveValidityBond.value(_amount)();
         }
--- a/source/contracts/reporting/Market.sol
+++ b/source/contracts/reporting/Market.sol
@@ -70,7 +70,7 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
         require(feePerEthInAttoeth <= MAX_FEE_PER_ETH_IN_ATTOETH);
         require(_creator != NULL_ADDRESS);
         require(_cash.getTypeName() == "Cash");
-        // FIXME: require market to be on a non-forking universe; repeat this check up the stack as well if necessary (e.g., in reporting window)
+        //require(address(getUniverse().getForkingMarket()) == NULL_ADDRESS);
         // CONSIDER: should we allow creator to send extra ETH, is there risk of variability in bond requirements?
         require(msg.value == MarketFeeCalculator(controller.lookup("MarketFeeCalculator")).getMarketCreationCost(_reportingWindow));
         reportingWindow = _reportingWindow;
@@ -87,16 +87,6 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
         }
         approveSpenders();
         return true;
-
-        // TODO: we need to update this signature (and all of the places that call it) to allow the creator (UI) to pass in a number of other things which will all be logged here
-        // TODO: log short description
-        // TODO: log long description
-        // TODO: log min display price
-        // TODO: log max display price
-        // TODO: log tags (0-2)
-        // TODO: log outcome labels (same number as numOutcomes)
-        // TODO: log type (scalar, binary, categorical)
-        // TODO: log any immutable data associated with the market (e.g., endTime, numOutcomes, numTicks, cash address, etc.)
     }
 
     function createShareToken(uint8 _outcome) private returns (IShareToken) {
@@ -246,7 +236,6 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
         return true;
 
         // FIXME: when the market is finalized, we need to add `reportingTokens[finalPayoutDistributionHash].totalSupply()` to the reporting window.  This is necessary for fee collection which is a cross-market operation.
-        // TODO: figure out how to make it so fee distribution is delayed until all markets have been finalized; we can enforce it contract side and let the UI deal with the actual work
         // FIXME: if finalPayoutDistributionHash != getIdentityDistributionId(), pay back validity bond holder
         // FIXME: if finalPayoutDistributionHash == getIdentityDistributionId(), transfer validity bond to reportingWindow (reporter fee pot)
         // FIXME: if designated report is wrong, transfer designated report bond to reportingWindow
--- a/source/contracts/reporting/Market.sol
+++ b/source/contracts/reporting/Market.sol
@@ -240,7 +240,7 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
         return true;
 
         // TODO: create a floating designated report bond and charge it for designated reports  
-        // TODO: if designated report is wrong, transfer designated report bond to reportingWindow
+        // TODO: if designated report is wrong, transfer designated report bond to the winning reporting token
         // TODO: if designated report is right, transfer designated report bond to market creator
     }
 
@@ -315,7 +315,7 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
 
     function doFeePayout(bool _toOwner, uint256 _amount) private returns (bool) {
         if (_toOwner) {
-            getOwner().transfer(_amount);
+            getOwner().value(_amount)();
         } else {
             getReportingWindow().receiveValidityBond.value(_amount)();
         }
--- a/source/contracts/reporting/Market.sol
+++ b/source/contracts/reporting/Market.sol
@@ -315,7 +315,7 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
 
     function doFeePayout(bool _toOwner, uint256 _amount) private returns (bool) {
         if (_toOwner) {
-            getOwner().value(_amount)();
+            getOwner().transfer(_amount);
         } else {
             getReportingWindow().receiveValidityBond.value(_amount)();
         }
--- a/source/contracts/reporting/Market.sol
+++ b/source/contracts/reporting/Market.sol
@@ -43,6 +43,7 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
     IShareToken[] private shareTokens;
     uint256 private finalizationTime;
     uint256 private designatedReportReceivedTime;
+    bytes32 private designatedReportPayoutHash;
     bytes32 private tentativeWinningPayoutDistributionHash;
     // We keep track of the second place winning payout hash since when a dispute bond is placed it counts negatively toward stake and we can't otherwise figure out which outcome to promote. Since we only store two hashes it may be the case that if promotion occurs this value is not actually second place, but there is only one case where promotion occurs in a market's lifetime, so it will no longer be relevant at that point.
     bytes32 private bestGuessSecondPlaceTentativeWinningPayoutDistributionHash;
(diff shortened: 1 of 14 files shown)


source/contracts/reporting/Market.sol
```diff
@@ -63,17 +64,18 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
 
     function initialize(IReportingWindow _reportingWindow, uint256 _endTime, uint8 _numOutcomes, uint256 _numTicks, uint256 _feePerEthInAttoeth, ICash _cash, address _creator, address _designatedReporterAddress) public payable beforeInitialized returns (bool _success) {
         endInitialization();
-        require(address(_reportingWindow) != NULL_ADDRESS);
-        require(_numOutcomes >= 2);
-        require(_numOutcomes <= 8);
+        require(_numOutcomes >= 2 && _numOutcomes <= 8);
```
MicahZoltu: Style: For bounded checks like this I prefer to have it look like:
```
require(2 <= _numOutcomes && _numOutcomes <= 8)
```
That way it looks kind of like `2 <= numOutcomes <= 8`.

source/contracts/reporting/Market.sol
```diff
@@ -63,17 +64,18 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
...
         require(_creator != NULL_ADDRESS);
         require(_cash.getTypeName() == "Cash");
-        // FIXME: require market to be on a non-forking universe; repeat this check up the stack as well if necessary (e.g., in reporting window)
-        // CONSIDER: should we allow creator to send extra ETH, is there risk of variability in bond requirements?
-        require(msg.value == MarketFeeCalculator(controller.lookup("MarketFeeCalculator")).getMarketCreationCost(_reportingWindow));
         reportingWindow = _reportingWindow;
+        require(address(getForkingMarket()) == NULL_ADDRESS);
+        // CONSIDER: should we allow creator to send extra ETH, is there risk of variability in bond requirements?
+        MarketFeeCalculator _marketFeeCaluclator = MarketFeeCalculator(controller.lookup("MarketFeeCalculator"));
+        reporterCasCostsFeeAttoeth = _marketFeeCaluclator.getTargetReporterGasCosts(_reportingWindow);
```
MicahZoltu: Typo:
`reporterCasCosts`
`reporterGasCosts`

nuevoalex: good catch!

source/contracts/reporting/Market.sol
```diff
@@ -63,17 +64,18 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
...
         require((_numTicks.isMultipleOf(_numOutcomes)));
         require(feePerEthInAttoeth <= MAX_FEE_PER_ETH_IN_ATTOETH);
         require(_creator != NULL_ADDRESS);
         require(_cash.getTypeName() == "Cash");
-        // FIXME: require market to be on a non-forking universe; repeat this check up the stack as well if necessary (e.g., in reporting window)
-        // CONSIDER: should we allow creator to send extra ETH, is there risk of variability in bond requirements?
-        require(msg.value == MarketFeeCalculator(controller.lookup("MarketFeeCalculator")).getMarketCreationCost(_reportingWindow));
         reportingWindow = _reportingWindow;
+        require(address(getForkingMarket()) == NULL_ADDRESS);
+        // CONSIDER: should we allow creator to send extra ETH, is there risk of variability in bond requirements?
```
MicahZoltu: I believe the answer to this CONSIDER is now "yes".  Depending on when the transaction is mined relative to when it was created, the fees may change so we should support the creator sending excess ETH and refunding unused at the end of the transaction.

source/contracts/reporting/Market.sol
```diff
@@ -157,11 +149,11 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
         reportingWindow.getReputationToken().trustedTransfer(msg.sender, allReportersDisputeBondToken, Reporting.allReportersDisputeBondAmount());
         reportingWindow.getUniverse().fork();
         IReportingWindow _newReportingWindow = getUniverse().getReportingWindowForForkEndTime();
-        updateTentativeWinningPayoutDistributionHash(tentativeWinningPayoutDistributionHash);
         return migrateReportingWindow(_newReportingWindow);
     }
 
     function migrateReportingWindow(IReportingWindow _newReportingWindow) private afterInitialized returns (bool) {
+        updateTentativeWinningPayoutDistributionHash(tentativeWinningPayoutDistributionHash);
```
MicahZoltu: I know you probably did this to address the code size issue, but it makes me sad none the less.

MicahZoltu: Actually, would you mind dropping a comment here to move it back out at some point.  `migrateReportingWindow` really shouldn't have a side effect of updating payouts and I'm worried we'll forget it here.

nuevoalex: Yea this one made me sad too :[ I'll actually a comment about it so if the refactor effort succeeds we can put it where it belongs

source/contracts/reporting/Market.sol
```diff
@@ -233,24 +225,23 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
...
+        if (getForkingMarket() == this) {
             tentativeWinningPayoutDistributionHash = getWinningPayoutDistributionHashFromFork();
         }
 
-        require(tentativeWinningPayoutDistributionHash != bytes32(0));
-
         finalPayoutDistributionHash = tentativeWinningPayoutDistributionHash;
         finalizationTime = block.timestamp;
         transferIncorrectDisputeBondsToWinningReportingToken();
+        // The validity bond is paid to the owner in any indeterminate outcome and the reporting window otherwise
```
MicahZoltu: This comment is backward.  Owner gets the bond if the final outcome is valid, reporting window gets it if the final outcome is invalid.

nuevoalex: Ah yup my bad

source/contracts/reporting/Market.sol
```diff
@@ -322,10 +313,19 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
         return true;
     }
 
+    function doFeePayout(bool _toOwner, uint256 _amount) private returns (bool) {
+        if (_toOwner) {
+            getOwner().transfer(_amount);
+        } else {
+            // We can't do a simple transfer here since the reporting window is a delegated contract and the fallback function will fail on a standard send/transfer
```
MicahZoltu: In general we should never be using `.send` or `.transfer` because it doesn't supply enough gas to contracts.  If that isn't the case, we should fix it.  If that is the case, I'm not sure if this comment is necessary.

MicahZoltu: Note: If you find the comment valuable then I am not opposed to leaving it.

nuevoalex: I'll remove the comment. We aren't `Sending` or `Transfering` to contracts anywhere else to my knowledge.

source/contracts/reporting/Market.sol
```diff
@@ -233,24 +225,23 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
...
             tentativeWinningPayoutDistributionHash = getWinningPayoutDistributionHashFromFork();
         }
 
-        require(tentativeWinningPayoutDistributionHash != bytes32(0));
-
         finalPayoutDistributionHash = tentativeWinningPayoutDistributionHash;
         finalizationTime = block.timestamp;
         transferIncorrectDisputeBondsToWinningReportingToken();
+        // The validity bond is paid to the owner in any indeterminate outcome and the reporting window otherwise
+        doFeePayout(!isIndeterminate(), validityBondAttoeth);
```
MicahZoltu: Wording: I believe we have officially decided to call it `valid` and `invalid` and strike `indeterminate` from the record.  Feel free to address naming in a separate PR, but we should try to get some consistency on that in the long run.

MicahZoltu: Also, I would prefer a function named `isValid` rather than `isInvalid` so we can avoid double negatives like this.

nuevoalex: I'll add it here. I was fighting calling it invalid the whole time anyway so that'll scratch an itch.

source/contracts/reporting/Market.sol
```diff
@@ -233,24 +225,23 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
...
         return true;
 
-        // FIXME: when the market is finalized, we need to add `reportingTokens[finalPayoutDistributionHash].totalSupply()` to the reporting window.  This is necessary for fee collection which is a cross-market operation.
-        // TODO: figure out how to make it so fee distribution is delayed until all markets have been finalized; we can enforce it contract side and let the UI deal with the actual work
-        // FIXME: if finalPayoutDistributionHash != getIdentityDistributionId(), pay back validity bond holder
-        // FIXME: if finalPayoutDistributionHash == getIdentityDistributionId(), transfer validity bond to reportingWindow (reporter fee pot)
-        // FIXME: if designated report is wrong, transfer designated report bond to reportingWindow
-        // FIXME: if designated report is right, transfer designated report bond to market creator
+        // TODO: create a floating designated report bond and charge it for designated reports  
+        // TODO: if designated report is wrong, transfer designated report bond to reportingWindow
```
MicahZoltu: Designated report bond is in REP, so it doesn't get transferred to the ReportingWindow.  It should be transferred to the winning reporting token I believe (verify by looking to see where dispute bonds go).

source/contracts/reporting/Market.sol
```diff
@@ -322,10 +313,19 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
         return true;
     }
 
+    function doFeePayout(bool _toOwner, uint256 _amount) private returns (bool) {
+        if (_toOwner) {
+            getOwner().transfer(_amount);
```
MicahZoltu: Speaking of not using `.transfer`... ;)  This should be `value(...)()`

nuevoalex: Hows that syntax work in this case exactly?

nuevoalex: If you mean `call.value(...)()` I'd actually argue against that since it's effectively `transfer` but more dangerous.

source/contracts/reporting/Reporting.sol
```diff
@@ -14,7 +14,8 @@ library Reporting {
     uint256 private constant LIMITED_REPORTERS_DISPUTE_BOND_AMOUNT = 11 * 10**21;
     uint256 private constant ALL_REPORTERS_DISPUTE_BOND_AMOUNT = 11 * 10**22;
 
-    uint256 private constant GAS_TO_REPORT = 567796;
+    // NOTE: We need to maintain this cost to roughly match the gas cost of reporting. This was last updated 10/02/2017
+    uint256 private constant GAS_TO_REPORT = 600000;
```
MicahZoltu: Did you actually run this number?  Is it for real?  😢

nuevoalex: Tis real. The target reporter gas cost fee is quite large due to this as you can imagine. :[

nuevoalex: If you look at the test I had in place before btw that is how I was calculating it

source/contracts/reporting/ReportingToken.sol
```diff
@@ -26,7 +26,7 @@ contract ReportingToken is DelegationTarget, Typed, Initializable, VariableSuppl
         require(_market.getNumberOfOutcomes() == _payoutNumerators.length);
         market = _market;
         payoutNumerators = _payoutNumerators;
-        // TODO: call a function on `self.getUniverse()` that logs the creation of this token with an index for the market, function needs to verify that caller is `universe.isContainerForReportingToken(thisToken)`
+        getUniverse().logCreateReportingToken(this, market);
```
MicahZoltu: Recommend against implementing any logs until we know what the UI needs.  Also, anywhere you see a "need to log this" comment feel free to delete.

PR #336, opened 2017-10-04
Handle fee migration during a fork

nuevoalex:
I also changed some code in `Cash` to be consistent with using `.call.value` instead of `transfer` and removed an inaccurate comment.

### DIFF
--- a/source/contracts/reporting/ReportingWindow.sol
+++ b/source/contracts/reporting/ReportingWindow.sol
@@ -84,6 +84,7 @@ contract ReportingWindow is DelegationTarget, Typed, Initializable, IReportingWi
         require(_originalUniverse.isContainerForReportingWindow(_shadyReportingWindow));
         IReportingWindow _originalReportingWindow = _shadyReportingWindow;
         require(_originalReportingWindow.isContainerForMarket(_market));
+        _shadyReportingWindow.migrateFeesDueToFork();
         privateAddMarket(_market);
         return true;
     }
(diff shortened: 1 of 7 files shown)


source/contracts/reporting/ReportingWindow.sol
```diff
@@ -84,6 +84,7 @@ contract ReportingWindow is DelegationTarget, Typed, Initializable, IReportingWi
         require(_originalUniverse.isContainerForReportingWindow(_shadyReportingWindow));
         IReportingWindow _originalReportingWindow = _shadyReportingWindow;
         require(_originalReportingWindow.isContainerForMarket(_market));
+        _shadyReportingWindow.migrateFeesDueToFork();
```
MicahZoltu: Recommend an intermediate variable `IReportingWindow legitReportingWindow = _shadyReportingWindow` on the line after the reporting window is deemed to be verified.

Also, just noticed this code doesn't call `_market` `_shadyMarket` like most of the other functions with this pattern.  Any unvalidated address should be named `_shady*` until it is validated, then renamed (copied) to `_legit*` (or no-prefix) to make it clear when validation has finished.

source/contracts/reporting/ReportingWindow.sol
```diff
@@ -216,6 +217,29 @@ contract ReportingWindow is DelegationTarget, Typed, Initializable, IReportingWi
...
+        // NOTE: Will need to figure out a way to transfer other denominations when that is implemented
+        ICash _cash = ICash(controller.lookup("Cash"));
+        uint256 _balance = _cash.balanceOf(this);
+        if (_balance == 0) {
+            return false;
+        }
+        IReportingWindow _shadyReportingWindow = IReportingWindow(msg.sender);
+        IUniverse _shadyUniverse = _shadyReportingWindow.getUniverse();
+        require(_shadyUniverse.isContainerForReportingWindow(_shadyReportingWindow));
+        require(universe.isParentOf(_shadyUniverse));
```
MicahZoltu: Critical: This needs to have a check that ensures that fees are _only_ migrated to the universe that won the fork.  This means the fork has to be finalized, and the destination universe must be the winner of the fork.

source/contracts/reporting/ReportingWindow.sol
```diff
@@ -216,6 +217,29 @@ contract ReportingWindow is DelegationTarget, Typed, Initializable, IReportingWi
...
+        _cash.transfer(_shadyReportingWindow, _balance);
```
MicahZoltu: Same comment as above, rename `_shady*` to something not shady as soon as we have finished validating to make it clear to the reader when validation is complete and the address can be trusted.

source/contracts/reporting/ReportingWindow.sol
nuevoalex: Will do

source/contracts/reporting/ReportingWindow.sol
```diff
@@ -234,7 +236,11 @@ contract ReportingWindow is DelegationTarget, Typed, Initializable, IReportingWi
         IUniverse _shadyUniverse = _shadyReportingWindow.getUniverse();
         require(_shadyUniverse.isContainerForReportingWindow(_shadyReportingWindow));
         require(universe.isParentOf(_shadyUniverse));
-        _cash.transfer(_shadyReportingWindow, _balance);
+        IUniverse _destinationUniverse = _shadyUniverse;
+        IReportingWindow _destinationReportingWindow = _shadyReportingWindow;
+        bytes32 _winningForkPayoutDistributionHash = universe.getForkingMarket().getFinalPayoutDistributionHash();
+        require(_destinationUniverse == universe.getOrCreateChildUniverse(_winningForkPayoutDistributionHash));
```
MicahZoltu: I think there is a `getChildUniverse` that is constant and throws if the universe doesn't exist.  Should be safe to call that one instead (less gas) since we will never be migrating fees to a universe that hasn't been created yet.

PR #337, opened 2017-10-04
Ut updates for new version pyethereum

bthaile:
update requirements files to point to new version of augur pyethereum.
change to include language for uploading contracts to chain 
updated order of reverting chain to snapshot

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -221,9 +221,6 @@ class ContractsFixture:
         self.chain.revert(captured['snapshot'])
         self.contracts = deepcopy(captured['contracts'])
 
-    def pushTimeStamp(self, setTime):
-        self.chain.head_state.timestamp = setTime
-
     ####
     #### Bulk Operations
     ####
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -192,7 +192,7 @@ class ContractsFixture:
         contractTranslator = ContractTranslator(signature)
         if len(constructorArgs) > 0:
             compiledCode += contractTranslator.encode_constructor_arguments(constructorArgs)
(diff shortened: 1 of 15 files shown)


tests/conftest.py
```diff
@@ -209,17 +211,19 @@ def resetSnapshot(self):
...
         if len(captured) < 4:
             raise "captured snapshot doesn't have all parameters in dictionary, need to call createSnapshot"
         self.chain.block = captured['block']
-        self.chain.revert(captured['snapshot'])
         self.chain.head_state = captured['head_state']
+        self.chain.revert(captured['snapshot'])
         self.contracts = deepcopy(captured['contracts'])
 
+    def pushTimeStamp(self, setTime):
+        self.chain.head_state.timestamp = setTime
```
MicahZoltu: What is the purpose of this function?  This changes call sites from an assignment to a function:
```
fixture.chain.head_state.timestamp += 10
```
```
fixture.pushTimestamp(fixture.chain.head_state.timestamp + 10)
```
The latter feels more awkward.

bthaile: an attempt to control chain changes, the timestamp changes added time to reporting window get end Time, dispute start time. They didn't refer to the chain's timestamp.

MicahZoltu:
Hmm, I thought we had some tests that did `timestamp += x`... Is the idea that we don't want tests touching the `chain` object directly?  Does Python have a mechanism for making it so we can do `fixture.timestamp += x` and have that effect the underlying chain timestamp?

I like the idea of limiting access, though I don't like losing the simple assignment syntax.

bthaile:
I removed the pushTimestamp method. I think helper method that have more meaningful context is in order ie. pushTimestampToEndOfReportingWindow() that would make the unit test code more readable and other other tests can use it.

PR #338, opened 2017-10-05
Designated report rep cost

nuevoalex:
This changes the designated report to be done by buying reporting tokens instead of calling a method on the `Market`.


source/contracts/extensions/MarketFeeCalculator.sol
```diff
@@ -13,11 +13,13 @@ contract MarketFeeCalculator {
...
     mapping (address => uint256) private validityBondInAttoeth;
     mapping (address => uint256) private targetReporterGasCosts;
+    mapping (address => uint256) private designatedReportStakeInAttoRep;
 
-    uint256 private constant DEFAULT_VALIDITY_BOND = 1 ether / 100;
-    uint256 private constant TARGET_INVALID_MARKETS_DIVISOR = 100; // 1% of markets
+    uint256 private constant TARGET_INVALID_MARKETS_DIVISOR = 100; // 1% of markets are expected to be invalid
     uint256 private constant TARGET_REP_MARKET_CAP_MULTIPLIER = 5;
 
+    uint256 private constant TARGET_INCORRECT_MARKETS_DIVISOR = 100; // 1% of markets are expected to have an incorrect designate report
```
MicahZoltu: Consider including `designated` somewhere in the name of this constant since it is specifically for the number of incorrect designated reports relative to total reports, not incorrect markets in general (since there is no such thing as an incorrectly resolved market).

source/contracts/extensions/MarketFeeCalculator.sol
```diff
@@ -13,11 +13,13 @@ contract MarketFeeCalculator {
     mapping (address => uint256) private shareSettlementPerEthFee;
     mapping (address => uint256) private validityBondInAttoeth;
     mapping (address => uint256) private targetReporterGasCosts;
+    mapping (address => uint256) private designatedReportStakeInAttoRep;
```
MicahZoltu: All of these should probably be moved to be stored in a non-extension contract like Universe.  The values should _not_ carry across universes, and things in Extensions are meant to be "helpers", not storing critical pieces of the system.  We may want to consider moving this contract out of extensions and into the core.  If we do keep data in here (it still needs to be indexed by Universe) we should probably delegate it.

My recommendation is to move this contract into the core suite of contracts and get these storage items moved over to Universe.

MicahZoltu: Feel free to create a clubhouse issue and do the above in a separate PR.

nuevoalex: Will do! I had similar thoughts about this contract since it's storing/writing data and would need to be audited.

source/contracts/extensions/MarketFeeCalculator.sol
```diff
@@ -28,41 +30,44 @@ contract MarketFeeCalculator {
...
+        uint256 _totalMarketsInPreviousWindow = _reportingWindow.getNumMarkets();
+        uint256 _incorrectDesignatedReportMarketsInPreviousWindow = _reportingWindow.getNumIncorrectDesignatedReportMarkets();
+        uint256 _previousDesignatedReportStakeInAttoRep = designatedReportStakeInAttoRep[_previousReportingWindow];
+
+        _currentDesignatedReportStakeInAttoRep = calculateFloatingValue(_incorrectDesignatedReportMarketsInPreviousWindow, _totalMarketsInPreviousWindow, TARGET_INCORRECT_DESIGNATED_REPORT_MARKETS_DIVISOR, _previousDesignatedReportStakeInAttoRep, Reporting.defaultDesignatedReportStake());
+        designatedReportStakeInAttoRep[_reportingWindow] = _currentDesignatedReportStakeInAttoRep;
+        return _currentDesignatedReportStakeInAttoRep;
+    }
+
+    function calculateFloatingValue(uint256 _invalidMarkets, uint256 _totalMarkets, uint256 _targetDivisor, uint256 _previousValue, uint256 _defaultValue) constant public returns (uint256 _newValue) {
```
MicahZoltu: We may want to rename `invalidMarkets` to something else since `invalid` has very specific meaning in our contracts.  Perhaps `badMarkets`?  (open to suggestions, not a huge fan of `badMarkets`)

source/contracts/reporting/Market.sol
```diff
@@ -67,10 +67,10 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
         require(2 <= _numOutcomes && _numOutcomes <= 8);
         require((_numTicks.isMultipleOf(_numOutcomes)));
         require(feePerEthInAttoeth <= MAX_FEE_PER_ETH_IN_ATTOETH);
-        require(_creator != NULL_ADDRESS);
-        require(_cash.getTypeName() == "Cash");
+        // TODO: require(_creator != NULL_ADDRESS);
+        // TODO: require(_cash.getTypeName() == "Cash");
         reportingWindow = _reportingWindow;
-        require(address(getForkingMarket()) == NULL_ADDRESS);
+        // TODO: require(address(getForkingMarket()) == NULL_ADDRESS);
```
MicahZoltu: I believe we can uncomment these now.

source/contracts/reporting/Market.sol
```diff
@@ -112,14 +112,14 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
...
 
-    function designatedReport(uint256[] _payoutNumerators) public returns (bool) {
-        // intentionally does not migrate the market as designated report markets won't actually migrate unless a dispute bond has been placed or the designated report doesn't occur
-        require(msg.sender == designatedReporterAddress);
-        require(getReportingState() == ReportingState.DESIGNATED_REPORTING);
-        // we have to create the reporting token so the rest of the system works (winning reporting token must exist)
-        getReportingToken(_payoutNumerators);
+    function designatedReport() public returns (bool) {
+        // Intentionally does not migrate the market as designated report markets won't actually migrate unless a dispute bond has been placed or the designated report doesn't occur
+        // CONSIDER: We are currently implicitly trusting the rporting token to call this at the appropriate state. We may want to add more validation here when the Market contract can afford the extra lines
```
MicahZoltu: Recommend following up on this `CONSIDER` now that we have some breathing room in Market.sol.

source/contracts/reporting/Market.sol
```diff
@@ -387,6 +391,10 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
         return finalPayoutDistributionHash;
     }
 
+    function getDesignatedReportPayoutHash() public constant returns (bytes32) {
+        return designatedReportPayoutHash;
+    }
```
MicahZoltu: Unrelated to this PR, I wonder if manual getters are still fewer bytes than auto-getters with optimizations enabled?

source/contracts/reporting/Market.sol
```diff
@@ -112,14 +112,14 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
...
     }
 
-    function designatedReport(uint256[] _payoutNumerators) public returns (bool) {
-        // intentionally does not migrate the market as designated report markets won't actually migrate unless a dispute bond has been placed or the designated report doesn't occur
-        require(msg.sender == designatedReporterAddress);
-        require(getReportingState() == ReportingState.DESIGNATED_REPORTING);
-        // we have to create the reporting token so the rest of the system works (winning reporting token must exist)
-        getReportingToken(_payoutNumerators);
+    function designatedReport() public returns (bool) {
+        // Intentionally does not migrate the market as designated report markets won't actually migrate unless a dispute bond has been placed or the designated report doesn't occur
```
MicahZoltu: Not a bug in this PR, but this comment is no longer accurate, all non-finalized markets should migrate, even ones with designated reporters and markets that are pre-reporting.

source/contracts/reporting/ReportingWindow.sol
```diff
@@ -131,10 +135,15 @@ contract ReportingWindow is DelegationTarget, Typed, Initializable, IReportingWi
         require(markets.contains(_market));
         require(_market.getReportingTokenOrZeroByPayoutDistributionHash(_payoutDistributionHash) == msg.sender);
         IMarket.ReportingState _state = _market.getReportingState();
-        require(_state == IMarket.ReportingState.ALL_REPORTING || _state == IMarket.ReportingState.LIMITED_REPORTING);
+        bool _validReportingState = _state == IMarket.ReportingState.ALL_REPORTING;
+        _validReportingState = _validReportingState || (_state == IMarket.ReportingState.LIMITED_REPORTING);
+        _validReportingState = _validReportingState || (_state == IMarket.ReportingState.DESIGNATED_REPORTING);
```
MicahZoltu: Style: Curious why you split these across multiple statements instead of something like:
```
require(_state == IMarket.ReportingState.ALL_REPORTING
    || _state == IMarket.ReportingState.LIMITED_REPORTING
    || _state == IMarket.ReportingState.DESIGNATED_REPORTING);
```

source/contracts/reporting/Market.sol
nuevoalex: Will experiment at some point with that, although we're well within size limits now too

source/contracts/reporting/ReportingWindow.sol
nuevoalex: Not a good reason. This is much better!

PR #340, opened 2017-10-05
Market contract size reductions. These will probably need optimization later

nuevoalex:
At some point it may be less expensive to pass everything needed to the extension methods instead of having them do multiple calls back into market to get out needed data.


source/contracts/reporting/IMarket.sol
```diff
@@ -38,13 +38,20 @@ contract IMarket is Typed, IOwnable {
...
     function getFinalWinningReportingToken() public constant returns (IReportingToken);
     function getReportingTokenOrZeroByPayoutDistributionHash(bytes32 _payoutDistributionHash) public constant returns (IReportingToken);
-    function getPayoutDistributionHashStake(bytes32 _payoutDistributionHash) public constant returns (int256);
+    function getTentativeWinningPayoutDistributionHash() public constant returns (bytes32);
+    function getBestGuessSecondPlaceTentativeWinningPayoutDistributionHash() public constant returns (bytes32);
+    function getForkingMarket() public constant returns (IMarket _market);
+    function getEndTime() public constant returns (uint256);
+    function getDesignatedReportDueTimestamp() public constant returns (uint256);
+    function getDesignatedReportReceivedTime() public constant returns (uint256);
+    function getDesignatedReportDisputeDueTimestamp() public constant returns (uint256);
```
MicahZoltu: The fact that these weren't in the interface suggests they weren't being called externally.  Instead of adding them to the interface, consider making them private.  The exception to this is getters that are just blindly returning some contract state, which we do want to make available to the general public.  Over time I suspect the set of public methods will grow as the UI asks us for access to things, but I would prefer to start small and expand rather than starting with everything.

tests/conftest.py
```diff
@@ -85,6 +93,13 @@ def getCompiledCode(relativeFilePath):
             print('using cached compilation for ' + name)
         with io_open(compiledOutputPath, mode='rb') as file:
             compiledCode = file.read()
+            contractSize = len(compiledCode)
+            if (contractSize >= CONTRACT_SIZE_LIMIT):
+                print('%sContract %s is OVER the size limit by %d bytes%s' % (bcolors.FAIL, name, contractSize - CONTRACT_SIZE_LIMIT, bcolors.ENDC))
+            elif (contractSize >= CONTRACT_SIZE_WARN_LEVEL):
+                print('%sContract %s is under size limit by only %d bytes%s' % (bcolors.WARN, name, CONTRACT_SIZE_LIMIT - contractSize, bcolors.ENDC))
+            elif (contractSize > 0):
+                print('Size: %i' % contractSize)
```
MicahZoltu: Ooh, snazzy.  This will be nice.

source/contracts/reporting/IMarket.sol
nuevoalex: I added them to the interface so that the `MarketExtensions` contract could call them. This could be optimized out at some point probably by passing this data to those functions, although that will be a battle with stack size limit I'm guessing.

PR #342, opened 2017-10-06
Miscellaneous fixups.

MicahZoltu:
Nothing major, just things I ran across while debugging some things on a branch.

### DIFF
--- a/source/contracts/reporting/Market.sol
+++ b/source/contracts/reporting/Market.sol
@@ -29,8 +29,8 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
     uint256 private numTicks;
     uint256 private feePerEthInAttoeth;
 
-    uint256 private constant MAX_FEE_PER_ETH_IN_ATTOETH = 5 * 10 ** 17;
-    uint256 private constant APPROVAL_AMOUNT = 2 ** 254;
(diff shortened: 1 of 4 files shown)


source/contracts/reporting/Market.sol
```diff
@@ -29,8 +29,8 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
     uint256 private numTicks;
     uint256 private feePerEthInAttoeth;
 
-    uint256 private constant MAX_FEE_PER_ETH_IN_ATTOETH = 5 * 10 ** 17;
-    uint256 private constant APPROVAL_AMOUNT = 2 ** 254;
+    uint256 private constant MAX_FEE_PER_ETH_IN_ATTOETH = 1 ether / 2;
+    uint256 private constant APPROVAL_AMOUNT = 2**256-1;
```
adrake33: What's the reason for changing this?  Was this value incorrect, or did we just decide to change this value?

MicahZoltu: The first one is just a simplification, the value doesn't actually change.  The second one was wrong since we changed `2**256-1` to be a magic number that means "infinite" and costs less gas to use.  This was a recent change (making `2**256-1` be magic) and we just missed changing this to match.

PR #343, opened 2017-10-06
Reporter fee payout

nuevoalex:


source/contracts/reporting/ReportingWindow.sol
```diff
@@ -117,16 +118,21 @@ contract ReportingWindow is DelegationTarget, Typed, Initializable, IReportingWi
         }
 
         if (_state == IMarket.ReportingState.FINALIZED) {
-            if (!_market.isValid()) {
-                invalidMarketCount++;
-            }
-            marketReports.record(numberOfReportsByMarket[_market]);
-            totalWinningReportingTokens = totalWinningReportingTokens.add(_market.getFinalWinningReportingToken().totalSupply());
+            updateFinalizedMarket(_market);
```
MicahZoltu: This should only be called if the market was previously not finalized and is now finalized.  I _suspect_ that this is probably only called in such a scenario, but I would be a lot more comfortable if we had some strong guarantees in the form of `require(...)` statements that ensured this method can only ever be called once per market.  If it is called multiple times for a market, we'll over-count a few things, some of which could be significant.

source/contracts/reporting/ReportingWindow.sol
```diff
@@ -209,6 +215,10 @@ contract ReportingWindow is DelegationTarget, Typed, Initializable, IReportingWi
         return getUniverse().getReportingWindowByTimestamp(_previousTimestamp);
     }
 
+    function allMarketsFinalized() constant public returns (bool) {
+        return markets.count == finalizedMarketCount;
```
MicahZoltu: We don't have any strong invariants that these will eventually equal each other.  As mentioned above, it would be nice if we could come up with some mechanisms that provide us strong guarantees without the reader/auditor having to read/understand everything just to assert that all reporting windows will eventually land in this state.  I don't have any great solution at the moment, though perhaps we can come up with some kind of specialized container library that gives us stronger guarantees that markets eventually end up in this state?

source/contracts/reporting/ReportingWindow.sol
```diff
@@ -218,6 +228,20 @@ contract ReportingWindow is DelegationTarget, Typed, Initializable, IReportingWi
         return true;
     }
 
+    function collectReportingFees(address _reporterAddress, uint256 _attoReportingTokens) public returns (bool) {
+        IReportingToken _shadyReportingToken = IReportingToken(msg.sender);
+        require(isContainerForReportingToken(_shadyReportingToken));
```
MicahZoltu: Style: I would be OK with inlining the cast in this case: `require(isContainerForReportingToken(IReportingToken(msg.sender)))`.  It is on the boundary of acceptable to me, but the line is so simple that I'm OK with it.  If you prefer the current two-statement style though I'm fine with leaving it as well.

source/contracts/reporting/ReportingWindow.sol
```diff
@@ -218,6 +228,20 @@ contract ReportingWindow is DelegationTarget, Typed, Initializable, IReportingWi
...
+        // NOTE: Will need to handle other denominations when that is implemented
+        ICash _cash = ICash(controller.lookup("Cash"));
+        uint256 _balance = _cash.balanceOf(this);
+        uint256 _feePayoutShare = _balance.mul(_attoReportingTokens).div(totalWinningReportingTokens);
+        if (_feePayoutShare > 0) {
+            _cash.withdrawEtherTo(_reporterAddress, _feePayoutShare);
+        }
+        totalWinningReportingTokens = totalWinningReportingTokens.sub(_attoReportingTokens);
```
MicahZoltu: Security: Always update state _before_ calling out to external contracts (`.withdrawEtherTo` will call an external contract).  At the moment, this is subject to a reentrancy attack.

source/contracts/reporting/ReportingToken.sol
```diff
@@ -63,29 +63,31 @@ contract ReportingToken is DelegationTarget, Typed, Initializable, VariableSuppl
...
-        var _reputationSupply = _reputationToken.balanceOf(this);
-        var _attotokens = balances[msg.sender];
-        var _reporterReputationShare = _reputationSupply * _attotokens / supply;
+        IReportingWindow _reportingWindow = market.getReportingWindow();
+        require(_reportingWindow.allMarketsFinalized());
+        IReputationToken _reputationToken = getReputationToken();
+        uint256 _reputationSupply = _reputationToken.balanceOf(this);
+        uint256 _attotokens = balances[msg.sender];
+        uint256 _reporterReputationShare = _reputationSupply * _attotokens / supply;
+        _reportingWindow.collectReportingFees(msg.sender, _attotokens);
```
MicahZoltu: Security: Since this will call out to an external contract, it should be the last line of this function, after all other state updates are complete.  This should go even after the reputationToken transfer.

source/contracts/reporting/ReportingWindow.sol
```diff
@@ -327,6 +351,16 @@ contract ReportingWindow is DelegationTarget, Typed, Initializable, IReportingWi
         return registrationToken == _shadyRegistrationToken;
     }
 
+    function isContainerForReportingToken(IReportingToken _shadyReportingToken) public afterInitialized constant returns (bool) {
```
MicahZoltu: Elsewhere I changed a bunch of these to take in a `Typed` as a parameter, then after type checking I cast to the expected type.  It looks like you already have the type check, but since this is taking in an `IReportingToken` it means the caller must already know the correct type.  Consider changing this to take in a `Typed` like the others (I think Universe is where I updated them all).  This change is really a callsite convenience thing, feel free to not do it if you think it hurts more than helps.  My only argument is consistence.

Same note for the others in this file.

source/contracts/reporting/ReportingWindow.sol
nuevoalex: What would a container library like that buy us exactly? Is the worry that a market will finalize and incorrectly not update the count?

source/contracts/reporting/ReportingWindow.sol
nuevoalex: Good catches. Will do

source/contracts/reporting/ReportingWindow.sol
nuevoalex: Will change to Typed

source/contracts/reporting/ReportingWindow.sol
MicahZoltu: I'm not sure exactly what it would look like.  I just always am worried anytime there is something as large/complex as Market and it doesn't have strong invariants at every step.  I don't have a specific bug that I know of, but my gut tells me that there is _probably_ a bug somewhere in market that could result in a market not being accounted for properly.

In general, the level of complexity in market makes me uncomfortable, and denormalized data like this without strong invariants is a big source of that.

nuevoalex: Yea it's a complex state machine so I get the concern.

joeykrug: As a note, if you think of assertions / invariants that you think should always be true in the code in certain places, add them in as `assert`s, then we can statically analyze things and actually prove that they're always true, or if not, find when they're _not_ true.

PR #344, opened 2017-10-06
Upgrades

nuevoalex:
Upgrades for solium and solc


package.json
```diff
@@ -18,8 +18,8 @@
         "mocha": "^3.5.3",
         "path": "^0.12.7",
         "recursive-readdir": "^2.2.1",
-        "solc": "^0.4.15",
-        "solium": "github:AugurProject/solium",
+        "solc": "^0.4.17",
```
MicahZoltu: To avoid problems in the future, we should remove the `^` on this.

source/contracts/reporting/Market.sol
```diff
@@ -265,7 +265,8 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
 
     function doFeePayout(bool _toOwner, uint256 _amount) private returns (bool) {
         if (_toOwner) {
-            require(getOwner().call.value(_amount)());
+            bool _success = getOwner().call.value(_amount)();
+            require(_success);
```
MicahZoltu: Style: I would just do `require(getOwner().call.value(_amount)())` for all of these.

Though, in this case it looks like you already had that.  What made you decide to split this into two lines?

nuevoalex: Solium complaints

PR #345, opened 2017-10-06
optimize

nuevoalex:


tests/conftest.py
```diff
@@ -119,6 +119,10 @@ def compileSolidity(relativeFilePath):
             'settings': {
                 # TODO: Remove 'remappings' line below and update 'sources' line above
                 'remappings': [ '=%s/' % resolveRelativePath("../source/contracts") ],
+                'optimizer': {
+                    'enabled': True,
+                    'runs': 500
+                },
```
MicahZoltu: We'll need the same change in the TS code in `CompileSolidity.ts` line 85 or so.

PR #346, opened 2017-10-07
added a few reporting tests for designated reporter

bthaile:
Added tests for designated reporter. A bit confused on cases where DR splits coins 
example: market.designatedReport([10**9,10**9], sender=tester.k0). I'll ask for clarity in #dev


MicahZoltu:
Looks like there are some build failures.  Ping me in #dev if you run into trouble getting them resolved.

PR #348, opened 2017-10-07
Adds custom contract caller.

MicahZoltu:
This is very similar to the ethjs contract mechanism, but it allows the caller to override things like value, sender and whether the call should be done as a constant call or not.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -8,8 +8,7 @@
             "program": "${workspaceRoot}/node_modules/mocha/bin/_mocha",
             "cwd": "${workspaceRoot}",
             "args": [
-                "-r", "ts-node/register",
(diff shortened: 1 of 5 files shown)


.vscode/launch.json
```diff
@@ -8,8 +8,7 @@
             "program": "${workspaceRoot}/node_modules/mocha/bin/_mocha",
             "cwd": "${workspaceRoot}",
             "args": [
-                "-r", "ts-node/register",
-                "source/tests-unit/**/*.ts",
+                "output/tests-unit/**/*.js",
```
adrake33: What was the reason for changing this?  Won't this tell us the line in the JS file that is failing instead of the line in the TS file?

PR #349, opened 2017-10-08
Test infrastructure improvements.

MicahZoltu:
Fixtures are now built up in a hierarchy.  This way, tests that don't require a lot of setup can run quickly by using a simple base fixture, while tests that require the kitchen sink can start with that as their base.  When running individual tests, this can greatly improve test turn around time if the tests you are iterating on are based on the base (only need minimal setup).  See `test_controlled.py` for an example, the tests run in under a second (including test setup time).

Most existing tests just use the kitchen sink snapshot which contains everything that `contractsFixture` used to contain and an alias has been created for it as contractsFixture and sessionFixture, though over time we should preference better names and more targeted fixtures.

In the process, I fixed the Controller tests to no longer test the implementation details of Controlled (what actually started me down this path) and added tests for Controlled.

Removed lazy_fixture, since it turns out we don't actually need it.  When I added it I didn't know how to have fixtures as part of parameterized tests.  I now do (someone else figured it out and I copied).

In general, I recommend following the patterns found in `test_controlled.py` rather than the pre-existing patterns.

### DIFF
--- a/tests/test_controller.py
+++ b/tests/test_controller.py
@@ -2,7 +2,7 @@
 
 from ethereum.tools import tester
 from ethereum.tools.tester import TransactionFailed
-from pytest import raises, fixture, mark
+from pytest import raises, fixture
 from utils import longToHexString, bytesToHexString
 
 def test_whitelists(controller):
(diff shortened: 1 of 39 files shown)


MicahZoltu:
Note: This is currently failing the registration tests.  If anyone is feeling magnanimous feel free to try to debug them.  The problem is that we can't take a proper snapshot without first mining a block, and mining a block resets the timestamp.  These tests want to take a snapshot after the timestamp has been changed but the way we are doing timestamps makes this not work.

tests/test_controller.py
```diff
@@ -32,16 +32,13 @@ def test_registry(controller, decentralizedController):
     with raises(TransactionFailed): decentralizedController.assertOnlySpecifiedCaller(tester.a2, key2, sender = tester.k0)
     with raises(TransactionFailed): controller.assertOnlySpecifiedCaller(tester.a2, key2, sender = tester.k2)
 
-def test_suicideTokens(contractsFixture, controller, decentralizedController, controllerUser):
-    legacyRepContract = contractsFixture.contracts['LegacyRepContract']
-    legacyRepContract.faucet(0)
-    legacyRepContract.transfer(controllerUser.address, 40 * 10**18)
-    with raises(TransactionFailed): controller.suicideFunds(controllerUser.address, tester.a0, [legacyRepContract.address], sender = tester.k2)
+def test_suicide(contractsFixture, controller, decentralizedController, controllerUser):
```
MicahZoltu: @joeykrug I now know why these tests were passing before and why it was correct.  These tests are meant to test `Controller`, not `Controlled`.  I have added some tests for testing `Controlled` and adjusted these tests to focus on just `Controller`.  The test solidity file intentionally implements `IControlled` rather than `Controlled` because it is meant to act as a mock for testing, not as a legitimate `Controlled`.  Meanwhile, over in `Controlled`, you'll notice that the tests don't depend on `Controller` and instead just `IController`.

PR #350, opened 2017-10-08
Fixed all the incorrect constant usages, added pure/view appropriate usage

joeykrug:
This also preps for solidity 0.5 support :)


requirements.txt
```diff
@@ -7,7 +7,7 @@
 asn1crypto==0.23.0        # via coincurve
 cffi==1.11.1              # via coincurve
 coincurve==6.0.0          # via ethereum-augur-temp
-ethereum-augur-temp==2.1.4
+-e git://github.com/AugurProject/pyethereum.git#egg=ethereum-augur-temp
```
MicahZoltu: There should be a change in `requirements.in` as well I would assume?

joeykrug: Just reverted this in a commit, was a mistake to commit it

PR #351, opened 2017-10-09
Fixes market creation.

MicahZoltu:
Also applies some minor cleanup.


source/libraries/ContractDeployer.ts
```diff
@@ -294,48 +294,23 @@ export class ContractDeployer {
     }
 
     private async createCategoricalMarket(universe, numOutcomes, endTime, feePerEthInWei, denominationToken, designatedReporterAddress, numTicks): Promise<ContractBlockchainData> {
+        const constant = { constant: true };
```
adrake33: Good idea -- this is cleaner and reusable.

PR #353, opened 2017-10-09
added tests to gain more test coverage on universe sol

bthaile:
added test for universe to get more code coverage and get a handle on the logic.

### DIFF
--- a/tests/reporting/test_universe.py
+++ b/tests/reporting/test_universe.py
@@ -5,7 +5,6 @@ from pytest import fixture
 def test_universe_creation(universeFixture):
     universe = universeFixture.createUniverse(3, "5")
     reputationToken = universeFixture.applySignature('ReputationToken', universe.getReputationToken())
-
     assert universe.getParentUniverse() == longToHexString(3)
     assert universe.getParentPayoutDistributionHash() == stringToBytes("5")
     assert universe.getForkingMarket() == longToHexString(0)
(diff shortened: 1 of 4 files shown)


tests/reporting/test_universe.py
```diff
@@ -10,13 +11,139 @@ def test_universe_creation(contractsFixture):
...
     assert reputationToken.getTopMigrationDestination() == longToHexString(0)
-
-def test_universe_name(contractsFixture):
-    universe = contractsFixture.createUniverse(3, "5")
     assert universe.getTypeName() == stringToBytes('Universe')
+    assert universe.getForkEndTime() == 0
+    assert universe.getChildUniverse("5") == longToHexString(0)
+
+def test_get_reporting_window(contractsFixture):
+    universe = contractsFixture.createUniverse(1, "10")
```
MicahZoltu: Consider using a snapshot/fixture for this instead of creating a new universe for each test.  Check out the new fixture stuff I added, or ask me about it in #dev for details.  This will let you have tests that run notably faster because they aren't spinning up the entire world every time and instead spin it all up once and then take a snapshot, then reset for each test.  Resetting a snapshot is very cheap compared to actually doing on-chain work.

bthaile: Alright, I’ll make the change to add a new fixture to increase test run speed.

PR #354, opened 2017-10-09
Integration Test

adrake33:
Changes made:

- Added `source/libraries/AbiParser.ts` (similar to ethjs-contract, but allows the caller to override value, sender & whether the call should be done as a constant call).

- Updated `source/libraries/ContractDeployer.ts` to make use of ABIParser in `createCategoricalMarket`.

- Created `source/tests-integration/BinaryMarketTradeAndReport.ts` integration test (not finished yet).

- Created `source/libraries/HelperFunctions.ts` for several helper functions, some of which were in `source/deployment/deployContracts.ts`, and some of which are new.


source/contracts/extensions/MarketCreation.sol
```diff
@@ -9,7 +9,7 @@ import 'reporting/IReportingWindow.sol';
 
 contract MarketCreation {
     function createMarket(IUniverse _universe, uint256 _endTime, uint8 _numOutcomes, uint256 _feePerEthInWei, ICash _denominationToken, uint256 _numTicks, address _designatedReporterAddress) payable public returns (IMarket) {
-        IReportingWindow _reportingWindow = _universe.getReportingWindowByMarketEndTime(_endTime, (_designatedReporterAddress != 0));
+        IReportingWindow _reportingWindow = _universe.getReportingWindowByMarketEndTime(_endTime, true);
```
MicahZoltu: This shouldn't have changed.  This was my bad, I think I accidentally left it in when I PRed into your branch but it shouldn't be here. `_designatedReporterAddress != 0` is the correct value for this.

source/libraries/ContractDeployer.ts
```diff
@@ -2,48 +2,117 @@
...
+        // this.constants = await this.upload("../tests/solidity_test_helpers/Constants.sol");
 
         return true;
     }
 
-    public async uploadAndAddDelegatedToController(contractFileName: string, contractName: string): Promise<ContractBlockchainData|undefined> {
+    // Helper functions
+
+    public async parseBlockTimestamp(blockTimestamp): Promise<Date> {
+        const timestampHex = "0x" + JSON.stringify(blockTimestamp).replace(/\"/g, "");
```
MicahZoltu: Recommend using template strings rather than concatenation:
```
`0x${JSON.stringify(blockTimestamp).replace(/\"/g, "")}`;
```

Not a big deal in this case, but I find it a generally good pattern to follow.

MicahZoltu:
Recommend fixing the first comment before merging, but if not I can fix it after it is merged since it was my fault.

MicahZoltu:
Also recommend doing a squash merge when the time comes, since this branch has a _lot_ of commits.

source/contracts/extensions/MarketCreation.sol
adrake33: Yeah, I wasn't sure why that had changed and was meaning to ask you about it.  I'll change it back.

source/libraries/ContractDeployer.ts
adrake33: Ah, good idea.  Will change that as well.

PR #355, opened 2017-10-10
Fixing RPC error occurring when creating market

adrake33:
This fixes an RPC error that was occurring when attempting to create a binary/categorical/scalar market.

### DIFF
--- a/source/libraries/ContractDeployer.ts
+++ b/source/libraries/ContractDeployer.ts
@@ -298,7 +298,6 @@ export class ContractDeployer {
     private async createCategoricalMarket(universe, numOutcomes, endTime, feePerEthInWei, denominationToken, designatedReporterAddress, numTicks): Promise<ContractBlockchainData> {
         const constant = { constant: true };
         const myUniverse = await parseAbiIntoMethods(this.ethjsQuery, this.signatures["Universe"], { to: universe.address, from: this.testAccounts[0].address, gas: "0x5b8d80" });
-        const marketCreation = await parseAbiIntoMethods(this.ethjsQuery, this.signatures["MarketCreation"], { to: this.contracts["MarketCreation"].address, from: this.testAccounts[0].address, gas: "0x5b8d80" });
         const marketFeeCalculator = await parseAbiIntoMethods(this.ethjsQuery, this.signatures["MarketFeeCalculator"], { to: this.contracts["MarketFeeCalculator"].address, from: this.testAccounts[0].address, gas: "0x5b8d80" });
 
         // necessary because it is used part of market creation fee calculation
@@ -310,6 +309,8 @@ export class ContractDeployer {
 
         const reportingWindowAddress = await myUniverse.getCurrentReportingWindow.bind(constant)();
         const marketCreationFee = await marketFeeCalculator.getMarketCreationCost.bind(constant)(reportingWindowAddress);
+
+        const marketCreation = await parseAbiIntoMethods(this.ethjsQuery, this.signatures["MarketCreation"], { to: this.contracts["MarketCreation"].address, from: this.testAccounts[0].address, gas: "0x5b8d80", value: marketCreationFee });
         const marketAddress = await marketCreation.createMarket.bind({ to: this.contracts["MarketCreation"].address, from: this.testAccounts[0].address, gas: "0x5b8d80", value: marketCreationFee, constant: true })(universe.address, endTime, numOutcomes, feePerEthInWei, denominationToken.address, numTicks, designatedReporterAddress);
         if (!marketAddress) {
             throw new Error("Unable to create new categorical market.");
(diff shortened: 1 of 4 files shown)


source/libraries/ContractDeployer.ts
```diff
@@ -291,11 +292,11 @@ export class ContractDeployer {
         return market;
     }
 
-    private async createBinaryMarket(universe, endTime: number, feePerEthInWei: number, denominationToken, designatedReporterAddress, numTicks: number): Promise<ContractBlockchainData> {
+    private async createBinaryMarket(universe, endTime: number, feePerEthInWei: number, denominationToken, designatedReporterAddress, numTicks: number): Promise<{ [methodName: string]: ContractMethod }> {
```
MicahZoltu: We should probably create a new type for `Contract` that this returns (as well as updating `parseAbiIntoMethods` to return it instead of its inline type).

source/libraries/ContractDeployer.ts
```diff
@@ -310,16 +311,20 @@ export class ContractDeployer {
...
+        const market = await parseAbiIntoMethods(this.ethjsQuery, this.signatures["Market"], { to: marketAddress, from: this.testAccounts[0].address, gas: "0x5b8d80" });
+        const marketNameHex = await padAndHexlify("Market", 64, "right");
+        if (await market.getTypeName() !== marketNameHex) {
+            throw new Error("Unable to create new categorical market");
+        }
+        return market;
     }
 
-    private async createScalarMarket(universe, endTime, feePerEthInWei, denominationToken, numTicks, designatedReporterAddress): Promise<ContractBlockchainData> {
+    private async createScalarMarket(universe, endTime, feePerEthInWei, denominationToken, numTicks, designatedReporterAddress): Promise<{ [methodName: string]: ContractMethod }> {
```
MicahZoltu: We can delete one of these two functions and rename the other to `createMarket`.  There isn't any difference between scalar and categorical anymore (there used to be).

source/libraries/HelperFunctions.ts
```diff
@@ -22,16 +22,21 @@ interface RpcClientOptions {
...
     for (let i = 0; i < length - (stringToEncode.length * 2); i++) {
         pad += "\x00";
     }
-    const paddedStringToEncode = pad + stringToEncode;
+    let paddedStringToEncode: string;
+    if (direction === "right") {
+        paddedStringToEncode = stringToEncode + pad;
+    } else {
+        paddedStringToEncode = pad + stringToEncode;
+    }
```
MicahZoltu: Recommend adding a helper method like `stringTo32ByteHex` that takes in a string as a parameter returns a 66 character hex value.  This is a pretty common operation, so there is value in making it simple to call.

source/libraries/HelperFunctions.ts
```diff
@@ -22,16 +22,21 @@ interface RpcClientOptions {
 }
 
 /**
- * Pads a string with zeros on the left and converts it to hexidecimal format.
+ * Pads a string with zeros (on the left by default) and converts it to hexidecimal format.
  * @param stringToEncode ASCII string to be padded & hexlified
  * @param length Desired length of the returned string (minus the "0x")
  */
-export async function padAndHexlify(stringToEncode: string, length: number): Promise<string> {
+export async function padAndHexlify(stringToEncode: string, length: number, direction: string = "left"): Promise<string> {
```
MicahZoltu: THis doesn't `await` so it doesn't need to be `async`.  Making it a normal function will allow you to change the return type to just `string` and make it so you don't have to `await` the method when you call it, allowing it to be inlined easier.

source/libraries/ContractDeployer.ts
```diff
@@ -7,6 +7,7 @@ import * as EthjsContract from "ethjs-contract";
 import * as EthjsQuery from "ethjs-query";
 // TODO: Update TS type definition for ContractBlockchainData to allow for empty object (e.g. upload() & uploadAndAddToController())?
 import { ContractBlockchainData, ContractReceipt } from "contract-deployment";
+import { ContractMethod } from "./AbiParser";
```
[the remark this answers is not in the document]
adrake33: Oops, didn't see the existing line.  I'll combine them.

source/libraries/ContractDeployer.ts
adrake33: Do we need to keep binary market?  Isn't it just a categorical market with a specific number of outcomes?

source/libraries/HelperFunctions.ts
adrake33: Ah, yeah.  Not sure why I originally made this async.  I'll change it.

source/libraries/HelperFunctions.ts
adrake33: Should I also have a parameter for which side to pad on?  Or will it always be one side?

MicahZoltu: For strings it will always be right padded.

source/libraries/ContractDeployer.ts
MicahZoltu: I didn't see a `createBinaryMarket` function, if there is one it can be removed as well.  The `createReasonable*Market` helpers are still potentially valuable though.  They just should all call into the same `createMarket` function under the hood.

adrake33: Ok, I'll keep those but just have a single createMarket() that they call.

adrake33: Actually, I'm going to leave this as a TODO item so we can get this merged quickly.

PR #356, opened 2017-10-10
Remove registration token

nuevoalex:
This is a rather large pull request so understood if it takes some time.

One thing that I am not pleased with in here is that I had to move market initialization out of the factory call and into the reporting window so that the trusted payment of the no-show REP bond can be done during market initialization. If you have nicer ideas on how we could do that I'm open to em.


source/contracts/extensions/MarketExtensions.sol
```diff
@@ -71,22 +75,22 @@ contract MarketExtensions {
         int256 _payoutStake = int256(_reportingToken.totalSupply());
 
         IDisputeBond _designatedDisputeBond = _market.getDesignatedReporterDisputeBondToken();
-        IDisputeBond _limitedDisputeBond = _market.getLimitedReportersDisputeBondToken();
-        IDisputeBond _allDisputeBond = _market.getAllReportersDisputeBondToken();
+        IDisputeBond _firstDisputeBond = _market.getFirstReportersDisputeBondToken();
+        IDisputeBond _lastDisputeBond = _market.getLastReportersDisputeBondToken();
```
MicahZoltu: I would still like a better name for the reporting phases.  This name makes it seem like `_firstDisputeBond` is the first dispute bond, but it isn't since the designated reporter dispute bond comes first.

tinybike: `_secondDisputeBond`?

nuevoalex: `initialReporting` & `finalReporting`?

source/contracts/extensions/MarketFeeCalculator.sol
```diff
@@ -15,42 +15,65 @@ contract MarketFeeCalculator {
...
     uint256 private constant TARGET_INVALID_MARKETS_DIVISOR = 100; // 1% of markets are expected to be invalid
     uint256 private constant TARGET_REP_MARKET_CAP_MULTIPLIER = 5;
 
     uint256 private constant TARGET_INCORRECT_DESIGNATED_REPORT_MARKETS_DIVISOR = 100; // 1% of markets are expected to have an incorrect designate report
 
-    function getValidityBond(IReportingWindow _reportingWindow) public returns (uint256) {
+    uint256 private constant TARGET_DESIGNATED_REPORT_NO_SHOWS_DIVISOR = 100; // 1% of markets are expected to have an incorrect designate report
+
+    function getValidityBond(IUniverse _universe) public returns (uint256) {
+        IReportingWindow _reportingWindow = _universe.getCurrentReportingWindow();
```
MicahZoltu: I'm curious why you decided to change this to take a `universe` instead of `reportingWindow`?  It appears you only use universe to get the reporting window, suggesting that you could just take in the reporting window instead?

MicahZoltu: Same a couple other times below.

nuevoalex: Since we always want these to use the universe's current reporting window if we took in a reporting window we'd have to get the universe from it and then get the current reporting window

source/contracts/reporting/Market.sol
```diff
@@ -70,26 +70,36 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
...
+        if (_refund > 0) {
+            require(owner.call.value(_refund)());
+        }
+        return true;
+    }
+
+    function assessFees() private returns (bool) {
+        MarketFeeCalculator _marketFeeCaluclator = MarketFeeCalculator(controller.lookup("MarketFeeCalculator"));
+        IUniverse _universe = getUniverse();
+        reportingWindow.getReputationToken().trustedTransfer(owner, this, _marketFeeCaluclator.getDesignatedReportNoShowBond(_universe));
```
MicahZoltu: Security: Someone can create a market and put another person as the market creator and this will withdraw tokens from `creator` without their authorization.  This combined with the initialization in the factory issue makes me wonder if we should come up with a better pattern for this.  As with many other pieces of code, the problem here is being able to provide transfer authorizations through an untrusted interface.

We have discussed previously using authorization signatures, but this requires an extract signing step by users which is a sucky experience.  We could make the whole stack trusted, and use `msg.sender` as the creator at the root and trust it to correctly set `creator` parameter that is passed through, but this increases the auditable code surface area and makes auditing generally harder.

Unfortunately, I don't have any particularly great ideas.  One simple (but still lame) option is to just have a separate market activation function that the user must call (directly) after the market is create (e.g., `market.activate()`.  This would be where it transfers the REP.  Another option would be to have the the REP get transferred through each contract (like msg.value gets passed along with each step).  This requires every step along the way can use `ReputationToken.trustedTransfer`, and will eat through a lot of gas constantly updating storage variables along the way (though I think there is an EIP out to make it so that only net state changes at end of transaction are billed).

MicahZoltu: If we do end up having two-step construction for market, we should just do all fee assessment as step 2, and allow any one to pay the fees.  This would allow someone to have a setup where an automated system creates a bunch of markets, and then some other system goes through and funds them with ETH/REP.  This could allow for clever security systems in complex automated market creation services.

source/contracts/extensions/MarketExtensions.sol
MicahZoltu: I'm 👍 on your suggestion in #dev for `round1` and `round2`.  This expands easily if we decide to add more rounds.  Alternatively, we _could_ create an array for each of these things and a constant that defines its size.  That may be over-engineering though, your call.

source/contracts/reporting/Market.sol
```diff
@@ -166,6 +192,8 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
     function updateTentativeWinningPayoutDistributionHash(bytes32 _payoutDistributionHash) public returns (bool) {
         var (_firstPlaceHash, _secondPlaceHash) = MarketExtensions(controller.lookup("MarketExtensions")).getOrderedWinningPayoutDistributionHashes(this, _payoutDistributionHash);
 
+        require(_firstPlaceHash != bytes32(0));
```
MicahZoltu: This requirement is only when placing a dispute bond.  It is acceptable for someone to buy outcome A, B or C when the current stakes are:
A: -10,000 REP
B: 0 REP
C: -500 REP

source/contracts/reporting/Market.sol
```diff
@@ -166,6 +192,8 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
...
+        require(_firstPlaceHash != _secondPlaceHash);
```
MicahZoltu: In what scenarios can this happen?  It feels like if `getOrderedWinningPayoutDistributionHashes` returns first == second then there is a bug (with the exception of when it returns `0,0`, which is acceptable)

source/contracts/reporting/Market.sol
```diff
@@ -124,38 +134,54 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
...
     }
 
-    function disputeDesignatedReport() public triggersMigration returns (bool) {
+    function disputeDesignatedReport(uint256[] _payoutNumerators, uint256 _attotokens) public triggersMigration returns (bool) {
         require(getReportingState() == ReportingState.DESIGNATED_DISPUTE);
         designatedReporterDisputeBondToken = DisputeBondTokenFactory(controller.lookup("DisputeBondTokenFactory")).createDisputeBondToken(controller, this, msg.sender, Reporting.designatedReporterDisputeBondAmount(), tentativeWinningPayoutDistributionHash);
         reportingWindow.getReputationToken().trustedTransfer(msg.sender, designatedReporterDisputeBondToken, Reporting.designatedReporterDisputeBondAmount());
-        updateTentativeWinningPayoutDistributionHash(tentativeWinningPayoutDistributionHash);
+        if (_attotokens > 0) {
+            require(derivePayoutDistributionHash(_payoutNumerators) != tentativeWinningPayoutDistributionHash);
```
MicahZoltu: 🤔 While it would be silly, is there really any reason to disallow someone from disputing A and then buying up A to the point where it is winning again?  They can do that anyway by just bringing B up to above `0` and then staking a bunch on A as a separate transaction.  TL;DR: I am not sure this requirement actually adds value.

source/contracts/reporting/Market.sol
nuevoalex: We should be able to assert this for all cases as we should never end up in the situation above since this applies when placing a dispite bond. We can remove this but it would break the system if we end up making changes where this is no longer the case

source/contracts/reporting/Market.sol
nuevoalex: There was a bug that let this happen so I added this to safegaurd. We should be able to remove it now but if something results in this case it will be a hard non obvious big to solve

source/contracts/reporting/Market.sol
nuevoalex: A second call to activate the market and do fee assessment sounds like the nicest way to me. Will refactor to that!

source/contracts/reporting/Market.sol
```diff
@@ -124,38 +134,54 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
...
-        reportingWindow.getReputationToken().trustedTransfer(msg.sender, limitedReportersDisputeBondToken, Reporting.limitedReportersDisputeBondAmount());
+    function disputeFirstReporters(uint256[] _payoutNumerators, uint256 _attotokens) public triggersMigration returns (bool) {
+        require(getReportingState() == ReportingState.FIRST_DISPUTE);
+        firstReportersDisputeBondToken = DisputeBondTokenFactory(controller.lookup("DisputeBondTokenFactory")).createDisputeBondToken(controller, this, msg.sender, Reporting.firstReportersDisputeBondAmount(), tentativeWinningPayoutDistributionHash);
+        reportingWindow.getReputationToken().trustedTransfer(msg.sender, firstReportersDisputeBondToken, Reporting.firstReportersDisputeBondAmount());
         IReportingWindow _newReportingWindow = getUniverse().getNextReportingWindow();
-        return migrateReportingWindow(_newReportingWindow);
+        _newReportingWindow.migrateMarketInFromSibling();
```
MicahZoltu: What made you decide to extract out this function?  In any migration we _must_ add the market to a new window and remove it from its current window and update the reporting window.  It feels like we would decrease the chance of error by packing those 3 operations (which must be atomic) into a function together so someone can't forget one.

nuevoalex: actually that's a good point. I believe this was not done mindfully. Will add back.

source/contracts/reporting/ReportingWindow.sol
```diff
@@ -33,39 +31,35 @@ contract ReportingWindow is DelegationTarget, Typed, Initializable, IReportingWi
...
     uint256 private totalWinningReportingTokens;
 
     function initialize(IUniverse _universe, uint256 _reportingWindowId) public beforeInitialized returns (bool) {
         endInitialization();
         universe = _universe;
         startTime = _reportingWindowId * universe.getReportingPeriodDurationInSeconds();
-        RegistrationTokenFactory _registrationTokenFactory = RegistrationTokenFactory(controller.lookup("RegistrationTokenFactory"));
-        registrationToken = _registrationTokenFactory.createRegistrationToken(controller, this);
-        // Initialize these to some reasonable value to handle the first market ever created without branching code
+        // Initialize these to some reasonable value to handle the first market ever created without branching code
```
MicahZoltu: s/these/this/

source/contracts/reporting/ReportingWindow.sol
```diff
@@ -146,30 +139,17 @@ contract ReportingWindow is DelegationTarget, Typed, Initializable, IReportingWi
...
-            // always give credit for events in all-reporters phase
-            privateNoteReport(_market, _reporter);
-        } else if (_state == IMarket.ReportingState.DESIGNATED_REPORTING) {
-            privateNoteReport(_market, _reporter);
-        } else if (numberOfReportsByMarket[_market] < getMaxReportsPerLimitedReporterMarket()) {
-            // only give credit for limited reporter markets up to the max reporters for that market
-            privateNoteReport(_market, _reporter);
-        }
-        // no credit in all other cases (but user can still report)
+        reportingGasPrice.record(tx.gasprice);
```
MicahZoltu: We should probably rename this method to `noteReportingGasPrice`.  I think we can also get rid of most of the require statements, all we really care about is just asserting that the caller actually staked some REP as part of this transaction, meaning we just need to do a provenance check on `msg.sender` to make sure it was from a reporting token (`markets.contains(_market)` and `market.isContainerForReportingToken(msg.sender)`).  I don't think we need the state checks anymore.

source/contracts/extensions/MarketFeeCalculator.sol
MicahZoltu: Ah, I see.  This allows us to not have to trust the caller to provide current reporting window.  👍

source/contracts/reporting/Market.sol
MicahZoltu: Actually, I suppose since dispute bonds are the only thing that _decreases_ a staked outcome, this will always be true after every normal report so the check _effectively_ is only on the disputes.  Feel free to leave it in that case.

source/contracts/reporting/Market.sol
nuevoalex: We decided after beginning to implement this and seeing how much of a sprawling hazard it actually is that we'll instead have one transaction still, but will no longer exist on `MarketCreation`, which will be removed. Instead `ReportingWindow.createMarket` will do a trusted transfer from the `msg.sender` to the `MarketFactory` which will then transfer the REP to the `Market` before initialization. If during initialization the `Market` has a REP balance < the bond it will fail

PR #357, opened 2017-10-11
Adds docker-compose files.

MicahZoltu:
Also adds TS source map support for stack traces.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,29 @@
+version: "3"
+services:
+  parity-dev-node:
+    build: ./parity-dev-node
+    ports:
+      - "47622:8545"
+      - "47621:8180"
+  # geth-dev-node:
+  #   build: ./geth-dev-node
+  #   ports:
+  #     - "47624:8545"
+  parity-integration-tests:
+    build:
+      context: .
+      dockerfile: Dockerfile-integration-tests
+    environment:
+      - ETHEREUM_HOST=parity-dev-node
+      - ETHEREUM_PORT=8545
+    links:
+      - parity-dev-node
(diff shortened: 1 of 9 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,29 @@
+version: "3"
+services:
+  parity-dev-node:
+    build: ./parity-dev-node
+    ports:
+      - "47622:8545"
+      - "47621:8180"
+  # get-dev-node:
```
adrake33: Minor typo: should be `# geth-dev-node:`.

source/libraries/ContractDeployer.ts
```diff
@@ -42,11 +42,12 @@ export class ContractDeployer {
     }
 
     public async deploy(): Promise<boolean> {
-        this.testAccounts = await generateTestAccounts(this.testAccountSecretKeys);
+        const accounts = await this.ethjsQuery.accounts();
+        this.testAccounts = accounts;
```
adrake33: Do you think it's worth keeping `generateTestAccounts` for unit tests, or should I remove it?

adrake33: Also, should I remove the `secretKeys` being passed into the constructor along with `this.testAccountSecretKeys`?

MicahZoltu: We may end up bringing back `generateTestAccounts` for unit tests, but we can always just pull them out of git history.  Same with the secret keys I think.  So no need to keep them around for now I think.

docker-compose.yml
MicahZoltu: Fixed.

PR #358, opened 2017-10-12
added check when ReportToken is initialize check payout numerators ad…

bthaile:
added check when ReportToken is initialize check payout numerators add up to numTicks


MicahZoltu:
Probably a good idea to add a test of this change along with the change, just to prove it works.

PR #359, opened 2017-10-12
Fixes initial deploy issues.

MicahZoltu:


source/libraries/ContractDeployer.ts
```diff
@@ -317,55 +251,27 @@ export class ContractDeployer {
...
+        return await this.createMarket(universe, numOutcomes, blockDateTimePlusDay.getTime()/1000, 10 ** 16, denominationToken, this.testAccounts[0], numOutcomes * 10**4);
     }
 
     private async createReasonableScalarMarket(universe, priceRange, denominationToken): Promise<Contract> {
         const block = await this.ethjsQuery.getBlockByNumber(0, true);
         const blockDateTime = await this.parseBlockTimestamp(block.timestamp);
         const blockDateTimePlusDay = new Date();
         blockDateTimePlusDay.setDate(blockDateTime.getDate() + 1);
-        return await this.createScalarMarket(universe, blockDateTimePlusDay.getTime()/1000, 10 ** 16, denominationToken, 40 * 10 ** 18, this.testAccounts[0]);
+        return await this.createMarket(universe, 2, blockDateTimePlusDay.getTime()/1000, 10 ** 16, denominationToken, this.testAccounts[0], 10**4);
```
adrake33: It seems like this function is doing the exact same thing as `createReasonableBinaryMarket`.  Is that supposed to be the case now after the `numTicks` changes?  If so, maybe one of these functions should just call the other?

source/deployment/deployContracts.ts
```diff
@@ -32,8 +32,7 @@ export async function compileAndDeployContracts(): Promise<ContractDeployer> {
     }
     const ethjsHttpProvider = new EthjsHttpProvider(`http://${ethjsHttpProviderHost}:${ethjsHttpProviderPort}`);
     const ethjsQuery = new EthjsQuery(ethjsHttpProvider);
-    const contractJson = await fs.readFile(CONTRACT_OUTPUT_DIR_PATH + "/" + COMPILED_CONTRACT_OUTPUT_FILE_NAME, "utf8");
-    const contractDeployer = new ContractDeployer(ethjsQuery, contractJson, GAS_AMOUNT, TEST_ACCOUNT_SECRET_KEYS);
+    const contractDeployer = new ContractDeployer(ethjsQuery, compilerOutput, GAS_AMOUNT, TEST_ACCOUNT_SECRET_KEYS);
```
adrake33: Thanks again for catching and fixing this race condition.

source/libraries/ContractDeployer.ts
MicahZoltu: Ah, good point.  Binary markets and scalar markets are no longer different aside from potentially the scale of `numTicks`.  I'll delete one of them.  I may actually reduce down to a single `createReasonableMarket` that just takes in `numOutcomes`....

PR #361, opened 2017-10-12
Add ability to forgo fees when redeeming reporting tokens

nuevoalex:


source/contracts/reporting/ReportingToken.sol
```diff
@@ -115,7 +117,9 @@ contract ReportingToken is DelegationTarget, Typed, Initializable, VariableSuppl
         if (_reporterReputationShare != 0) {
             _reputationToken.transfer(msg.sender, _reporterReputationShare);
         }
-        _reportingWindow.collectReportingFees(msg.sender, _attotokens);
+        if (!forgoFees) {
+            _reportingWindow.collectReportingFees(msg.sender, _attotokens);
+        }
```
MicahZoltu: If the user is forgoing fees, we will need to deduct their reporting tokens from the total here, so other people get their share of fees and the fees don't end up sitting unclaimed.  This is over in `collectReportingFees` function, we basically need to do the same thing just without paying out.
```
totalWinningReportingTokens = totalWinningReportingTokens.sub(_attoReportingTokens);
```

MicahZoltu: Also recommend adding an explicit test for this, since its a subtlety that is easy to miss/break in the future.

nuevoalex: Good catch, and will do

nuevoalex:
Added a CONSIDER for the future as well

source/contracts/reporting/ReportingToken.sol
```diff
@@ -102,13 +102,16 @@ contract ReportingToken is DelegationTarget, Typed, Initializable, VariableSuppl
 
     // NOTE: UI should warn users about calling this before first calling `migrateLosingTokens` on all losing tokens with non-dust contents
     // NOTE: we aren't using the convertToAndFromCash modifier here becuase this isn't a whitelisted contract. We expect the reporting window to handle disbursment of ETH
-    function redeemWinningTokens() public afterInitialized returns (bool) {
+    // CONSIDER: If all markets are finalized should we allow forgoing fees?
```
MicahZoltu: Meh, whatever is easier.  There is no reason someone would other than on accident, and the UI can help protect users from that.  There is also no problem with someone skipping fees if they like.

PR #365, opened 2017-10-14
Race condition

tinybike:


package-lock.json
```diff
@@ -1887,6 +1887,12 @@
             "integrity": "sha1-VgiurfwAvmwpAd9fmGF4jeDVl8g=",
             "dev": true
         },
+        "nan": {
```
adrake33: Are these non-sleep changes supposed to be here?  What are they for?

tinybike: I'm not sure, that was automatically added from `npm install`!

adrake33: Looks like nan is a dependency: https://www.npmjs.com/package/sleep.  `string_decoder` looks like it's the same and just got moved elsewhere in the file, so these changes should be ok.

package.json
```diff
@@ -20,6 +20,7 @@
         "mocha": "3.5.3",
         "path": "0.12.7",
         "recursive-readdir": "2.2.1",
+        "sleep": "5.1.1",
```
MicahZoltu: This library blocks the main event loop so shouldn't be used.  Instead, I recommend a little function like:
```
async function sleep(milliseconds: number): Promise<void> {
  return new Promise((resolve, reject) => {
    setTimeout(resolve, milliseconds);
  }
}
```
Then you can use it like:
```
await sleep(500);
```

source/libraries/ContractDeployer.ts
```diff
@@ -19,6 +19,7 @@ export class ContractDeployer {
     public readonly contracts = [];
     public readonly signatures = [];
     public readonly bytecodes = [];
+    public readonly transactionReceipts = [];
```
MicahZoltu: This doesn't appear to be used anywhere, should be deleted.

source/libraries/ContractDeployer.ts
```diff
@@ -193,8 +196,9 @@ export class ContractDeployer {
     private async createGenesisUniverse(): Promise<ContractBlockchainData> {
         const delegatorBuilder = this.ethjsContract(this.signatures["Delegator"], this.bytecodes["Delegator"], { from: this.testAccounts[0], gas: this.gasAmount });
         const universeBuilder = this.ethjsContract(this.signatures["Universe"], this.bytecodes["Universe"], { from: this.testAccounts[0], gas: this.gasAmount });
-        const receiptAddress = await delegatorBuilder.new(this.controller.address, `0x${binascii.hexlify("Universe")}`);
-        const receipt = await this.ethjsQuery.getTransactionReceipt(receiptAddress);
+        const transactionHash = await delegatorBuilder.new(this.controller.address, `0x${binascii.hexlify("Universe")}`);
+        await waitForTransactionToBeSealed(this.ethjsQuery, transactionHash);
+        const receipt: ContractReceipt = await this.ethjsQuery.getTransactionReceipt(transactionHash);
```
MicahZoltu: Style: Remove the type definition on this variable, it should be inferred from `getTransactionReceipt`. If it isn't, we should update the `.d.ts` file.

source/libraries/ContractDeployer.ts
```diff
@@ -193,8 +196,9 @@ export class ContractDeployer {
     private async createGenesisUniverse(): Promise<ContractBlockchainData> {
         const delegatorBuilder = this.ethjsContract(this.signatures["Delegator"], this.bytecodes["Delegator"], { from: this.testAccounts[0], gas: this.gasAmount });
         const universeBuilder = this.ethjsContract(this.signatures["Universe"], this.bytecodes["Universe"], { from: this.testAccounts[0], gas: this.gasAmount });
-        const receiptAddress = await delegatorBuilder.new(this.controller.address, `0x${binascii.hexlify("Universe")}`);
-        const receipt = await this.ethjsQuery.getTransactionReceipt(receiptAddress);
+        const transactionHash = await delegatorBuilder.new(this.controller.address, `0x${binascii.hexlify("Universe")}`);
+        await waitForTransactionToBeSealed(this.ethjsQuery, transactionHash);
```
MicahZoltu: It feels like this is the wrong place to sleep, in particular, we aren't doing an `eth_call` next, we are calling another on-chain transaction (`universe.initialize`).  Perhaps I am misunderstanding the need for the sleep here?

source/libraries/HelperFunctions.ts
```diff
@@ -71,3 +74,11 @@ export async function initializeTestRpcClientOptions(ethHttpProviderPort: number
 
     return { gasLimit: GAS_BLOCK_AMOUNT, accounts: accountOptions };
 }
+
+export async function waitForTransactionToBeSealed(ethjsQuery: EthjsQuery, transactionHash: string): Promise<void> {
+    let transaction = await ethjsQuery.getTransactionByHash(transactionHash);
+    while (transaction.blockNumber == null) {
+        await sleep.sleep(5);
```
MicahZoltu: 5 seconds is _way_ too long.  This should probably be something on the order of 100 milliseconds or less, maybe even 10 milliseconds.

source/libraries/HelperFunctions.ts
```diff
@@ -71,3 +74,11 @@ export async function initializeTestRpcClientOptions(ethHttpProviderPort: number
...
+        transaction = await ethjsQuery.getTransactionByHash(transactionHash);
```
MicahZoltu: It seems like the common case is that we'll want the transaction receipt after the transaction shows up in the block.  Consider using `ethjsQuery.getTransactionReceipt(transactionHash)` instead (which also includes block number and block hash) and returning it when done.  This will save the caller from having to then do an extra round trip afterwards to get the receipt.

source/libraries/HelperFunctions.ts
tinybike: If the block interval is 10-15 seconds, a 10-100ms retry will result in a lot of fast, failed retries.  I guess for something like instantseal it should be faster though.

source/libraries/ContractDeployer.ts
adrake33: When I was testing in Parity, sleeping just before creating the new market seemed to be sufficient.  While this may be working, I agree that it should be changed.  I didn't really have a way to test it with Geth until yesterday, when Scott helped me get set up using Geth.  I now have a way to debug better with Geth, so I'll try moving it to see if the error still occurs.

source/libraries/HelperFunctions.ts
tinybike: Using `getTransactionReceipt` (without first checking `getTransactionByHash`) was partially responsible for this problem, because Parity and geth make the transaction receipt available at different times, and when geth doesn't make the transaction receipt available it simply returns `null`.  `getTransactionByHash` behaves the same way in both clients AFAICT (and is also the way that ethrpc checks for this FWIW).

source/libraries/ContractDeployer.ts
adrake33: I was using this to debug so I could look at the receipts just before creating the new market.  I'll remove it.

source/libraries/HelperFunctions.ts
```diff
@@ -1,5 +1,8 @@
 import * as binascii from "binascii";
+import * as sleep from "sleep";
 import * as EthjsAccount from "ethjs-account";
+import * as EthjsQuery from 'ethjs-query';
+import { ContractReceipt } from "contract-deployment";
```
tinybike: Unused; should be deleted.

source/libraries/HelperFunctions.ts
MicahZoltu: We shouldn't assume a particular block time because when integration testing we can set the block time to whatever we want.  The only reason for the sleep at all is just so we don't melt the CPU while spinning.  I'm guessing 10ms is enough of a sleep to keep CPUs cool, hence the recommendation.

source/libraries/HelperFunctions.ts
MicahZoltu: Couldn't we just do this to address the "geth doesn't make receipts available until transaction is mined"?
```
while (!receipt || !receipt.blockNumber)
```

source/libraries/HelperFunctions.ts
adrake33: I'll remove these.

source/libraries/HelperFunctions.ts
tinybike: IMO we should use the lookup (`getTransactionByHash`) that we've tested extensively and that we know works.  The receipts data structure (within the ethereum node) is different than the transactions data structure and it appears to have some strange bugs/edge cases, see for example https://github.com/ethereum/go-ethereum/issues/1746

source/libraries/HelperFunctions.ts
tinybike: Note: JUST `ContractReceipt` should be removed; not ethjs-query.

source/libraries/ContractDeployer.ts
tinybike: Without the sleep here, the script does not wait until the transaction is sealed.  As I understand it `await myContract.new()` is equivalent to ethrpc's `onSent` callback, i.e. it just waits until the transaction is _broadcast_.  If there's _nothing_ downstream of this contract creation that requires this transaction to be sealed prior to proceeding then that is fine.  However, if anything downstream uses this contract for anything, you have to make sure it's sealed before moving on.

source/libraries/HelperFunctions.ts
tinybike: This function will be called when we use the uploader on a live chain, not just when we're using it for integration testing / against instantseal Parity or testrpc.  Each iteration of this loop sends off an RPC request to the chain.  Each RPC request is small, but if you send one out every 10ms and the mainnet or Ropsten block interval is 15s, you'll send out on average 15000 RPC requests per transaction.  That represents a significant amount of network I/O and should be avoided if possible IMO.

source/libraries/ContractDeployer.ts
MicahZoltu: Agreed.  My line of thinking is that in order to keep the deployment fast, even against a live chain with 15 second blocks, we should only be waiting for receipts when its actually necessary to make progress.  If we wait for a block to be mined after every transaction deployment will take over 30-60 minutes on a live Byzantium chain vs 1-2 minutes.

source/libraries/HelperFunctions.ts
tinybike: @MicahZoltu I tried changing getTransactionByHash to getTransactionReceipt to see what would happen.  Here is the result in geth:
```
{ Error: [ethjs-rpc]  error with payload {"id":6350900769400,"jsonrpc":"2.0","params":["0x3be152e828c8321a73267d3528e00a86f5d4a27f22ea3b754c85c7abf5e5204f"],"method":"eth_getTransactionReceipt"} Error: [ethjs-provider-http] Invalid JSON RPC response from provider
    host: http://localhost:8545
    response:  ""
    responseURL: undefined
    status: 0
    statusText: 
  
    at /home/jack/src/augur-core/node_modules/ethjs-rpc/lib/index.js:53:26
    at XMLHttpRequest.request.onreadystatechange (/home/jack/src/augur-core/node_modules/ethjs-provider-http/lib/index.js:91:7)
    at XMLHttpRequestEventTarget.dispatchEvent (/home/jack/src/augur-core/node_modules/xhr2/lib/xhr2.js:64:18)
    at XMLHttpRequest._setReadyState (/home/jack/src/augur-core/node_modules/xhr2/lib/xhr2.js:353:12)
    at XMLHttpRequest._onHttpRequestError (/home/jack/src/augur-core/node_modules/xhr2/lib/xhr2.js:542:12)
    at ClientRequest.<anonymous> (/home/jack/src/augur-core/node_modules/xhr2/lib/xhr2.js:413:24)
    at emitOne (events.js:115:13)
    at ClientRequest.emit (events.js:210:7)
    at Socket.socketErrorListener (_http_client.js:401:9)
    at emitOne (events.js:115:13)
  value: 
   { Error: [ethjs-provider-http] Invalid JSON RPC response from provider
    host: http://localhost:8545
    response:  ""
    responseURL: undefined
    status: 0
    statusText: 
  
    at invalidResponseError (/home/jack/src/augur-core/node_modules/ethjs-provider-http/lib/index.js:43:23)
    at XMLHttpRequest.request.onreadystatechange (/home/jack/src/augur-core/node_modules/ethjs-provider-http/lib/index.js:88:17)
    at XMLHttpRequestEventTarget.dispatchEvent (/home/jack/src/augur-core/node_modules/xhr2/lib/xhr2.js:64:18)
    at XMLHttpRequest._setReadyState (/home/jack/src/augur-core/node_modules/xhr2/lib/xhr2.js:353:12)
    at XMLHttpRequest._onHttpRequestError (/home/jack/src/augur-core/node_modules/xhr2/lib/xhr2.js:542:12)
    at ClientRequest.<anonymous> (/home/jack/src/augur-core/node_modules/xhr2/lib/xhr2.js:413:24)
    at emitOne (events.js:115:13)
    at ClientRequest.emit (events.js:210:7)
    at Socket.socketErrorListener (_http_client.js:401:9)
    at emitOne (events.js:115:13)
     value: 
      XMLHttpRequest {
        onloadstart: null,
        onprogress: null,
        onabort: null,
        onerror: null,
        onload: null,
        ontimeout: [Function],
        onloadend: null,
        _listeners: {},
        onreadystatechange: [Function],
        _anonymous: undefined,
        readyState: 4,
        response: null,
        responseText: '',
        responseType: '',
        status: 0,
        statusText: '',
        timeout: 0,
        upload: [Object],
        _method: 'POST',
        _url: [Object],
        _sync: false,
        _headers: [Object],
        _loweredHeaders: [Object],
        _mimeOverride: null,
        _request: null,
        _response: null,
        _responseParts: null,
        _responseHeaders: null,
        _aborting: null,
        _error: null,
        _loadedBytes: 0,
        _totalBytes: 0,
        _lengthComputable: false } } }
```

source/libraries/HelperFunctions.ts
MicahZoltu: Are we planning on using this script against a remote chain?  15,000 requests against a local node on a modern computer shouldn't be too bad.  I'm amenable to 100ms (10 requests/second) if people are finding their computers to be catching on fire though.

tinybike: I think it's sensible to make it usable with a remote chain.  I guess I don't see the importance of a short sleep time here.  If you're using InstantSeal or testrpc the sleep shouldn't happen at all, since sealing happens immediately.  It could be annoying if you're using a private chain without InstantSeal, but with a shorter block interval.  What we could do is make it read from an environment variable `process.env.ETHEREUM_BLOCK_INTERVAL_MILLISECONDS` and if that variable isn't set default to reasonable values for the main chain (5 seconds is reasonable IMO).

source/libraries/ContractDeployer.ts
tinybike: Aside from the controller, the contracts are _uploaded_ at the same time, aren't they?  On geth (Ropsten) what appears to happen is all of the contract creation transactions are submitted simultaneously, and then slowly are sealed (since they're too big to cram all into one block).

source/libraries/HelperFunctions.ts
adrake33: Right, I was referring to `ContractReceipt` and `sleep` since Micah suggested returning a promise with setTimeout instead.

tinybike:
I don't know if this error is related to or caused by the changes here, but using geth (on Ropsten) I get the following in my terminal:
```
(node:32289) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: [ethjs-rpc] rpc error with payload {"id":6510089875622,"jsonrpc":"2.0","params":[{"from":"0x45abe632be8a96ef9ebf4a80fb1b5c5157b7b15d","gas":"0x5b8d80","gasPrice":"0x3b9aca00","to":"0x79266f095620ace1564c331407ce76424e03f03b","data":"0x095ea7b3000000000000000000000000d282ba2a51bd0f80b58b6580d08a69f8b34816ba0000000000000000000000000000000000000000000000000000000000000001"}],"method":"eth_sendTransaction"} null
(node:32289) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
(node:32289) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): Error: [ethjs-rpc] rpc error with payload {"id":6510089875622,"jsonrpc":"2.0","params":[{"from":"0x45abe632be8a96ef9ebf4a80fb1b5c5157b7b15d","gas":"0x5b8d80","gasPrice":"0x3b9aca00","to":"0x79266f095620ace1564c331407ce76424e03f03b","data":"0x095ea7b3000000000000000000000000d282ba2a51bd0f80b58b6580d08a69f8b34816ba0000000000000000000000000000000000000000000000000000000000000001"}],"method":"eth_sendTransaction"} null
(node:32289) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 3): Error: [ethjs-rpc] rpc error with payload {"id":6510089875620,"jsonrpc":"2.0","params":[{"from":"0x7c0d52faab596c08f484e3478aebc6205f3f5d8c","gas":"0x5b8d80","gasPrice":"0x3b9aca00","to":"0x79266f095620ace1564c331407ce76424e03f03b","data":"0x095ea7b3000000000000000000000000d282ba2a51bd0f80b58b6580d08a69f8b34816ba0000000000000000000000000000000000000000000000000000000000000001"}],"method":"eth_sendTransaction"} null
(node:32289) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 4): Error: [ethjs-rpc] rpc error with payload {"id":6510089875620,"jsonrpc":"2.0","params":[{"from":"0x7c0d52faab596c08f484e3478aebc6205f3f5d8c","gas":"0x5b8d80","gasPrice":"0x3b9aca00","to":"0x79266f095620ace1564c331407ce76424e03f03b","data":"0x095ea7b3000000000000000000000000d282ba2a51bd0f80b58b6580d08a69f8b34816ba0000000000000000000000000000000000000000000000000000000000000001"}],"method":"eth_sendTransaction"} null
(node:32289) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 5): Error: [ethjs-rpc] rpc error with payload {"id":6510089875619,"jsonrpc":"2.0","params":[{"from":"0xe1e212c353f7a682693c198ba5ff85849f8300cc","gas":"0x5b8d80","gasPrice":"0x3b9aca00","to":"0x79266f095620ace1564c331407ce76424e03f03b","data":"0x095ea7b3000000000000000000000000d282ba2a51bd0f80b58b6580d08a69f8b34816ba0000000000000000000000000000000000000000000000000000000000000001"}],"method":"eth_sendTransaction"} null
(node:32289) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 6): Error: [ethjs-rpc] rpc error with payload {"id":6510089875619,"jsonrpc":"2.0","params":[{"from":"0xe1e212c353f7a682693c198ba5ff85849f8300cc","gas":"0x5b8d80","gasPrice":"0x3b9aca00","to":"0x79266f095620ace1564c331407ce76424e03f03b","data":"0x095ea7b3000000000000000000000000d282ba2a51bd0f80b58b6580d08a69f8b34816ba0000000000000000000000000000000000000000000000000000000000000001"}],"method":"eth_sendTransaction"} null
(node:32289) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 7): Error: [ethjs-rpc] rpc error with payload {"id":6510089875618,"jsonrpc":"2.0","params":[{"from":"0xaff9cb4dcb19d13b84761c040c91d21dc6c991ec","gas":"0x5b8d80","gasPrice":"0x3b9aca00","to":"0x79266f095620ace1564c331407ce76424e03f03b","data":"0x095ea7b3000000000000000000000000d282ba2a51bd0f80b58b6580d08a69f8b34816ba0000000000000000000000000000000000000000000000000000000000000001"}],"method":"eth_sendTransaction"} null
(node:32289) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 8): Error: [ethjs-rpc] rpc error with payload {"id":6510089875618,"jsonrpc":"2.0","params":[{"from":"0xaff9cb4dcb19d13b84761c040c91d21dc6c991ec","gas":"0x5b8d80","gasPrice":"0x3b9aca00","to":"0x79266f095620ace1564c331407ce76424e03f03b","data":"0x095ea7b3000000000000000000000000d282ba2a51bd0f80b58b6580d08a69f8b34816ba0000000000000000000000000000000000000000000000000000000000000001"}],"method":"eth_sendTransaction"} null
(node:32289) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 9): Error: [ethjs-rpc] rpc error with payload {"id":6510089875621,"jsonrpc":"2.0","params":[{"from":"0x796325c525fed19c0e59a85be17ea4277f47689b","gas":"0x5b8d80","gasPrice":"0x3b9aca00","to":"0x79266f095620ace1564c331407ce76424e03f03b","data":"0x095ea7b3000000000000000000000000d282ba2a51bd0f80b58b6580d08a69f8b34816ba0000000000000000000000000000000000000000000000000000000000000001"}],"method":"eth_sendTransaction"} null
(node:32289) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 10): Error: [ethjs-rpc] rpc error with payload {"id":6510089875621,"jsonrpc":"2.0","params":[{"from":"0x796325c525fed19c0e59a85be17ea4277f47689b","gas":"0x5b8d80","gasPrice":"0x3b9aca00","to":"0x79266f095620ace1564c331407ce76424e03f03b","data":"0x095ea7b3000000000000000000000000d282ba2a51bd0f80b58b6580d08a69f8b34816ba0000000000000000000000000000000000000000000000000000000000000001"}],"method":"eth_sendTransaction"} null
```
I'm not sure where these errors are coming from exactly (somewhere in ethjs-rpc apparently, but whether this is an ethjs-rpc error, a geth error, or an us error is unclear).  @MicahZoltu @adrake33 if the cause of this isn't immediately apparent to you guys, I propose going ahead and merging this PR, then opening a new issue for these errors.

source/libraries/HelperFunctions.ts
tinybike: Just a heads-up @adrake33 I went ahead and removed ContractReceipt as part of a larger push.  (Sorry for scooping you!)

adrake33: No worries, I just saw that.

source/libraries/HelperFunctions.ts
MicahZoltu: As of Byzantium we _have_ to look at the receipt so we can know whether the transaction succeeded or failed.  The GitHub issue you linked is about chain reorgs which is actually a much harder problem to solve and using `getTransactionByHash` doesn't solve it either.  Are there other "known issues" with transaction receipts that we need to worry about?

At the moment, this script assumes that there are no chain reorgs and that transactions do not fail.  Neither of these are actually safe assumptions and we should weigh whether or not we want to devote development effort to dealing with them in the deployment scripts.  I think we _should_ come up with a solution for transaction failure since its pretty easy (see below).  I'm not sure what to do about chain reorgs as that is a much harder problem to solve.

```
// TODO: make this not an `async` function because Node is stupid and will log a warning about unhandled promises
async function exitOnTransactionFailure(transactionHash: string) {
  const receipt = await waitForTransactionToSeal(transactionHash);
  if (receipt.status) return;
  console.log(`Transaction failed: ${transactionHash}`);
  process.exit(1);
}
```

source/libraries/ContractDeployer.ts
MicahZoltu: They are, but we _should_ be waiting for them to seal before making any `eth_call`s that depend on them being available on the chain.  This is what I'm getting at in that we should target our "wait for X" to just when we actually need to wait for X, rather than liberally waiting for X for everything.

source/libraries/HelperFunctions.ts
MicahZoltu: Geth doesn't have instaseal, and even with Parity its a race as to whether it seals before or after you make the following `eth_call` since Parity is multithreaded.

I'm 👍 on an environment variable, though I would either name the variable `ETHEREUM_BLOCK_TIME_IN_MILLISECONDS` and have it be 15 seconds for main network or name the variable `ETHEREUM_POLLING_INTERVAL_IN_MILLISECONDS` and have it set to the desired polling interval.  I vote is for the latter, but I'm fine with the former as well (and then internally just have polling time be `block_time / 10` or something.

source/libraries/HelperFunctions.ts
MicahZoltu: @tinybike Hmm, that error suggests that your Geth node isn't actually responding to you, not that Geth returned an error.  Are you sure Geth was on/live at the time of the call?  I would have expected a null JSON-RPC response, not Geth to simply reject your HTTP connection (that isn't even an HTTP error, that is a TCP connection failure error).

MicahZoltu:
That error is a combination of a few things.
1. A bug in ethjs (patch has been submitted) around it not correctly including error details in failure responses.
2. A bug in ethjs (patch has been submitted) around it not correctly using promises internally in a couple places.
3. Something causing your transactions to fail.  This is a bug in the code that we should look into.  Its pretty easy to address (1) and (2) via a debugger.

source/libraries/HelperFunctions.ts
tinybike: Right, the receipt tells you if the transaction succeeded or failed.  This is an extra and quite valuable piece of info.  `getTransactionByHash` just tells you whether or not the transaction has been _sealed_.  There are definitely pathological cases the current simple implementation won't catch, for example if a transaction is silently dropped from the txpool and needs to be re-sent, or (like you said) if a transaction fails, if a chain reorg happens, etc.  However `getTransactionByHash` by itself without other checks should make this work most of the time!  `getTransactionReceipt` should (as you pointed out) accomplish the same thing but it appears to have some sharp edges so IMO for now it's sensible to just use `getTransactionByHash`.

source/libraries/HelperFunctions.ts
```diff
@@ -1,9 +1,12 @@
 import * as binascii from "binascii";
 import * as EthjsAccount from "ethjs-account";
+import * as EthjsQuery from 'ethjs-query';
 
-const DEFAULT_TEST_ACCOUNT_BALANCE = 1 * 10 ** 20; // Denominated in wei
+const DEFAULT_ETHEREUM_BLOCK_INTERVAL_MILLISECONDS: number = 1200;
+const ETHEREUM_BLOCK_INTERVAL_MILLISECONDS: number = process.env.ETHEREUM_BLOCK_INTERVAL_MILLSECONDS ? parseInt(process.env.ETHEREUM_BLOCK_INTERVAL_MILLISECONDS!, 10) : DEFAULT_ETHEREUM_BLOCK_INTERVAL_MILLISECONDS;
```
MicahZoltu: Recommend renaming `ETHEREUM_BLOCK_INTERVAL_MILLISECONDS` to `ETHEREUM_POLLING_INTERVAL_MILLISECONDS`.  This variable isn't used as the block interval, it is used as the polling interval.

MicahZoltu: Also recommend inlining `DEFAULT_ETHEREUM_BLOCK_INTERVAL_MILLISECONDS` since it is only used once and its usage context makes it obvious that its a default.

MicahZoltu: Style: No need for the type on either of these (or the ones below).  It is inferred from the right side of the equal sign.

source/libraries/HelperFunctions.ts
tinybike: > @tinybike Hmm, that error suggests that your Geth node isn't actually responding to you, not that Geth returned an error. Are you sure Geth was on/live at the time of the call? I would have expected a null JSON-RPC response, not Geth to simply reject your HTTP connection (that isn't even an HTTP error, that is a TCP connection failure error).
Digging a bit.  That error was thrown when I had the sleep time set to 50ms.  Just re-tried it (w/ 50ms sleep time) and the error is _also_ thrown when `getTransactionByHash` is used instead of `getTransactionReceipt`.  My guess is what's happening is just that geth's RPC server is overloading and sending back empty (??) responses.  I would have thought this would be a `'0x'` rather than a `''` response though.

source/libraries/HelperFunctions.ts
```diff
@@ -1,9 +1,12 @@
 import * as binascii from "binascii";
 import * as EthjsAccount from "ethjs-account";
+import * as EthjsQuery from 'ethjs-query';
 
-const DEFAULT_TEST_ACCOUNT_BALANCE = 1 * 10 ** 20; // Denominated in wei
+const DEFAULT_ETHEREUM_BLOCK_INTERVAL_MILLISECONDS: number = 1200;
```
MicahZoltu: The default will be used when running integration tests against TestRPC, so I think we should have the interval be something like 10ms since in that case sealing should be instant (I actually don't know how TestRPC sealing works, may need to investigate but we should set it up to seal ASAP).

tinybike: Per my other comment, if this number is set too low apparently geth will start returning invalid / malformed RPC responses.  How about 500ms or 1s default?

source/libraries/HelperFunctions.ts
tinybike: Ah, good call.

source/libraries/HelperFunctions.ts
MicahZoltu: The error is a TCP error, not an HTTP or JSON-RPC error.  You can tell by the `status: 0` bit (that is HTTP status codes).  No idea why Geth is completely tanking like this.

MicahZoltu: Another option would be an exponential backoff.  This would make it so against insta-seal or short block time chains we get the response quickly, and against long block time chains we don't hammer the server.  I'm thinking 10 ms to 5 seconds exponential (2*prev) backoff would be good?

source/libraries/HelperFunctions.ts
MicahZoltu: It feels like if Geth can't handle fist polling interval then you should use the environment variable to override.  I don't think it is a good idea to slow down our integration tests (which don't use environment variables, only the default) because Geth has a bug/odd behavior.

MicahZoltu: Note: The way TestRPC runs work is that it checks if environment variables are set and, if so, uses them and if not it starts TestRPC and uses defaults.  This way you can just run `mocha` and have the tests "just work", without any environmental setup.

source/libraries/ContractDeployer.ts
tinybike: As I understand it the sequence of events should be as follows:
1. controller contract creation broadcast
2. controller contract creation sealed
3. all other contract creations broadcast (in parallel)
4. all other contract creations sealed
5. contract initialization functions broadcast (in parallel (?) -- not sure if this is true or even should be true)
6. contract initialization functions sealed
7. integration tests (create market, etc.) which are currently not working (or semi-working, on parity instantseal)
Is this the _actual_ sequence of events occurring when we run the deployer, in its current form?  If so, this seems like the correct (although not optimized) sequence to me.  If not, then I agree it should be fixed.

source/libraries/HelperFunctions.ts
tinybike: > It feels like if Geth can't handle fist polling interval then you should use the environment variable to override. I don't think it is a good idea to slow down our integration tests (which don't use environment variables, only the default) because Geth has a bug/odd behavior.

Have you or @adrake33 run this against non-instantseal Parity?  I'd like to know if it exhibits the same behavior / how its RPC server behaves under heavy load.  Keep in mind that geth is by far the most common client (as well as the client we're currently running on our servers, and I believe it is the client used by INFURA as well) so the default settings should be configured to work correctly with geth IMO.

> Note: The way TestRPC runs work is that it checks if environment variables are set and, if so, uses them and if not it starts TestRPC and uses defaults. This way you can just run mocha and have the tests "just work", without any environmental setup.

I normally run tests using `npm test`, which can have any environment variable settings needed baked in.  However if you guys are using `mocha` by itself often I don't feel strongly about this.

tinybike: Follow-up: there isn't currently a `test` script defined in `package.json` (there are a number of other fields missing as well), and when I run `mocha` by itself I get the following error:
```
jack@substrate:~/src/augur-core$ mocha output/tests-unit/


  Universe
    1) "before each" hook for "#getTypeName()"


  0 passing (159ms)
  1 failing

  1) Universe "before each" hook for "#getTypeName()":
     Error: [ethjs-rpc]  error with payload {"id":3323123508113,"jsonrpc":"2.0","params":[],"method":"eth_accounts"} Error: [ethjs-provider-http] Invalid JSON RPC response from provider
    host: http://localhost:47622
    response:  ""
    responseURL: undefined
    status: 0
    statusText: 
  
      at node_modules/ethjs-rpc/lib/index.js:53:26
      at XMLHttpRequest.request.onreadystatechange (node_modules/ethjs-provider-http/lib/index.js:91:7)
      at XMLHttpRequestEventTarget.dispatchEvent (node_modules/xhr2/lib/xhr2.js:64:18)
      at XMLHttpRequest._setReadyState (node_modules/xhr2/lib/xhr2.js:353:12)
      at XMLHttpRequest._onHttpRequestError (node_modules/xhr2/lib/xhr2.js:542:12)
      at ClientRequest.<anonymous> (node_modules/xhr2/lib/xhr2.js:413:24)
      at Socket.socketErrorListener (_http_client.js:401:9)
      at emitErrorNT (internal/streams/destroy.js:64:8)
      at _combinedTickCallback (internal/process/next_tick.js:138:11)
      at process._tickCallback (internal/process/next_tick.js:180:9)
```
Do I need to have testrpc/parity/something else running / have some environmental variable(s) set to run the unit tests?

adrake33: @tinybike If there are no environment variables set, it will use testrpc by default.  Try doing it from another terminal with no environment variables set.

source/libraries/ContractDeployer.ts
MicahZoltu: No, the required sequence is:
1. all contracts uploaded (parallel)
2. contracts added to whitelist (parallel)
3. all contracts initialized (parallel)
4. integration tests are run

There is no need to wait for sealing for steps 1-3 because we don't need to do any `eth_call`s as part of the process.  I believe instantiating the market is the first time we need to make an `eth_call`, since we need to get the address of the universe.  We could technically work around this, but probably not worth the effort.

tinybike:
I'm going to go ahead and merge this, since as per our conversation above there aren't any blockers left, although we should continue discussing the issues raised above until we reach a conclusion.

source/libraries/ContractDeployer.ts
MicahZoltu: Note the above only works if the backing node will return transaction receipts for unsealed blocks (everything except Geth).  Without the ability to get the receipt for unfinished blocks, we either need to wait for sealing to get the address of the newly created contracts, or we need to calculate the contract addresses client side.

tinybike: Currently the script uploads the Controller before the other contracts.  The other contracts are uploaded in parallel.  Since you need all contract creations sealed before you run step 4, is there any benefit to not waiting while you're in the previous steps?  The rate limiting step here is the number of blocks needed (since the contract creations require many blocks total); getting to parts of steps 2/3 sooner shouldn't make a difference if you still need all of 1-3 to be done before 4.

Transaction receipts for unsealed blocks (with accompanying contract addresses) are subject to change until the block is sealed.  In my view, it doesn't make sense to rely on them.

> (everything except Geth)

Meaning Parity, cpp-ethereum, and pyethereum/pyethapp?  Have you guys tested in all these?  The only two clients that see serious use are Parity and geth.  They should be mostly compatible, but when they're not, all of our code _must_ consider geth to be the default, as it is the most widely used client.

adrake33: @tinybike The only testing I've done so far of the deployment script is with Parity/Geth.

MicahZoltu:
This script is not intended to be used by the general public, so "most common node implementation" is not important.  What matters is whether _we_ can deploy to mainnet and whether developers can easily test the contracts against Geth and Parity.  For the latter reason, we _do_ need to work with Geth, but IMO getting it working outside of docker-compose (which is what CI will use), is an incredibly low priority.  Since docker compose will explicitly specify every environment variable (it won't use defaults) having the default work with Geth doesn't add value.  Similarly, to work with anything other than TestRPC, the caller _must_ specify at least the `ETHEREUM_HOST` variable which means that the caller must have the capacity to set environment variables, know that they need to be set, and know what they are.  Since someone running against a local/production Geth node has to get over that barrier to entry already, requiring one more variable is not a big hurdle.

Running the test suite though is something that I would like to "just work" for everyone without having to do any custom setup.  Also, TestRPC and instant-seal/fast-block nodes will be the most common op by far, while running against local/production Geth nodes is something I see happening once in a blue moon (when we want to deploy and the person doing the deploy happens to use Geth instead of Parity).  In fact, I actually would argue that the production/testnet deploys should be automated and run as part of CI against a cloud node, in which case we can make that be Parity, which may not have the fast check problem.

PR #367, opened 2017-10-14
Changes polling interval to follow an exponential backoff.

MicahZoltu:
This reduces the amount of configuration necessary to run the node while supporting both fast chains and slow chains.  It also minimizes the negative impact of spamming a node with requests by having the interval slow down over time.

Note: When working against a remote node, the round-trip time to the node is included as part of the delay between transactions, so this actually starts at `round_trip+10` then increases to `round_trip+20` on the first iteration, `round_trip+30` on the next iteration, etc.

### DIFF
--- a/source/libraries/HelperFunctions.ts
+++ b/source/libraries/HelperFunctions.ts
@@ -2,11 +2,9 @@ import * as binascii from "binascii";
 import * as EthjsAccount from "ethjs-account";
 import * as EthjsQuery from 'ethjs-query';
 
-const DEFAULT_ETHEREUM_BLOCK_INTERVAL_MILLISECONDS = 1200;
 const DEFAULT_TEST_ACCOUNT_BALANCE = 1 * 10 ** 20; // Denominated in wei
 // Set gas block limit extremely high so new blocks don"t have to be mined while uploading contracts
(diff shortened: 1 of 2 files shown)


source/libraries/HelperFunctions.ts
```diff
@@ -80,9 +78,11 @@ async function sleep(milliseconds: number): Promise<object> {
 }
 
 export async function waitForTransactionToBeSealed(ethjsQuery: EthjsQuery, transactionHash: string): Promise<void> {
+    let pollingInterval = 10;
```
tinybike: Recommend testing this vs live nodes (Parity and geth) to make sure this doesn't overload their RPC servers.

tinybike:
Have you confirmed that the sleep loop is being triggered when the deploy script is run against Parity InstantSeal and/or TestRPC?  It's not clear to me if there's a real issue to be addressed here or not.

Note: if the polling interval is being removed entirely, the associated environment variable in docker-compose.yml should be removed.  My suggestion would be to instead keep the environment variable and use it as the starting value for the exponential backoff.

MicahZoltu:
> the associated environment variable in docker-compose.yml should be removed

Fixed.

PR #368, opened 2017-10-14
Fork payout increases

nuevoalex:
I did the 5% bonus along with the dispute bond migration work since they interact with each other indirectly.

Something to note about this interaction in particular is that the 5% applies to the initial migration and is factored in to the remaining extra bond payout.

So if I migrate a bond with 100 REP and a total desired payout of 200, the migration itself will result in 105 REP in the destination universe. My remaining REP to be paid is now 95 rather than 100. This is to attempt and get everyone who has bonds their fair share of the matching pot from the disputed universe.

Additionally the 5% bonus is not counted toward the REP that may be used to match bond payout minting. We track migrated REP separate from the total supply in order to do this.


MicahZoltu:
For migrated bond REP bonus payout, I'm OK with either +95, +100 or +105, whichever is easier.  The 2x multiplier is already arbitrary, so being off by 5% in either direction isn't a big deal.

source/contracts/reporting/DisputeBondToken.sol
```diff
@@ -38,14 +40,19 @@ contract DisputeBondToken is DelegationTarget, Typed, Initializable, ERC20Basic,
...
+        if (bondRemainingToBePaidOut < bondAmount) {
+            uint256 _amountToDeductFromExtraPayout = _amountToTransfer.min(bondAmount - bondRemainingToBePaidOut);
+            getUniverse().decreaseExtraDisputeBondRemainingToBePaidOut(_amountToDeductFromExtraPayout); 
+        }
         _reputationToken.transfer(bondHolder, _amountToTransfer);
         return true;
     }
 
-    // FIXME: We should be minting coins in this scenario in order to achieve 2x target payout for bond holders during a fork.  Ideally, the amount minted is capped at the amount of tokens redeemed on other universes, so we may have to require the user to supply universes to deduct from with their call to this.
+    // NOTE: The UI should warn users about doing this before REP has migrated fully out of the forking universe in order to get their extra payout, which is sourced from REP migrated to other Universes. There is no penalty for waiting (In fact waiting will only ever get them a greater payout)
```
MicahZoltu: This comment (and implementation I think) aren't quite right.  You only get the 5% bonus REP if you call this function during the fork (not after the fork is over).  This means that there _is_ a penalty (opportunity cost) for migrating late.

Looking over the code, I would expect to see a `if (forkingMarket.isFinalized)` statement somewhere but I don't.  Perhaps I'm missing it?  If not, then that is necessary for calculating the % bonus.

I believe the 2x bonus should apply even if you migrate after the fork is finalized.

nuevoalex: The 5% bonus is applied in the ReputationToken as it applies to all migration of REP. The comment is just referring to the 2x bonus.

nuevoalex: More on that too since it's kind of interesting. I did it this way so there was a balance between encouraging migration of dispute bonds during the fork but also not forgoing potential payouts from REP that may later be migrated to the disputed universe. This way there is a theoretical benefit to the bond holder whether they wait or not, in particular because the 5% bonus will count toward the desired 2x payoff, so people cashing in earlier take less from the pool of REP migrated to the disputed universe.

MicahZoltu: Is it possible for a dispute bond holder to force StakeTokens to be redeemed in their respective forks?  There may be value in allowing anyone to force migrate stake so that the dispute bond holders can force other people's REP to migrate before they call this function.  Alternatively, we could include staked-but-not-migrated REP in the available-to-mint calculation, though that may result in making the accounting more complicated/difficult.

MicahZoltu: Also of note, the 3 dispute bond holders will be competing with each other for what may be a limited pool of REP available for minting.  It would be nice if we could allocate REP to the smaller bonds first, bigger bonds second but I'm worried that would make the accounting more difficult and not really provide significant value.  Thoughts?

nuevoalex: The issue with the forcing migration function is that the REP holder may select a different universe as right now this only matches with the disputed universe.

Regarding the 3 bonds the calculation right now is based on the percentage of "extra" payout the bond needs relative to all other bonds in that universe, so it should be proportional. While it doesn't favor the smaller bonds it does at least attempt to do something reasonably fair.

This logic existsi n the `mintRepTokensForBondMigration` function

source/contracts/reporting/ReputationToken.sol
```diff
@@ -45,8 +46,14 @@ contract ReputationToken is DelegationTarget, Typed, Initializable, StandardToke
     }
 
     function migrateIn(address _reporter, uint256 _attotokens) public afterInitialized returns (bool) {
-        require(ReputationToken(msg.sender) == universe.getParentUniverse().getReputationToken());
+        IUniverse _parentUniverse = universe.getParentUniverse();
+        require(ReputationToken(msg.sender) == _parentUniverse.getReputationToken());
         balances[_reporter] = balances[_reporter].add(_attotokens);
+        // Only count tokens migrated toward the available to be matched in other universes. The bonus should not be added
+        universe.increaseRepAvailableForExtraBondPayouts(_attotokens);
```
MicahZoltu: If I am understanding the code correctly, the amount of REP available for migration _out_ is stored as a storage variable on the child universe.  Then when someone mints REP in another universe that is deducted.  The problem with this is that if there are 3 competing universes (with activity) then REP migrated to child A should be mintable on _both_ children B and C.  This means that really we should be keeping a `mapping(mint_destination => uint256)`.

If this greatly increases complexity we can probably just put a TODO in and deal with it post-launch.

MicahZoltu: Curious: Why did you choose to track `rep_available_to_mint` instead of `rep_minted`?  If it were inversed and you instead just tracked REP minted on other universes sourced from this one then you would only need to update the storage variable when someone mints REP in a destination universe (max 3 times per universe per fork) rather than every time someone migrates REP.  This is a minor gas cost optimization, so may not be worth changing, but I'm curious if there was a reason to go this direction?

MicahZoltu: Note: inverting the above _may_ make addressing the 3-universe problem above a bit easier.

nuevoalex: Right now the REP would be mintable in Universe B and C as described if the dispute bond was against universe A.

The issue did mention possibly allowing the user to specify other universes to attempt and pull from, which isn't incredibly difficult but this seemed like a fair starting point. Pulling automatically would require a bit more work since we'd have to do some iterative logic and take from each pool equally.

My opinion is it's puntable

nuevoalex: The comment above this actually describes why. I didn't want the 5% bonus to count toward this so that we actually stay zero-sum and no weirdness with timing gets involved. Otherwise we could just track amount minted and totalSupply as you are pointing out.

MicahZoltu: Would things be simplified if we were OK with counting the 5% towards mintable volume?

MicahZoltu: > Right now the REP would be mintable in Universe B and C as described if the dispute bond was against universe A.

Hmm, I'm not seeing this.  Scenario:
1. round1 has tentative winning outcome A, it is disputed (dispute against A) with 100 REP dispute bond
1. round2 has tentative winning outcome A (again), it is disputed (dispute against A) with 1000 REP dispute bond
1. During fork, universe A has 100 REP
1. round1 dispute holder migrates to universe B, pulling REP from A.  They should get 200 REP in universe B.
1. round2 dispute holder migrates to universe C, pulling REP from A.  They should get 1100 REP in universe C.

Since the dispute holders are migrating to different universes, they can both pull (separately) from the REP migrated to universe A, meaning they can each extract up to 100 REP from universe A, not a combined total of 100 REP extracted from universe A.

MicahZoltu: I'm fine with letting the 5% bonus count towards mintable volume if you think its worth it for simplicity.

nuevoalex: Each bond pull would decrease the amount available to match for minting from Universe A proportional to the amount needed by the bond.

So in the above scenario the round1 bond holder would actually mint 100 * 100/1200 = 8.3 REP
and since there are only 2 bonds in this universe the round2 bond holder would mint what remains.

PR #370, opened 2017-10-15
Docker files for starting Geth

adrake33:
This pre-generates the DAG used by Geth when the docker container is built.  Then it creates a new account, initializes the genesis block, unlocks the account, and starts Geth mining using the pre-generated DAG.

To run:
```
docker build . -t geth-dev-node
docker run --rm -it -p 8545:8545 geth-dev-node
```

Then, in a separate window, run:
```
export ETHEREUM_HOST=localhost
export ETHEREUM_PORT=8545
export ETHEREUM_GAS_PRICE_IN_NANOETH=1
node ./output/deployment/deployContracts.js
```


geth-dev-node/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM ethereum/client-go:latest
+COPY genesis.json /app/genesis.json
+RUN echo "p" > /app/password.txt
```
MicahZoltu: Recommend using the empty password, in my experience this simplifies a number of things.  `echo "" > /app/password.txt`

geth-dev-node/start.sh
```diff
@@ -0,0 +1,3 @@
+geth --password /app/password.txt account new
+geth init /app/genesis.json
+geth --password /app/password.txt --unlock 0 --mine --networkid=10000 --ws --wsapi eth,net,web3,personal --wsport 8546 --rpc --rpcapi eth,net,web3,personal,miner --rpcaddr 0.0.0.0 --targetgaslimit 6700000 --ethash.dagdir "/app/.ethash"
```
MicahZoltu: What does the `--unlock 0` represent here?  I thought this was supposed to take in an address like `--unlock 0x00a329c0648769a73afac7f9381e08fb43dbea72` or something?

geth-dev-node/start.sh
```diff
@@ -0,0 +1,3 @@
+geth --password /app/password.txt account new
```
MicahZoltu: I believe this will create an account with an otherwise unknown address.  I recommend instead creating an account with a well known address.  For example, this private key is created with the empty seed and has no password:
```
{
    "id": "e1042fa7-e661-aeb8-ca86-6b9999adc120",
    "version": 3,
    "crypto": {
        "cipher": "aes-128-ctr",
        "cipherparams": {
            "iv": "50f44a53e4e11a424074063200b2e333"
        },
        "ciphertext": "190abdf9a8e4135defcd3f890a574b178f24a25fbc46f0daab3ffbb83c83a59d",
        "kdf": "pbkdf2",
        "kdfparams": {
            "c": 10240,
            "dklen": 32,
            "prf": "hmac-sha256",
            "salt": "90a1e58cfa95479588cf754dbad73f8019d2469303dd69028e8d7a2f27bf8d82"
        },
        "mac": "27a556b61e874e7f41376b4b01930caf36cee1977722f441fb73ad8cb45fba03"
    },
    "address": "00a329c0648769a73afac7f9381e08fb43dbea72",
    "name": "Development Account",
    "meta": "{\"description\":\"Never use this account outside of develoopment chain!\",\"passwordHint\":\"Password is empty string\"}"
}
```

geth-dev-node/genesis.json
```diff
@@ -0,0 +1,18 @@
+{
+  "config": {
+        "chainId": 10000,
+        "homesteadBlock": 0,
+        "eip155Block": 0,
+        "eip158Block": 0
+    },
+  "alloc"      : {},
+  "coinbase"   : "0x0000000000000000000000000000000000000000",
+  "difficulty" : "0x400",
```
MicahZoltu: Were you able to figure out how to make Geth have shorter block times?  Nothing in this file stands out to me as such, but maybe I just don't know what to look for.

What does `difficulty` represent?

geth-dev-node/genesis.json
```diff
@@ -0,0 +1,18 @@
...
+  "extraData"  : "",
+  "gasLimit"   : "9000000",
```
MicahZoltu: Recommend setting this to 6.5M to be somewhat in line with the production network.

geth-dev-node/start.sh
adrake33: It represents the index in the array of accounts.  (I believe you might also be able to use the word `primary`.)  It was simpler to use the index than trying to get the exact address generated from `geth --password /app/password.txt account new`.

geth-dev-node/Dockerfile
adrake33: I think I did try this, but it seemed to be causing issues at the time.  I'll double-check, though.

geth-dev-node/genesis.json
adrake33: Yeah, good idea.  I'll update this.

geth-dev-node/genesis.json
adrake33: I checked with @epheph about this, but he said you have to use proof of authority, become the authority, and then run a miner as the authority at a specific interval.  However, he has not tried this before.

adrake33: I believe difficulty just represents the block difficulty.

geth-dev-node/start.sh
adrake33: Ok, I'll try using that account instead.

MicahZoltu: Note: it is possible that using a pre-created account causes problems is Geth, if so ignore this.

MicahZoltu:
Here is a genesis file for a clique node.  I just found it on the internet so it may not actually work.
```
{
  "config": {
    "chainId": 3762,
    "homesteadBlock": 1,
    "eip150Block": 2,
    "eip150Hash": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "eip155Block": 3,
    "eip158Block": 3,
    "clique": {
      "period": 15,
      "epoch": 30000
    }
  },
  "nonce": "0x0",
  "timestamp": "0x5908570a",
  "parentHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
  "extraData": "0x4f6e6469666c6f20436c69717565203337363200000000000000000000000000035e19356880f86804b5fbf7015192be271ab8e426177afe720b3ad16a84d04e195dc6d4a66f82cd801e1cf176afdf340b190a98799666d534f8e953c4849b7564b9763e65f63d0754472374a420af940000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
  "gasLimit": "0x47b760",
  "difficulty": "0x1",
  "mixHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
  "coinbase": "0x0000000000000000000000000000000000000000",
  "alloc": {
    "0000000000000000000000000000000000000000": {
      "balance": "0x1"
    },
  }
}
```

geth-dev-node/genesis.json
epheph: difficulty represents initial block difficulty. Initial blocks are mined VERY quickly, but they converge toward normal block times, and I don't believe there's an override for target block time.

Another solution is to come up with a system that doesn't mine 0-TX blocks, you'd keep that difficulty low for a long time.

adrake33:
Do we need to be able to run this on Ropsten?  My understanding from talking to @epheph is that Rinkeby and Kovan are POA, but Ropsten is POW.

MicahZoltu:
For the integration tests in docker we shouldn't be connecting to any public network, we should be running a private test network that is just us.  The scripts themselves need to be able to deploy to any network of course, so we can deploy to Ropsten or Foundation, but that is a separate problem from these docker hosts.

adrake33:
Ok, sounds good.  I'll try to use the clique node file.

geth-dev-node/start.sh
adrake33: I tried doing: 
```
RUN echo "" > /app/password.txt
```
in Dockerfile, and then 
```
geth --password /app/password.txt --unlock "00a329c0648769a73afac7f9381e08fb43dbea72" --mine --networkid=10000 --ws --wsapi eth,net,web3,personal --wsport 8546 --rpc --rpcapi eth,net,web3,personal,miner --rpcaddr 0.0.0.0 --targetgaslimit 6700000 --ethash.dagdir "/app/.ethash"
```
in start.sh, but it failed to unlock the account.  I also tried passing in "" instead of /app/password.txt, but it prompted me to type the password.

I'm not sure where in `genesis.json` I should be putting what you have above.  I don't think I've seen an example like that.

MicahZoltu: The JSON file goes in a keys folder (not sure where Geth keeps them exactly, should be listed on the Geth wiki).  I don't think filename actually matters.

adrake33: Alright, I think I know what to do now, but I'm going to focus on fixing the `getTypeName` bug first, since that's higher priority.

adrake33: BTW, I think the "UnhandledPromiseRejectionWarning" messages @tinybike was seeing may have been because the account became locked after the contracts started uploading.  Once I made the Docker file unlock the account initially, those messages went away.  Now all I see is the `createMarket` bug.

adrake33:
Actually, if we're going to try to use POA instead of POW, do we even need to worry about generating the DAG?  According to this, it sounds like maybe we don't?

https://ethereum.stackexchange.com/questions/6851/faster-way-to-generate-dag

geth-dev-node/start.sh
tinybike: I ran geth with a permanently unlocked account, so I don't think this is the case.

adrake33: @tinybike How did you permanently unlock it?  Did you do it when starting Geth?

adrake33: Hmm, interestingly, I finally saw the `UnhandledPromiseRejectionWarning` messages just now while using Docker.

NOTE: I've made 2 changes to my setup that you may not have, @tinybike:

- I made the change in this patch by Micah for ethjs-rpc: https://github.com/MicahZoltu/ethjs-rpc/commit/1343a7a2667f1d0b96c644f18f2be1e174835229.  (I made it in `augur-core/node_modules/ethjs/src/index.js` and `augur-core/node_modules/ethjs/lib/index.js`.)

- I also added `return result;` to `augur-core/source/libraries/AbiParser.ts` right after your call to `waitForTransactionToBeSealed`.

Doing both of these can give more helpful error messages.

When running with docker, I saw these slightly more helpful messages, which say "authentication needed: password or unlock":
```
(node:2163) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: [ethjs-rpc] rpc error with payload {"id":3285766361764,"jsonrpc":"2.0","params":[{"from":"0x3ccf9547e8238c04329bf434b8b64081891c4dc0","gas":"0x5b8d80","gasPrice":"0x3b9aca00","to":"0xd400c43145602d93c1545f979a3bf279dc420d2b","data":"0x095ea7b3000000000000000000000000853f471794ca98036d3c2fbf34e7a1fa2e8137da0000000000000000000000000000000000000000000000000000000000000001"}],"method":"eth_sendTransaction"} {"code":-32000,"message":"authentication needed: password or unlock"}
(node:2163) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
(node:2163) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): Error: [ethjs-rpc] rpc error with payload {"id":3285766361764,"jsonrpc":"2.0","params":[{"from":"0x3ccf9547e8238c04329bf434b8b64081891c4dc0","gas":"0x5b8d80","gasPrice":"0x3b9aca00","to":"0xd400c43145602d93c1545f979a3bf279dc420d2b","data":"0x095ea7b3000000000000000000000000853f471794ca98036d3c2fbf34e7a1fa2e8137da0000000000000000000000000000000000000000000000000000000000000001"}],"method":"eth_sendTransaction"} {"code":-32000,"message":"authentication needed: password or unlock"}
```

What's odd, though, is that Docker is unlocking the account when Geth starts, and I don't typically get these errors.

adrake33: I also noticed that when I restarted my Docker container just now, I was no longer getting `UnhandledPromiseRejectionWarning`.  When I closed my laptop last night (with the Docker container was still running), I was not seeing the `UnhandledPromiseRejectionWarning` messages at all.  However, once I started running the deployment script again this morning with the Docker container that had been running all night, the messages were appearing consistently.

PR #371, opened 2017-10-15
Forking bugs

nuevoalex:
I'm intentionally leaving out any testing for disavowed tokens since Tom is working on new StakeToken tests. We should also test the `disavowTokens` method on the `Market` before the fork is finalized.


source/contracts/libraries/collections/Map.sol
```diff
@@ -0,0 +1,53 @@
+pragma solidity 0.4.17;
+
+import 'libraries/DelegationTarget.sol';
+import 'libraries/Ownable.sol';
+import 'libraries/Initializable.sol';
+
+
+contract Map is DelegationTarget ,Ownable, Initializable {
+    mapping(bytes32 => address) private items;
```
MicahZoltu: Consider making this `bytes32 => bytes32` and providing function overloads for the different 256-bit types (insertion) and getters that do casting for the user:
```
function add(bytes32 _key, bytes32 _value) ... { items[_key] = _value; ... }
function add(bytes32 _key, address _value) ... { return add(_key, bytes32(_value)); }
function add(bytes32 _key, uint256 _value) ... { return add(_key, bytes32(_value)); }
function add(uint256 _key, address _value) ... { return add(bytes32(_key), bytes32(_value)); }
function getAsAddress(bytes32 _key) ... { return address(items[_key]); }
function getAsAddress(uint256 _key) ... { return address(items[bytes32(_key])); }
```

I recommend initially only providing `bytes32, bytes32` overloads and then adding additional overloads as needed (it looks like in this case you need `bytes32 => address`) since there are so many possible combinations.  This allows us to keep `Map` as a generic/general purpose construct, not tightly coupled to its initial usage (in Market).

nuevoalex: Will do

source/contracts/libraries/collections/Map.sol
```diff
@@ -0,0 +1,65 @@
...
+    function add(bytes32 _key, bytes32 _value) public onlyOwner returns (bool) {
+        if (contains(_key)) {
+            return false;
+        }
+        items[_key] = _value;
+        count += 1;
+        return true;
+    }
+
+    function addAsAddress(bytes32 _key, address _value) public onlyOwner returns (bool) {
```
MicahZoltu: I don't believe there is a need to have this function named differently.  It can just be `add(bytes32,address)` and the compiler will pick the right one.  This will make it easier on the callsite as they can just pass whatever they have and it will "do the right thing".

nuevoalex: That at least didn't seem to work properly in testing. It may just be a problem with pyethereum but given that I opted to make it different. If it is just a problem with pyethereum I can make the test helper method name different but keep the map name the same. Thoughts?

MicahZoltu:
Was there something in particular that made you choose this strategy over the other?  It sounded like you were leaning the other way initially.

source/contracts/libraries/collections/Map.sol
MicahZoltu: Ah, yeah.  Pyethereum may struggle with overloads but I'm reasonably confident that Solidity supports function overloading.  My vote would be to allow for the function overloads and create a helper contract if necessary to address the pyethereum issue (or figure out some other way to deal with the fact that Pyethereum is a dynamic language).

nuevoalex: Will do then!

nuevoalex:
I started out the other way and getting into it I realized I didn't like the table-cloth trick nature of simply checking the universe. This approach feels more dependable and less prone to some strange bug .

source/contracts/reporting/Market.sol
```diff
@@ -245,30 +248,40 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
...
+        require(getReportingState() == ReportingState.AWAITING_FORK_MIGRATION);
+        require(stakeTokens.getCount() > 0);
+        stakeTokens = MapFactory(controller.lookup("MapFactory")).createMap(controller, this);
+        return true;
+    }
+
     function getStakeToken(uint256[] _payoutNumerators, bool _invalid) public returns (IStakeToken) {
         bytes32 _payoutDistributionHash = derivePayoutDistributionHash(_payoutNumerators, _invalid);
-        IStakeToken _stakeToken = stakeTokens[_payoutDistributionHash];
+        IStakeToken _stakeToken = IStakeToken(stakeTokens.getAsAddressOrZero(_payoutDistributionHash));
```
MicahZoltu: I want generics in Solidity so badly.  😢

source/contracts/reporting/Market.sol
```diff
@@ -377,7 +390,7 @@ contract Market is DelegationTarget, Typed, Initializable, Ownable, IMarket {
     }
 
     function getFinalWinningStakeToken() public view returns (IStakeToken) {
-        return stakeTokens[finalPayoutDistributionHash];
+        return IStakeToken(stakeTokens.getAsAddressOrZero(finalPayoutDistributionHash));
```
MicahZoltu: It feels like this function should throw if there is no token at `finalPayoutDistributionHash`.  Should never be possible I suppose, so I guess doesn't really matter...

PR #373, opened 2017-10-16
Fixes a race condition in the contract deployer when calling `Market.getTypeName`

adrake33:


source/libraries/ContractDeployer.ts
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 
+// TODO: Add type mappings for array items
```
MicahZoltu: Is this comment related to binascii or something else?

adrake33: No, this is in reference to the conversation that you, Jack, and I were having yesterday regarding loops where `hasOwnProperty` is not being checked.  I just put it here at the top of the file to serve as a more visible reminder.

PR #374, opened 2017-10-16
Modify Geth Docker to use POA to shorten block time to 1 second

adrake33:
This creates a default account with a large balance that is also designated as being able to seal blocks.  Block time is set to 1 second.

To use the docker container, run:
```
docker build -t geth-dev-node .
docker run --rm -it -p 8545:8545 geth-dev-node -p 8545:8545 ethereum/client-go --dev --ws --wsapi eth,net,web3,personal --wsport 8546 --rpc --rpcapi eth,net,web3,personal,miner --rpcaddr 0.0.0.0 --targetgaslimit 6500000
```


geth-dev-node/Dockerfile
```diff
@@ -1,64 +1,19 @@
...
-RUN geth makedag 0 /app/.ethash
+
+RUN mkdir /home/\.ethereum
+RUN mkdir /home/\.ethereum/keystore
+RUN mkdir /home/\.ethash
+
+COPY genesis.json /home/genesis.json
+RUN printf "\n" > /home/password.txt
+# Sealing account (pre-funded with a large balance)
+COPY UTC--2017-10-16T19-13-42.464934000Z--c5ed899b0878656feb06467e2e9ede3ae73cbcb7 /home/\.ethereum/keystore/UTC--2017-10-16T19-13-42.464934000Z--c5ed899b0878656feb06467e2e9ede3ae73cbcb7
```
MicahZoltu: Recommend renaming this file to something much simpler like `dev-key.json`.  The file names for keys doesn't have any impact on how they are used by Geth.

geth-dev-node/Dockerfile
```diff
@@ -1,64 +1,19 @@
 FROM ethereum/client-go:latest
-COPY genesis.json /app/genesis.json
-RUN echo "p" > /app/password.txt
-RUN geth makedag 0 /app/.ethash
+
+RUN mkdir /home/\.ethereum
+RUN mkdir /home/\.ethereum/keystore
+RUN mkdir /home/\.ethash
```
MicahZoltu: Consider `/geth` instead of `/home` to match `/parity` in the other node and to also make it more clear to the reader what is in the folder.  Also consider not using `.` prefixed folder names since those don't show up in `ls`.

geth-dev-node/UTC--2017-10-16T19-13-42.464934000Z--c5ed899b0878656feb06467e2e9ede3ae73cbcb7
```diff
@@ -0,0 +1 @@
+{"address":"c5ed899b0878656feb06467e2e9ede3ae73cbcb7","crypto":{"cipher":"aes-128-ctr","ciphertext":"4c8da4df20e90fa044d1679c8e29dffb4d5cdf126770dccddfb3136da5ce74ee","cipherparams":{"iv":"0fcfef299dd9faa41163ff32180e17f5"},"kdf":"scrypt","kdfparams":{"dklen":32,"n":262144,"p":1,"r":8,"salt":"951f29c628ef1f71c25732b43c14fae581928d68bc9f944b6878e10c7656cf32"},"mac":"854cfdebf1941d5c0ec9ebc5296c796a3254029f13ab57a4218ee200e47be9c7"},"id":"97c0f84f-bc0f-4e06-a466-67716ca4a9aa","version":3}
```
MicahZoltu: In VSCode you can set the filetype to JSON and then open the command pallet and hit "Format Document".  This will pretty print this key file as JSON which will make it easier to read and diff.

geth-dev-node/genesis.json
```diff
@@ -1,17 +1,798 @@
...
+    },
+    "00000000000000000000000000000000000000fd": {
+      "balance": "0x1"
+    },
+    "00000000000000000000000000000000000000fe": {
+      "balance": "0x1"
+    },
+    "00000000000000000000000000000000000000ff": {
+      "balance": "0x1"
+    },
```
MicahZoltu: Consider doing a find-and-replace and making all of these accounts one-line rather than 3 lines.  This would just be to shrink the file a bit and make it easier to read.

Also, do we really need 0xff worth of these accounts with a balance of 1?  I suspect we could delete all but the first 10-20 and be fine.

geth-dev-node/start.sh
```diff
@@ -1,3 +1,2 @@
-geth --password /app/password.txt account new
-geth init /app/genesis.json
-geth --password /app/password.txt --unlock 0 --mine --networkid=10000 --ws --wsapi eth,net,web3,personal --wsport 8546 --rpc --rpcapi eth,net,web3,personal,miner --rpcaddr 0.0.0.0 --targetgaslimit 6700000 --ethash.dagdir "/app/.ethash"
+geth --datadir /home/\.ethereum init /home/genesis.json
```
MicahZoltu: This step should be moved to the Dockerfile as:
```
RUN geth --datadir /home/\.ethereum init /home/genesis.json
```

geth-dev-node/start.sh
```diff
@@ -1,3 +1,2 @@
...
+geth --datadir /home/\.ethereum --keystore /home/\.ethereum/keystore --password /home/password.txt --unlock "0xc5ed899b0878656feb06467e2e9ede3ae73cbcb7" --mine --ws --wsapi eth,net,web3,personal --wsport 8546 --rpc --rpcapi eth,net,web3,personal,miner --rpcaddr 0.0.0.0 --targetgaslimit 6500000
```
MicahZoltu: Parity lets you have a config file rather than configuring Parity exclusively via CLI arguments.  Does Geth have something similar or is CLI arguments the only way to configure it?

MicahZoltu: Is `--mine` still necessary even though this is a PoA network?

geth-dev-node/Dockerfile
adrake33: Yeah, I don't like this naming convention, but I tried renaming them, and for some reason the accounts weren't getting unlocked.  I guess maybe something else was preventing the unlocking.  I'll check on this.

geth-dev-node/Dockerfile
adrake33: Sounds good.  I'll make this change.

geth-dev-node/genesis.json
adrake33: Yeah, this was how puppeth generated the file.  I wasn't sure if these accounts were for anything in particular, and that's how one of the clique files you gave me looked, so I left them in.  I'll remove some of them and reformat these lines.

geth-dev-node/start.sh
adrake33: I'm not sure -- I'll check on that.  I only did it this was because this was the only way I'd seen it done elsewhere.

adrake33: Good question.  I'll try taking it out, and if everything still works, I'll leave it out.

adrake33: Yep, looks like it is necessary.  Nothing happens after unlocking the account without `--mine`.

MicahZoltu: Interesting.  It must use `--mine` to mean `validate` when running in Proof of Authority mode.

PR #375, opened 2017-10-17
Adds Parity PoA chain mode.

MicahZoltu:
Using PoA chain requires changing `start.sh` to point at the other toml file.  This can be improved in the future via a smarter scirpt and docker COMMAND arguments.

PoA has 1 second block time, which allows testing of block timing delays.

Also did some cleanup/minor quality of life improvements to contract deployer library.

### DIFF
--- a/parity-dev-node/aura-chain-spec.json
+++ b/parity-dev-node/aura-chain-spec.json
@@ -1,0 +1,34 @@
+{
+    "name": "AuraChain",
+    "engine": {
+        "authorityRound": {
+            "params": {
+                "stepDuration": "1",
+                "validators" : { "list": [ "0x913da4198e6be1d5f5e4a40d0667f70c0b5430eb" ] }
+            }
+        }
+    },
+    "params": {
+        "gasLimitBoundDivisor": "0x400",
+        "maximumExtraDataSize": "0x20",
+        "minGasLimit": "0x1",
+        "networkID" : "0x2323"
+    },
+    "genesis": {
+        "seal": {
+            "authorityRound": {
+                "step": "0x0",
+                "signature": "0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
+            }
+        },
+        "difficulty": "0x20000",
+        "gasLimit": "0x632EA0"
+    },
+    "accounts": {
+        "0x0000000000000000000000000000000000000001": { "balance": "1", "builtin": { "name": "ecrecover", "pricing": { "linear": { "base": 3000, "word": 0 } } } },
+        "0x0000000000000000000000000000000000000002": { "balance": "1", "builtin": { "name": "sha256", "pricing": { "linear": { "base": 60, "word": 12 } } } },
+        "0x0000000000000000000000000000000000000003": { "balance": "1", "builtin": { "name": "ripemd160", "pricing": { "linear": { "base": 600, "word": 120 } } } },
+        "0x0000000000000000000000000000000000000004": { "balance": "1", "builtin": { "name": "identity", "pricing": { "linear": { "base": 15, "word": 3 } } } },
+        "0x913da4198e6be1d5f5e4a40d0667f70c0b5430eb": { "balance": "0xffffffffffffffffffffffffffffffff"}
+    }
+}
--- a/parity-dev-node/start.sh
+++ b/parity-dev-node/start.sh
@@ -3,17 +3,12 @@
 # make it so we can do job control inside the script (fg at the end)
 set -m
 
-# launch parity in the foreground
-/parity/parity \
-	--chain dev \
-	--gasprice 2 \
-	--no-discovery \
-	--force-ui --ui-no-validation --ui-interface 0.0.0.0 \
-	--jsonrpc-interface all --jsonrpc-cors "*" --jsonrpc-hosts all --jsonrpc-apis web3,eth,net,personal,parity,parity_set,traces,rpc,parity_accounts \
-	--unlock 0x00a329c0648769a73afac7f9381e08fb43dbea72 --password /parity/password \
-	&
+# launch parity in the background
+/parity/parity --config /parity/instant-seal-config.toml --gasprice 1 &
 
 # spin until node is connectable
 while ! curl --silent --show-error localhost:8545 -H "Content-Type: application/json" -X POST --data '{"jsonrpc":"2.0","method":"net_version","id": 1}'; do sleep 0.1; done
-curl --silent --show-error localhost:8545 -H "Content-Type: application/json" -X POST --data '{"jsonrpc":"2.0","method":"eth_sendTransaction","params":[{"value":"0x0","to":"0x0000000000000000000000000000000000000000","from":"0x00a329c0648769a73afac7f9381e08fb43dbea72","data":"0x","gasPrice":"0x3B9ACA00"}], "id": 1}'
+curl --silent --show-error localhost:8545 -H "Content-Type: application/json" -X POST --data '{"jsonrpc":"2.0","method":"eth_sendTransaction","params":[{"value":"0x0","to":"0x0000000000000000000000000000000000000000","from":"0x913da4198e6be1d5f5e4a40d0667f70c0b5430eb","data":"0x","gasPrice":"0x3B9ACA00"}], "id": 1}'
+
+# bring parity to the foreground
 fg
--- a/source/deployment/deployContracts.ts
+++ b/source/deployment/deployContracts.ts
@@ -3,8 +3,8 @@
 import * as fs from "async-file";
 import * as getPort from "get-port";
 import * as path from "path";
-import * as EthjsHttpProvider from "ethjs-provider-http";
-import * as EthjsQuery from "ethjs-query";
+import EthjsHttpProvider = require('ethjs-provider-http');
+import EthjsQuery = require("ethjs-query");
 import { SolidityContractCompiler } from "../libraries/CompileSolidity";
 import { ContractDeployer } from "../libraries/ContractDeployer";
 import { initializeTestRpcClientOptions } from "../libraries/HelperFunctions";
--- a/source/libraries/AbiParser.ts
+++ b/source/libraries/AbiParser.ts
@@ -1,7 +1,6 @@
 import * as EthjsAbi from 'ethjs-abi';
 import * as EthjsQuery from 'ethjs-query';
 import { CompilerOutputContractAbiFunction, CompilerOutputContractAbiEvent } from 'solc';
-import { waitForTransactionToBeSealed } from './HelperFunctions';
 
 interface TransactionOptions {
     to?: string,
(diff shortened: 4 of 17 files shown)


source/libraries/ContractDeployer.ts
```diff
@@ -3,13 +3,14 @@
...
-import * as EthjsQuery from "ethjs-query";
+import EthjsAbi = require("ethjs-abi");
+import EthjsContract = require("ethjs-contract");
+import EthjsQuery = require("ethjs-query");
+import { TransactionReceipt } from 'ethjs-shared';
 // TODO: Update TS type definition for ContractBlockchainData to allow for empty object (e.g. upload() & uploadAndAddToController())?
 import { ContractBlockchainData, ContractReceipt } from "contract-deployment";
 import { Contract, parseAbiIntoMethods } from "./AbiParser";
-import { generateTestAccounts, padAndHexlify, stringTo32ByteHex, waitForTransactionToBeSealed } from "./HelperFunctions";
+import { generateTestAccounts, padAndHexlify, stringTo32ByteHex, waitForTransactionReceipt, sleep } from "./HelperFunctions";
```
adrake33: Sleep can be removed here.

source/libraries/ContractDeployer.ts
```diff
@@ -140,64 +143,69 @@ export class ContractDeployer {
...
                     // this.contracts[contractName] = this.applySignature(contractName, this.contracts[contractName].address)
                 } else {
-                    uploadedContractPromises.push(this.uploadAndAddToController(contractFileName));
+                    promises.push(this.uploadAndAddToController(contractFileName));
                 }
             }
         }
 
-        await Promise.all(uploadedContractPromises);
+        const latestBlock = await this.ethjsQuery.getBlockByNumber('latest', true);
```
adrake33: Why did you add this line?

source/libraries/ContractDeployer.ts
```diff
@@ -249,7 +256,6 @@ export class ContractDeployer {
             throw new Error("Unable to get address for new categorical market.");
         }
         const createMarketHash = await targetReportingWindow.createMarket.bind({ value: marketCreationFee })(endTime, numOutcomes, numTicks, feePerEthInWei, denominationToken, designatedReporter);
-        waitForTransactionToBeSealed(this.ethjsQuery, createMarketHash);
```
adrake33: We don't need to wait before calling `market.getTypeName`?

source/libraries/HelperFunctions.ts
```diff
@@ -74,14 +75,15 @@ export async function initializeTestRpcClientOptions(ethHttpProviderPort: number
     return { gasLimit: GAS_BLOCK_AMOUNT, accounts: accountOptions };
 }
 
-async function sleep(milliseconds: number): Promise<object> {
-    return new Promise((resolve, reject) => setTimeout(resolve, milliseconds));
+export async function sleep(milliseconds: number): Promise<void> {
```
adrake33: Do we actually need to export this?  I didn't see where it's being used outside of this file.

typings/ethjs-shared.d.ts
```diff
@@ -0,0 +1,82 @@
...
+        logIndex: BN;
+        transactionIndex: BN;
+        transactionHash: string;
+        blockHash: string;
+        blockNumber: BN;
+        address: string;
+        data: string;
+        topics: Array<string>;
+    }
+}
```
adrake33: Nice, thanks for adding these type definition files!

parity-dev-node/aura-chain-spec.json
```diff
@@ -0,0 +1,34 @@
+{
+    "name": "AuraChain",
```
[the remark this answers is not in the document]
MicahZoltu: Aura is one of the Parity Proof of Authority algorithms, like Clique is for Geth.  Parity also has Tendermint Proof of Authority I believe.

parity-dev-node/start.sh
```diff
@@ -4,16 +4,9 @@
...
 # launch parity in the foreground
-/parity/parity \
-	--chain dev \
-	--gasprice 2 \
-	--no-discovery \
-	--force-ui --ui-no-validation --ui-interface 0.0.0.0 \
-	--jsonrpc-interface all --jsonrpc-cors "*" --jsonrpc-hosts all --jsonrpc-apis web3,eth,net,personal,parity,parity_set,traces,rpc,parity_accounts \
-	--unlock 0x00a329c0648769a73afac7f9381e08fb43dbea72 --password /parity/password \
-	&
+/parity/parity --config /parity/instant-seal-config.toml --gasprice 1 &
```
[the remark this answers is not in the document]
MicahZoltu: The trailing ampersand is necessary to launch Parity in the background, though your comment makes me realize that the comment above this line is wrong.  :)

parity-dev-node/aura-chain-spec.json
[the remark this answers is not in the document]
adrake33: Ah, that's why I hadn't seen it before.  Ok, good to know.

parity-dev-node/start.sh
```diff
@@ -4,16 +4,9 @@
...
 
 # spin until node is connectable
 while ! curl --silent --show-error localhost:8545 -H "Content-Type: application/json" -X POST --data '{"jsonrpc":"2.0","method":"net_version","id": 1}'; do sleep 0.1; done
-curl --silent --show-error localhost:8545 -H "Content-Type: application/json" -X POST --data '{"jsonrpc":"2.0","method":"eth_sendTransaction","params":[{"value":"0x0","to":"0x0000000000000000000000000000000000000000","from":"0x00a329c0648769a73afac7f9381e08fb43dbea72","data":"0x","gasPrice":"0x3B9ACA00"}], "id": 1}'
+curl --silent --show-error localhost:8545 -H "Content-Type: application/json" -X POST --data '{"jsonrpc":"2.0","method":"eth_sendTransaction","params":[{"value":"0x0","to":"0x0000000000000000000000000000000000000000","from":"0x913da4198e6be1d5f5e4a40d0667f70c0b5430eb","data":"0x","gasPrice":"0x3B9ACA00"}], "id": 1}'
 fg
```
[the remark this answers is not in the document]
MicahZoltu: We want Parity to eventually run in the foreground, but we also want to run the while loop and then send that transaction once it starts up.  To achieve that, we start Parity in the background, then run the curls that we need, then when they are done we bring Parity back to the foreground.

source/deployment/deployContracts.ts
```diff
@@ -3,8 +3,8 @@
 import * as fs from "async-file";
 import * as getPort from "get-port";
 import * as path from "path";
-import * as EthjsHttpProvider from "ethjs-provider-http";
-import * as EthjsQuery from "ethjs-query";
+import EthjsHttpProvider = require('ethjs-provider-http');
+import EthjsQuery = require("ethjs-query");
```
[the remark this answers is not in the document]
MicahZoltu: It turns out `import * as EthjsQuery from 'ethjs-query';` is not correct and only works by happenstance and will likely break in the future.  Fixing these up was just some code cleanup.  Note, this style is only necessary for imports that you need to `new` directly.  For an import that is just a property bag, the other style is still appropriate (as far as I understand).

source/libraries/AbiParser.ts
```diff
@@ -1,7 +1,7 @@
 import * as EthjsAbi from 'ethjs-abi';
 import * as EthjsQuery from 'ethjs-query';
 import { CompilerOutputContractAbiFunction, CompilerOutputContractAbiEvent } from 'solc';
-import { waitForTransactionToBeSealed } from './HelperFunctions';
+import { waitForTransactionReceipt } from './HelperFunctions';
```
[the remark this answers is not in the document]
MicahZoltu: Fixed.

source/libraries/ContractDeployer.ts

source/libraries/ContractDeployer.ts
MicahZoltu: Oops, debugging.  Will remove.

parity-dev-node/start.sh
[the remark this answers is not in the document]
adrake33: Oops, yeah, I didn't catch that the ampersand was also there before the change.  Never mind.

source/libraries/ContractDeployer.ts
MicahZoltu: This call was missing the `await` so it wasn't actually waiting.  This line was effectively a no-op, so I removed it.  We _may_ need to wait here, and I'll be digging into the market creation issue tomorrow to find out.  I removed this because at the moment, it wasn't doing anything but clogging up the intertubes.  😄

parity-dev-node/start.sh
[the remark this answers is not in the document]
adrake33: Ah, I see now.  Ok, that makes more sense.

source/libraries/HelperFunctions.ts
MicahZoltu: In the past I have found that having an async sleep can be really useful when debugging, so even though we may only use it directly here, having it exported can be really useful.  Since this is in a `HelperFunctions` file I made an exception to my usual personal rule of "only export things you actually need to export", since it is so useful for debugging.

source/deployment/deployContracts.ts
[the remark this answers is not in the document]
adrake33: That's weird, why does it actually work if it's not correct?

source/libraries/ContractDeployer.ts
adrake33: Haha, oops, good catch!  I was probably sleep-deprived when I added that line. :P

source/deployment/deployContracts.ts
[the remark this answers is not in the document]
MicahZoltu: Apparently just because of the way TypeScript compiler happens to be implemented makes it so it happens to be valid and they haven't bother to write the code that will trigger an error when someone does this incorrectly.  This Q&A talks a bit about the situation: https://stackoverflow.com/questions/39415661/what-does-resolves-to-a-non-module-entity-and-cannot-be-imported-using-this

source/libraries/HelperFunctions.ts
```diff
@@ -74,14 +75,15 @@ export async function initializeTestRpcClientOptions(ethHttpProviderPort: number
...
 
-export async function waitForTransactionToBeSealed(ethjsQuery: EthjsQuery, transactionHash: string): Promise<void> {
-    let transaction = await ethjsQuery.getTransactionByHash(transactionHash);
-    while (transaction.blockNumber == null) {
+export async function waitForTransactionReceipt(ethjsQuery: EthjsQuery, transactionHash: string): Promise<TransactionReceipt> {
+    let transactionReceipt = await ethjsQuery.getTransactionReceipt(transactionHash);
+    while (!transactionReceipt) {
         await sleep(ETHEREUM_POLLING_INTERVAL_MILLISECONDS);
-        transaction = await ethjsQuery.getTransactionByHash(transactionHash);
+        transactionReceipt = await ethjsQuery.getTransactionReceipt(transactionHash);
```
epheph: Is what we really want to wait the transaction receipt, or for it to be mined? Since we've seen parity give receipts for un-mined TX. The contract_address will be correct, but an eth_call() won't have the data available to it yet

MicahZoltu: Ah, good catch.  I meant to wait for the transaction receipt to have a block.  Will fix in next PR.

PR #376, opened 2017-10-17
#3936 3 reporting token tests

bthaile:
created branch before reporting token was renamed to stake token
* Create tests for stake token (still ongoing)
* Create mocks to help isolate logic to only staketoken
* Added a TODO comment to finish up a test when redeemDisavowedTokens is fixed


bthaile:
Take a look at my approach with mocks. I'd like to get as close to actual unit tests as possible. Seems there so much setup just to test simple methods.

tests/solidity_test_helpers/MockUniverse.sol
```diff
@@ -0,0 +1,315 @@
...
+            // FXP formula: previous_amount * actual_percent / (2 * target_percent) + 0.5;
+            _newValue = _badMarkets.mul(_previousValue).mul(_targetDivisor).div(_totalMarkets).div(2)  + _previousValue / 2; // FIXME: This is on one line due to solium bugs
+        } else {
+            // FXP formula: previous_amount * (1/(1 - target_percent)) * (actual_percent - target_percent) + 1;
+            _newValue = _targetDivisor.mul(_previousValue.mul(_badMarkets).div(_totalMarkets).sub(_previousValue.div(_targetDivisor))).div(_targetDivisor - 1) + _previousValue; // FIXME: This is on one line due to a solium bug
+        }
+
+        if (_newValue < _floor) {
+            _newValue = _floor;
+        }
```
MicahZoltu: Why not just have this return a previously set value like all of the other getters?

bthaile: I had an issue with the "pure" compile/upload contract failed. So I just copied the function from universe.

tests/solidity_test_helpers/MockUniverse.sol
```diff
@@ -0,0 +1,315 @@
...
+    }
+    /*
+    * Impl of IUniverse and ITyped
+     */
+    function getTypeName() public view returns (bytes32) {
+        return "Universe";
+    }
+    function initialize(IUniverse _parentUniverse, bytes32 _parentPayoutDistributionHash) external returns (bool) {
+        setParentUniverseValue = _parentUniverse;
+        setParentPayoutDistributionHashValue = _parentPayoutDistributionHash;
```
MicahZoltu: Consider having these be assigned a variable like `providedParentUniverseValue` to distinguish them from values that are set by the person mocking, rather than set by the contract under test calling this method.

bthaile: I'm leaning towards ignore the input values and just have the setters used. thoughts?

tests/solidity_test_helpers/MockUniverse.sol
MicahZoltu: Ah, yeah.  If this method is marked as pure the you would have to hard-code a value.  This is fine.

tests/solidity_test_helpers/MockUniverse.sol
MicahZoltu: Also, the variables for these should be public so the tests can read them to assert that the caller provided the expected parameters to the `initialize` call.

bthaile: ok on verify input for initialize method, I also want to add a mech to verify a method was called. I'll need to look up a good pattern in python. I'll add that functionality.

tests/libraries/token/test_stake_token.py
```diff
@@ -0,0 +1,325 @@
...
+from pytest import fixture, raises
+from ethereum.tools.tester import ABIContract, TransactionFailed
+from reporting_utils import proceedToDesignatedReporting, proceedToRound1Reporting, proceedToRound2Reporting, proceedToForking, finalizeForkingMarket, initializeReportingFixture
+
+def test_stake_token_creation_binary(tokenFixture, binaryMarket):
+    bin_market = binaryMarket
+    numTicks = 10 ** 18
+    stakeToken_bin = tokenFixture.upload('../source/contracts/reporting/StakeToken.sol', 'stakeToken')
+    # stake token needs same payout numerators as market
+    with raises(TransactionFailed, message="payout numerators can not be more than market"):
```
MicahZoltu: Style: Recommend either including a comment above these raise clauses or including a message.  Having both the comment and the message ends up being a bit redundant.

tests/libraries/token/test_stake_token.py
```diff
@@ -0,0 +1,325 @@
...
+    # Proceed to the FORKING phase
+    proceedToForking(tokenFixture, universe, market_forking, True, tester.k1, tester.k3, tester.k4, [0,numTicks], [numTicks,0], tester.k2, [numTicks,0], [numTicks,0], [0,numTicks])
+    assert market_forking.getReportingState() == tokenFixture.contracts['Constants'].FORKING()
+
+    # during a fork non-forking markets should be in limbo and allow users to redeemDisavowedTokens
+    assert market_other.getReportingState() == tokenFixture.contracts['Constants'].AWAITING_FORK_MIGRATION()
+    stakeToken_non_forking = tokenFixture.applySignature('StakeToken', market_other.getStakeToken([0, numTicks]))
+    assert market_other.isContainerForStakeToken(stakeToken_non_forking.address) == True
+
+    # // :TODO Finish up test when redeem has been fixed
```
MicahZoltu: @nuevoalex Is redeem fully working now?

nuevoalex: I'm working on finishing it up now. That said I'm also writing tests that probably cover what the TODO is hoping to, so I'd say this TODO could just be removed :)

tests/libraries/token/test_stake_token.py
```diff
@@ -0,0 +1,325 @@
...
+    universe = ABIContract(fixture.chain, kitchenSinkSnapshot['universe'].translator, kitchenSinkSnapshot['universe'].address)
+    return universe
+
+@fixture
+def binaryMarket(fixture, kitchenSinkSnapshot):
+    market = ABIContract(fixture.chain, kitchenSinkSnapshot['binaryMarket'].translator, kitchenSinkSnapshot['binaryMarket'].address)
+    return market
+
+@fixture
+def cash(fixture, kitchenSinkSnapshot):
```
MicahZoltu: These 4 fixtures (universe, binaryMarket, categorigcalMarket, cash) should take in `tokenFixture` as the first parameter, not fixture.  This will guarantee that they are called _after_ `tokenFixture`, which is necessary to ensure that the ABI is mapped correctly after the fixture was last reset.

tests/libraries/token/test_stake_token.py
```diff
@@ -0,0 +1,325 @@
...
+    return market
+
+@fixture
+def tokenFixture(fixture, universe, localSnapshot):
+    fixture.resetToSnapshot(localSnapshot)
+    value = 1 * 10**6 * 10**18
+    reputationToken = fixture.applySignature('ReputationToken', universe.getReputationToken())
+    # Give tester reputation tokens
+    for testAccount in [tester.a4, tester.a5, tester.a6, tester.a7, tester.a8, tester.a9]:
+        reputationToken.transfer(testAccount, value)
```
MicahZoltu: This should be moved to `localSnapshot` rather than here.  By having it here, this code is re-run before every test which increases time required to run the tests since transaction execution is slow.  Glancing at this function, I think most/all of this logic could be moved to `localSnapshot` instead of here, which will also allow you to remove the `universe` dependency on this fixture which will make resolving the issue I mentioned above about ordering easier to address.

I consider `Controlled.sol` to be a good example of local fixture setup, you may want to check it out as a reference, or just ping me in #dev if none of this is making sense.  😄

tests/libraries/token/test_stake_token.py
```diff
@@ -0,0 +1,325 @@
...
+    cash = ABIContract(fixture.chain, kitchenSinkSnapshot['cash'].translator, kitchenSinkSnapshot['cash'].address)
+    return cash
+
+@fixture
+def categoricalMarket(fixture, kitchenSinkSnapshot):
+    market = ABIContract(fixture.chain, kitchenSinkSnapshot['categoricalMarket'].translator, kitchenSinkSnapshot['categoricalMarket'].address)
+    return market
+
+@fixture
+def tokenFixture(fixture, universe, localSnapshot):
```
MicahZoltu: Style: I have been naming this `localFixture` in other test files.  This makes it so I don't have to remember what I named the file-local fixture for each test file.  😄

tests/solidity_test_helpers/MockUniverse.sol
MicahZoltu: That is probably reasonable, since this method shouldn't ever be called during testing.  Perhaps just have this method `assert(false)` on the first line to ensure it isn't called?

PR #377, opened 2017-10-17
Adds support for parity integration tests with either fixed block time or instant seal.

MicahZoltu:
Speeds up iteration time when working on test infrastructure by enhancing what Docker images cash.

### DIFF
--- a/Dockerfile-integration-tests
+++ b/Dockerfile-integration-tests
@@ -6,8 +6,13 @@ WORKDIR /app
 RUN npm install
 COPY tsconfig.json /app/tsconfig.json
 COPY typings/ /app/typings/
-COPY source/ /app/source/
+COPY source/contracts/ /app/source/contracts/
+COPY source/deployment/CompileContracts.ts /app/source/deployment/CompileContracts.ts
+COPY source/libraries/CompileSolidity.ts /app/source/libraries/CompileSolidity.ts
 RUN npx tsc
 RUN node ./output/deployment/CompileContracts.js
+
+COPY source/ /app/source/
+RUN npx tsc
 
 ENTRYPOINT [ "npx", "mocha", "output/tests-integration/**/*.js", "--no-timeouts", "--require", "source-map-support/register" ]
--- a/parity-dev-node/Dockerfile
+++ b/parity-dev-node/Dockerfile
@@ -16,7 +16,7 @@ COPY start.sh /start.sh
 RUN echo "" > /parity/password
 RUN chmod +x /start.sh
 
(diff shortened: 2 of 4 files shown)


parity-dev-node/Dockerfile
```diff
@@ -16,7 +16,7 @@ COPY start.sh /start.sh
 RUN echo "" > /parity/password
 RUN chmod +x /start.sh
 
-ENTRYPOINT /start.sh
+ENTRYPOINT [ "/start.sh" ]
```
adrake33: Why did you add the brackets?

Dockerfile-integration-tests
```diff
@@ -6,8 +6,13 @@ WORKDIR /app
 RUN npm install
 COPY tsconfig.json /app/tsconfig.json
 COPY typings/ /app/typings/
-COPY source/ /app/source/
+COPY source/contracts/ /app/source/contracts/
+COPY source/deployment/CompileContracts.ts /app/source/deployment/CompileContracts.ts
+COPY source/libraries/CompileSolidity.ts /app/source/libraries/CompileSolidity.ts
```
adrake33: Why are you copying only these files and then copying everything in source later on?

MicahZoltu: We only need contracts and these two files to compile the contracts and store the compiled results.  Docker will cache this and only recompile all of the contracts if one of the files copied above the contract compilation line changes.  This allows us to iterate on the integration tests without having to re-compile all of the contracts every run.

parity-dev-node/Dockerfile
MicahZoltu: Because Docker is stupid.  😢 `ENTRYPOINT /start.sh` is notably different from `ENTRYPOINT [ "/start.sh" ]`.  In this particular case, it needs to be the latter format to support passing in additional variables to the script and docker run time.

Dockerfile-integration-tests
adrake33: Ok, I figured it had to do with caching the contracts.  So Docker's cache works on a line-by-line basis?

MicahZoltu: Yeah, each Docker line is another cache layer.  If you do something like this though, its considered one line:
```
RUN echo hello \
    && echo world
```

adrake33: Yeah, mean I've seen it cache certain parts of the file that it's already run.  It just hadn't occurred to me that we could do the caching in this way so that if non-contract files are modified, it won't recompile.  Good idea!

PR #378, opened 2017-10-17
Changes waiting for receipts to wait for full receipt.

MicahZoltu:
* Removes gasAmount on a number of calls (so estimation is used instead).
* Adds gasPrice on some calls that were missing it.
* Removes async on `parseAbiIntoMethods` to make it clear the function returns immediately.

### DIFF
--- a/source/libraries/ContractDeployer.ts
+++ b/source/libraries/ContractDeployer.ts
@@ -202,8 +202,8 @@ export class ContractDeployer {
     }
 
     private async createGenesisUniverse(): Promise<ContractBlockchainData> {
-        const delegatorBuilder = this.ethjsContract(this.signatures["Delegator"], this.bytecodes["Delegator"], { from: this.testAccounts[0], gas: this.gasAmount });
-        const universeBuilder = this.ethjsContract(this.signatures["Universe"], this.bytecodes["Universe"], { from: this.testAccounts[0], gas: this.gasAmount });
+        const delegatorBuilder = this.ethjsContract(this.signatures["Delegator"], this.bytecodes["Delegator"], { from: this.testAccounts[0], gasPrice: this.gasPrice });
+        const universeBuilder = this.ethjsContract(this.signatures["Universe"], this.bytecodes["Universe"], { from: this.testAccounts[0], gasPrice: this.gasPrice });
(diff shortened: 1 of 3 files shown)


epheph:
👍

MicahZoltu:
@epheph `parseAbiMethods` really is a factory that returns methods that submit contract calls.  When you call one of the returned anonymous methods is when the call is made.  I'm open to suggestions on a better name for the factory, I also don't like it.

epheph:
Oh, i missed that it was a factory for it, I see that now, that's not too bad then.

MicahZoltu:
If to caller specifies gas, we use it.  We only use the estimator if no gas in supplied (here are a couple different ways to supply gas).  Also, if it is a constant call, gas isn't required so we don't calculate gas in that case.

source/libraries/ContractDeployer.ts
```diff
@@ -229,10 +229,11 @@ export class ContractDeployer {
...
 
-        const universe = await parseAbiIntoMethods(this.ethjsQuery, this.signatures["Universe"], { to: universeAddress, from: this.testAccounts[0], gas: this.gasAmount, gasPrice: this.gasPrice });
-        const legacyReputationToken = await parseAbiIntoMethods(this.ethjsQuery, this.signatures['LegacyRepContract'], { to: this.contracts['LegacyRepContract'].address, from: this.testAccounts[0], gas: this.gasAmount, gasPrice: this.gasPrice });
+        const universe = parseAbiIntoMethods(this.ethjsQuery, this.signatures["Universe"], { to: universeAddress, from: this.testAccounts[0], gasPrice: this.gasPrice });
+        const legacyReputationToken = parseAbiIntoMethods(this.ethjsQuery, this.signatures['LegacyRepContract'], { to: this.contracts['LegacyRepContract'].address, from: this.testAccounts[0], gasPrice: this.gasPrice });
         const reputationTokenAddress = await universe.getReputationToken();
-        const reputationToken = await parseAbiIntoMethods(this.ethjsQuery, this.signatures['ReputationToken'], { to: reputationTokenAddress, from: this.testAccounts[0], gas: this.gasAmount, gasPrice: this.gasPrice });
+        console.log(reputationTokenAddress);
```
adrake33: Should remove this debugging line.

PR #379, opened 2017-10-18
Rep holder fee fairness

nuevoalex:
Another giant PR so no speed expected w/ regard to the review.

I also deleted some tests or TODOs for testing in favor of writing a specific test file for fees.


source/contracts/factories/WindowParticipationTokenFactory.sol
```diff
@@ -0,0 +1,17 @@
+pragma solidity 0.4.17;
+
+
+import 'libraries/Delegator.sol';
+import 'reporting/IWindowParticipationToken.sol';
+import 'reporting/IReportingWindow.sol';
+import 'IController.sol';
+
+
+contract WindowParticipationTokenFactory {
```
MicahZoltu: Consider: Rename `Window*` to `ReportingWindow*`.  I know it makes these things get a bit long, but `Reporting` is a fairly important piece of information in the name IMO.

source/contracts/reporting/DisputeBondToken.sol
```diff
@@ -32,13 +32,17 @@ contract DisputeBondToken is DelegationTarget, ITyped, Initializable, ERC20Basic
...
-    function withdraw() public returns (bool) {
+    function withdraw(bool forgoFees) public returns (bool) {
         require(msg.sender == bondHolder);
         bool _isFinalized = market.getReportingState() == IMarket.ReportingState.FINALIZED;
         require(!market.isContainerForDisputeBondToken(this) || (_isFinalized && market.getFinalPayoutDistributionHash() != disputedPayoutDistributionHash));
         require(getUniverse().getForkingMarket() != market);
         IReputationToken _reputationToken = getReputationToken();
         uint256 _amountToTransfer = _reputationToken.balanceOf(this);
+        if (bondRemainingToBePaidOut > bondAmount) {
+            uint256 _amountToCollectFeesOn = _amountToTransfer.min(bondRemainingToBePaidOut - bondAmount);
```
MicahZoltu: `bondRemainingToBePaidOut - bondAmount` could be negative, which will underflow and result in a really big number.  I know we are doing a `min` here, but are we worried about this underflow?  If not, we should put a comment in explaining why its fine.

nuevoalex: This should be impossible due to the if condition it is in. I will add a comment though too

MicahZoltu: Oh, oops.  Maybe I should sleep before reviewing this PR...

MicahZoltu: Comment is probably not necessary, I should have seen that if statement.

source/contracts/reporting/StakeToken.sol
```diff
@@ -104,16 +105,12 @@ contract StakeToken is DelegationTarget, ITyped, Initializable, VariableSupplyTo
 
     // NOTE: UI should warn users about calling this before first calling `migrateLosingTokens` on all losing tokens with non-dust contents
     // NOTE: we aren't using the convertToAndFromCash modifier here becuase this isn't a whitelisted contract. We expect the reporting window to handle disbursment of ETH
-    // CONSIDER: If all markets are finalized should we allow forgoing fees?
+    // CONSIDER: If all markets are finalized and the reporting window is over should we allow forgoing fees?
```
MicahZoltu: I would say no, though only if it is simple to implement/reduces complexity.

nuevoalex:
I'll wait till tomorrow late morning to give some time to look over again if you see fit. I'm mostly motivated by wanting to write the escape hatch code that is dependent on this more than wanting to push forward with the code though, so if its still approved when I merge it also feel free to just look at stuff post merge

source/contracts/factories/WindowParticipationTokenFactory.sol
nuevoalex: I was going to do this but it puts us over the `string` limit, so for the sake of that I'll keep it the same

MicahZoltu: 😡

MicahZoltu: What about `ReportingParticipationToken` in that case?  I feel like if we have to cut `Reporting` or `Window` I would rather cut `Window`.  However, I feel like there was a reason we didn't go with that name in the first place?

nuevoalex: There was some discussion but nothing too important. `ReportingParticipationToken` sounds good!

nuevoalex: Heh, this also puts it over the character limit.

PR #380, opened 2017-10-18
Rep Price Oracle

nuevoalex:


source/contracts/reporting/RepPriceOracle.sol
```diff
@@ -0,0 +1,18 @@
+pragma solidity 0.4.17;
+
+import 'libraries/Ownable.sol';
+import 'reporting/IRepPriceOracle.sol';
+
+
+contract RepPriceOracle is Ownable, IRepPriceOracle {
+    
+    uint256 private repPriceInAttoEth = 11 * 10 ** 18;
```
MicahZoltu: This default is _way_ off.  This is the total REP supply, not the price of 1 REP measure in attoeth.  Current price of REP in ETH is `0.05822872`, so price of REP in attoeth would be `58228720000000000` or so.  Lets use that as the "reasonable default".

PR #382, opened 2017-10-18
Add explicit tests for contract existance checking and turn into a lib

nuevoalex:
Based on the conversation in Slack I'm not sure this will address the problem y'all are looking at but this is a good change I think either way.

### DIFF
--- a/source/contracts/libraries/ContractExists.sol
+++ b/source/contracts/libraries/ContractExists.sol
@@ -1,0 +1,10 @@
+pragma solidity 0.4.17;
+
+
+library ContractExists {
+    function exists(address _address) internal view returns (bool) {
+        uint256 size;
+        assembly { size := extcodesize(_address) }
+        return size > 0;
+    }
+}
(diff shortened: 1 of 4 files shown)


source/contracts/libraries/ContractExists.sol
```diff
@@ -0,0 +1,10 @@
+pragma solidity 0.4.17;
+
+
+library ContractExists {
+    function exists(address _address) internal view returns (bool) {
+        uint256 size;
+        assembly { size := extcodesize(_address) }
```
adrake33: I don't think this will work when the contract deployer calls `estimateGas`.  The error I was seeing earlier today was that it was saying this line was an invalid op code when passing in LegacyRepContract to `estimateGas`.  If you do `let size = ...`, it doesn't complain, so I think the logic related to `size` has to be inside the assembly block.

adrake33: I don't think size is accessible outside of the assembly block if you do `let size = ...`: http://solidity.readthedocs.io/en/develop/assembly.html#declaring-assembly-local-variables

nuevoalex: Ill iterate on this when im back at the computer then. Shouldnt be hard to make appropriate changes

PR #383, opened 2017-10-18
Checks status when waiting for a transaction receipt.

MicahZoltu:
* Adds Byzantium details to chain specifications for Parity.
* Minor cleanup/campsiting.
* Puts back the hard-coded gas on uploads, to work around the current problem of gas limits being hit.

### DIFF
--- a/parity-dev-node/instant-seal-chain-spec.json
+++ b/parity-dev-node/instant-seal-chain-spec.json
@@ -4,11 +4,24 @@
         "instantSeal": null
     },
     "params": {
-        "accountStartNonce": "0x0",
         "gasLimitBoundDivisor": "0x400",
+        "accountStartNonce": "0x0",
         "maximumExtraDataSize": "0x20",
         "minGasLimit": "0x1",
-        "networkID" : "0x11"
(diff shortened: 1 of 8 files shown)


parity-dev-node/instant-seal-chain-spec.json
```diff
@@ -4,11 +4,24 @@
...
+        "eip161dTransition": "0x0",
+        "eip649Reward": "0x29A2241AF62C0000",
+        "eip100bTransition": "0x0",
+        "eip649Transition": "0x0",
+        "eip98Transition": "0xffffffffffffffff",
+        "eip140Transition": "0x0",
+        "eip211Transition": "0x0",
+        "eip214Transition": "0x0",
+        "eip155Transition": "0x0",
+        "eip658Transition": "0x0"
```
adrake33: What are these settings for?

MicahZoltu: These enable the various hard-fork changes at block 0.  "Byzantium" is really a rollup of a bunch of changes, as is Homestead, etc.  Parity tracks them each separately, rather than all in one block.

PR #384, opened 2017-10-18
Estimating gas for new contracts before uploading them

adrake33:


source/libraries/ContractDeployer.ts
```diff
@@ -205,8 +189,10 @@ export class ContractDeployer {
     }
 
     private async createGenesisUniverse(): Promise<ContractBlockchainData> {
-        const delegatorBuilder = this.ethjsContract(this.signatures["Delegator"], this.bytecodes["Delegator"], { from: this.testAccounts[0], gasPrice: this.gasPrice });
-        const universeBuilder = this.ethjsContract(this.signatures["Universe"], this.bytecodes["Universe"], { from: this.testAccounts[0], gasPrice: this.gasPrice });
+        const delgatorGasEstimate = await this.ethjsQuery.estimateGas(Object.assign({ from: this.testAccounts[0], data: this.bytecodes["Delegator"] }));
+        const delegatorBuilder = this.ethjsContract(this.signatures["Delegator"], this.bytecodes["Delegator"], { from: this.testAccounts[0], gas: delgatorGasEstimate, gasPrice: this.gasPrice });
+        const universeGasEstimate = await this.ethjsQuery.estimateGas(Object.assign({ from: this.testAccounts[0], data: this.bytecodes["Universe"] }));
```
MicahZoltu: This isn't necessary as we don't actually upload the Universe in this code path.  We just instantiate `universeBuilder` and then point it at the uploaded delegator below.

source/libraries/ContractDeployer.ts
```diff
@@ -205,8 +189,10 @@ export class ContractDeployer {
     }
 
     private async createGenesisUniverse(): Promise<ContractBlockchainData> {
-        const delegatorBuilder = this.ethjsContract(this.signatures["Delegator"], this.bytecodes["Delegator"], { from: this.testAccounts[0], gasPrice: this.gasPrice });
-        const universeBuilder = this.ethjsContract(this.signatures["Universe"], this.bytecodes["Universe"], { from: this.testAccounts[0], gasPrice: this.gasPrice });
+        const delgatorGasEstimate = await this.ethjsQuery.estimateGas(Object.assign({ from: this.testAccounts[0], data: this.bytecodes["Delegator"] }));
+        const delegatorBuilder = this.ethjsContract(this.signatures["Delegator"], this.bytecodes["Delegator"], { from: this.testAccounts[0], gas: delgatorGasEstimate, gasPrice: this.gasPrice });
```
MicahZoltu: `gas: delegatorGasEstimate` should be put on the `delegatorBuilder.new(...)` call.  I forget the exact syntax for passing a transaction object to `new`, but it is possible.  The same goes for the contract uploads above.  The gas should go on the `new` call, not on the default transaction object.

source/libraries/ContractDeployer.ts
adrake33: Oops, yeah, good point.  I'll remove it.

source/libraries/ContractDeployer.ts
MicahZoltu: It looks like the syntax is `delegatorBuilder.new(this.controller.address, stringTo32ByteHex('Universe'), { gas: estimatedGas })`

adrake33: Thanks, I'll make this change.

MicahZoltu: Same comment down in `getStakeToken`.  Estimated gas should be provided to the `new` call, not set as part of the default transaction object.

PR #388, opened 2017-10-19
Fixes integration tests and renames LegacyReputationToken.

MicahZoltu:
The rename is to bring the naming convention in line with the rest of our contracts.

The fix was to wait for legacy reputation token to be approved before calling migrate reputation.  This is necessary because when we estimate the gas cost of migration, it must be _after_ approval otherwise it underestimates because the transaction fails.

### DIFF
--- a/source/libraries/AbiParser.ts
+++ b/source/libraries/AbiParser.ts
@@ -21,7 +21,7 @@ export function parseAbiIntoMethods(ethjsQuery: EthjsQuery, abi: (CompilerOutput
     const result: { [methodName: string]: ContractMethod } = {};
     const items = abi.filter(item => item.type === 'function').forEach(item => {
         result[item.name] = async function(this: TransactionOptions, ...vargs: any[]) {
-            const callConvention = (this.constant || defaultTransaction.constant || item.constant) ? 'call' : 'sendTransaction';
+            const callConvention = (Object.assign({ constant: item.constant }, defaultTransaction, this)).constant ? 'call' : 'sendTransaction';
             const from = this.from || defaultTransaction.from;
             if (!from && callConvention === 'sendTransaction') throw new Error("Must have a `from`.");
             const to = this.to || defaultTransaction.to;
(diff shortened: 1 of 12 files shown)


MicahZoltu:
@nuevoalex I tagged you since this includes a contract rename.

source/libraries/AbiParser.ts
```diff
@@ -21,7 +21,7 @@ export function parseAbiIntoMethods(ethjsQuery: EthjsQuery, abi: (CompilerOutput
     const result: { [methodName: string]: ContractMethod } = {};
     const items = abi.filter(item => item.type === 'function').forEach(item => {
         result[item.name] = async function(this: TransactionOptions, ...vargs: any[]) {
-            const callConvention = (this.constant || defaultTransaction.constant || item.constant) ? 'call' : 'sendTransaction';
+            const callConvention = (Object.assign({ constant: item.constant }, defaultTransaction, this)).constant ? 'call' : 'sendTransaction';
```
MicahZoltu: This change allows us to do `contract.method.bind({constant: false})(...parameters)` when we want to call a normally constant method on-chain.  Useful for debugging sometimes.

source/libraries/ContractDeployer.ts
```diff
@@ -134,7 +135,7 @@ export class ContractDeployer {
 
         const promises: Promise<ContractBlockchainData|undefined>[] = [];
         for (let contractFileName in this.compiledContracts) {
-            if (contractFileName === "Controller.sol" || contractFileName === "libraries/Delegator.sol") {
+            if (contractFileName === "Controller.sol" || contractFileName === "libraries/Delegator.sol" || contractFileName.startsWith('legacy_reputation/')) {
```
MicahZoltu: We should never have been uploading the old legacy rep contracts.  We should also remove them from compilation, but that required more understanding then I wanted to figure out at the moment.

source/libraries/ContractDeployer.ts
```diff
@@ -187,6 +188,18 @@ export class ContractDeployer {
         await Promise.all(promises);
     }
 
+    private async createGenesisUniverse(): Promise<ContractBlockchainData> {
```
MicahZoltu: `approveCentralAuthority` and `getStakeToken` were actually moved down to below the TODO/CONSIDER comments, this wasn't moved up but Git sees this moving up as the smaller change.

PR #389, opened 2017-10-19
Adds geth integration tests to docker-compose.

MicahZoltu:


source/libraries/Connector.ts
```diff
@@ -0,0 +1,35 @@
...
+    public async waitUntilConnected(): Promise<EthjsQuery> {
+        const ethjsQuery = await this.getEthjsQuery();
+        while (true) {
+            try {
+                await ethjsQuery.net_version();
+                break;
+            } catch {
+                // swallow the error, we just want to loop until
+            }
+        }
```
adrake33: I don't understand this loop.  Why does an exception being thrown on net_version mean it's connected?  Does it break out of the loop after going into the catch block?

MicahZoltu: Oops, looks like I didn't finish writing that comment.  :)  If the node is not available then `net_version()` will throw an exception, meaning the `break` statement won't be reached.  This loop causes it to loop until `net_version()` runs _without_ throwing an exception, at which point we know the node is online and connectable.  This is necessary because the Geth node starts up slower than Parity and the tests sometimes will run because the node is fully online.

MicahZoltu: Edited the comment to be a complete sentence.  :)

adrake33: Ok, I see.  That makes more sense now.

PR #390, opened 2017-10-19
Universe created event

nuevoalex:
Just adding a log

### DIFF
--- a/source/contracts/Augur.sol
+++ b/source/contracts/Augur.sol
@@ -3,13 +3,21 @@ pragma solidity 0.4.17;
 
 import 'Controlled.sol';
 import 'libraries/token/ERC20.sol';
+import 'reporting/IUniverse.sol';
 
 
 // AUDIT/CONSIDER: Is it better that this contract provide generic functions that are limited to whitelisted callers or for it to have many specific functions which have more limited and specific validation?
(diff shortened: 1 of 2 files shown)


MicahZoltu:
Is this part of Augur node requirements, or something else?

nuevoalex:
There was a one off issue for this event specifically.

The issue says it is also needed for the EventEmitter contract for testing, but I imagine it is needed by the Augur Node.

MicahZoltu:
We have discussed having all these logs come from Augur.sol.  Thoughts?

nuevoalex:
The question to me if we did that is how Augur.sol will do trust checks on the callers. Doing so will probably increase gas costs which seems wasteful just to centralized logging.

This is assuming we care about keeping people from messing up logs.

MicahZoltu:
The argument for centralizing the logging is that it makes it possible to query for them.  When we fetch logs, we need to provide the contract that wrote the log.  This is difficult for logs like create universe because we can't know the address of the created universe until _after_ it is created which means we can't actually fetch the log that tells us the address of the created universe. The same goes for any operation that creates a contract.

Also, it is simpler for consumers to query a single log source than to query across 100s or 1000s of sources.

We should discuss this with the Augur node team.

PR #393, opened 2017-10-20
Minor fix to `waitForTransactionReceipt` for Geth

adrake33:
Allowing receipt status to be either a number or a string since Geth & Parity return it as different types.


source/libraries/HelperFunctions.ts
```diff
@@ -85,7 +85,7 @@ export async function waitForTransactionReceipt(ethjsQuery: EthjsQuery, transact
         await sleep(ETHEREUM_POLLING_INTERVAL_MILLISECONDS);
         receipt = await ethjsQuery.getTransactionReceipt(transactionHash);
     }
-    if (!receipt.status) {
+    if (!parseInt(receipt.status.toString(), 16)) {
```
MicahZoltu: `16.toString()` will return `"16"`, but `parseInt` is expecting base 16 so this is technically incorrect.  It happens to work right now because 0 and 1 are the only two values for status, both of which are the same in base 10 and base 16.

Recommend something like this instead:
```
const status = (typeof receipt.status === 'number') ? receipt.status : parseInt(receipt.status, 16)
if (!status) { ... }
```

PR #394, opened 2017-10-21
Reporting escape hatch

nuevoalex:
As discussed in Slack I got rid of a centralized escape hatch contract for reporting and instead just place an emergency method on the relevant contracts.

I limited the `onlyInGoodTimes` modifiers to only public facing functions and initializers (to prevent creation of new contracts) just to save on gas costs of putting them everywhere. If we'd like to add more or just put them on everything I can do that as well.


source/contracts/reporting/DisputeBondToken.sol
```diff
@@ -95,6 +95,23 @@ contract DisputeBondToken is DelegationTarget, ITyped, Initializable, ERC20Basic
         return _amountToMint;
     }
 
+    function withdrawInEmergency() public onlyInBadTimes returns (bool) {
+        require(msg.sender == bondHolder);
+        IReputationToken _reputationToken = getReputationToken();
+        uint256 _amountToTransfer = _reputationToken.balanceOf(this);
+        if (bondRemainingToBePaidOut > bondAmount) {
+            uint256 _amountToCollectFeesOn = _amountToTransfer.min(bondRemainingToBePaidOut - bondAmount);
+            market.getReportingWindow().collectReportingFeesInEmergency(bondHolder, _amountToCollectFeesOn);
```
MicahZoltu: The escape hatch code should be as simple as possible, and for this reason I don't think we should try to distribute fees to stakers.  I think we should just let the various tokens be turned back into REP and nothing more.  We can address the problem of "what to do with the fees" separately.

At a glance, I think this withdraw function would be ~4 lines long and have no conditionals if we don't try to payout fees.

MicahZoltu:
I'm not sure if this is what you did or not (can't quite tell from your description) but I think the `onlyInGoodTimes` modifier should be applied to all `public`/`external` methods that are not `view` or `pure`.

source/contracts/reporting/ReportingAttendanceToken.sol
```diff
@@ -43,6 +43,16 @@ contract ReportingAttendanceToken is DelegationTarget, ITyped, Initializable, Va
         return true;
     }
 
+    function withdrawInEmergency() public onlyInBadTimes returns (bool) {
+        uint256 _attotokens = balances[msg.sender];
+        if (_attotokens != 0) {
+            burn(msg.sender, _attotokens);
+            reportingWindow.getReputationToken().transfer(msg.sender, _attotokens);
+        }
+        reportingWindow.collectReportingFeesInEmergency(msg.sender, _attotokens);
```
MicahZoltu: Same comment as above, don't try to redistribute fees.

source/contracts/reporting/ReputationToken.sol
```diff
@@ -31,7 +31,7 @@ contract ReputationToken is DelegationTarget, ITyped, Initializable, VariableSup
     }
 
     // AUDIT: check for reentrancy issues here, _destination will be called as contracts during validation
-    function migrateOut(IReputationToken _destination, address _reporter, uint256 _attotokens) public afterInitialized returns (bool) {
+    function migrateOut(IReputationToken _destination, address _reporter, uint256 _attotokens) public onlyInGoodTimes afterInitialized returns (bool) {
```
MicahZoltu: I think everything except the ERC20 token part of this contract (and VariableSupply mint/burn since those are needed for the escape) should be `onlyInGoodTimes` (meaning all `public`/`external` non `view`/`pure` methods).  The same goes for the other tokens as well.

MicahZoltu:
In general I think there should be far more `onlyInGoodTimes` modifiers.  For example, I don't see any modifiers on `Universe.sol`.  We should assume that if this escape hatch is triggered, there is a critical bug in the system and we should basically stop everything other than withdrawing.  This _may_ be overkill, but I think it is the safe path to take since this is the "last resort shut it all down" path.

source/contracts/reporting/Market.sol
```diff
@@ -329,18 +329,28 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
         return true;
     }
 
+    function withdrawInEmergency() public onlyInBadTimes onlyOwner returns (bool) {
+        if (this.balance > 0) {
+            require(msg.sender.call.value(this.balance)());
```
MicahZoltu: We should put the ETH transfer after the REP transfer to avoid reentrancy issues.  I don't think its actually a problem here, but its generally a good practice.

PR #397, opened 2017-10-22
#3936 5 reporting window tests

bthaile:
Here is the bulk of reporting window unit tests. Need to add a few more tests for fees. In the spirit of getting a PR in before it gets massive.


tests/reporting/test_reporting_windows.py
```diff
@@ -0,0 +1,485 @@
...
+from datetime import timedelta
+from utils import longToHexString, stringToBytes, bytesToHexString
+from pytest import fixture, raises
+from ethereum.tools.tester import ABIContract, TransactionFailed
+from reporting_utils import proceedToDesignatedReporting, proceedToRound1Reporting, proceedToRound2Reporting, proceedToForking, finalizeForkingMarket, initializeReportingFixture
+
+
+def test_reporting_window_initialize(localFixture, chain, mockUniverse, mockReportingAttendanceToken, mockReputationToken):
+    timestamp = chain.head_state.timestamp
+    mockUniverse.setReportingPeriodDurationInSeconds(timestamp)
```
nuevoalex: whats this value here? Reads a little confusingly as its setting a duration to a timestamp

tests/reporting/test_reporting_windows.py
```diff
@@ -0,0 +1,485 @@
...
+    assert mockMarketFactory.getCreateMarketDenominationTokenValue() == cash.address
+    assert mockMarketFactory.getCreateMarketCreatorValue() == bytesToHexString(tester.a0)
+    assert mockMarketFactory.getCreateMarketDesignatedReporterAddressValue() == bytesToHexString(designatedReporterAddressValue)
+    assert mockReputationToken.getTrustedTransferSourceValue() == bytesToHexString(tester.a0)
+    assert mockReputationToken.getTrustedTransferDestinationValue() == mockMarketFactory.address
+    assert mockReputationToken.getTrustedTransferAttotokensValue() == 10
+    assert reportingWindow3.getNumMarkets() == 1
+
+def test_reporting_window_migrate_market_in_from_sib(localFixture, mockUniverse, populatedReportingWindow):
+    with raises(TransactionFailed, message="method has be called from Market"):
```
nuevoalex: s/be/been

tests/reporting/test_reporting_windows.py
bthaile: I'm a bit confused what _reportingWindowId is suppose to be a date? It's uint256 is that in ticks?

I can set _reportingWindowId as a timestamp and have  

universe.getReportingPeriodDurationInSeconds() == Reporting.reportingDurationSeconds() + Reporting.reportingDisputeDurationSeconds();

tests/reporting/test_reporting_windows.py
```diff
@@ -0,0 +1,485 @@
...
+    assert populatedReportingWindow.getNumInvalidMarkets() == 1
+
+    with raises(TransactionFailed, message="market has already been finialized"):
+        mockMarket.callReportingWindowUpdateMarketPhase(populatedReportingWindow.address)
+
+def test_reporting_window_update_fin_market_calculations(localFixture, mockMarket, populatedReportingWindow):
+    mockMarket.setFinalPayoutDistributionHash(stringToBytes("101"))
+    mockMarket.setDesignatedReportPayoutHash(stringToBytes("10"))
+    mockMarket.setIsValid(True)
+    finializMarket(localFixture, mockMarket, populatedReportingWindow, 99, 15)
```
nuevoalex: s/finaliz/finalize

tests/reporting/test_reporting_windows.py
nuevoalex: The `ReportingPeriodDuration` is the total amount of time a reporting window takes to get to complete, which currently includes 27 days or reporting and 3 days for disputing the result.

The `ReportingWindowID` is essentially an index of these windows. So ID `0` would be the first 30 day window and ID `1` would be the window after it which takes place between days 30 and 60.

Right now this is setting the actual reporting period duration in seconds (which is normally 30 * 24 * 60 * 60) to whatever the current time in seconds since epoch is at the start of this test, which seems wrong to me

nuevoalex: My inclination is that instead of using the timestamp throughout the test as a time delta like this we should just be using some small understandable number like `10`.

PR #398, opened 2017-10-22
Fixes failing Parity integration test.

MicahZoltu:
Parity yest was failing because the end time for the market was resulting in a floating point number and BN.js _hates_ floating point numbers.  Unfortunately, the failure was swallowed somewhere in the stack so we just got an obscure failure rather than a useful one.

Geth doesn't seem to correctly estimate gas when there is a nested contract creation (in this case Reputation token is created as part of `universe.initialize`), and the ethjs default call doesn't supply gas so Geth is forced to do whatever it does in that situation.  By switching over to `parseAbiIntoMethods` (our home grown thing), we instead do `eth_estimateGas` which Geth apparently _does_ get right and the initialize call succeeds.

TODO: create a bug report with Geth indicating the problem.  It is also possible Geth has some internal limit it sets on no-gas transactions that is being hit.  Regardless, it works in Parity but not in Geth before, now it works in both.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,8 @@
         "@types/get-port": {
             "version": "3.2.0",
             "resolved": "https://registry.npmjs.org/@types/get-port/-/get-port-3.2.0.tgz",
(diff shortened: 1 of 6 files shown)


MicahZoltu:
Can I get a re-review on this @adrake33.  I updated it to also fix Geth tests.

adrake33:
Looks good!  Thanks for digging into this issue, @MicahZoltu!

PR #399, opened 2017-10-23
Short-term fix for WCL fuzzy test

adrake33:
Added snapshot resets after each test and replaced usages of cash with ETH.  (Longer term, the plan is to convert this from 3 tests into 24 parameterized tests.)


tests/trading/test_wcl_fuzzy.py
```diff
@@ -89,33 +94,35 @@ def acquireShortShareSet(outcome, amount, approvalAddress, sender):
...
-    completeSetFees = long(fxpAmount * market.getNumTicks() * fix('0.0101') / 10**18 / 10**18)
+def execute_bidOrder_tests(fixture, kitchenSinkSnapshot, universe, market, fxpAmount, fxpPrice):
+    longCost = long(fxpAmount * fxpPrice)
+    shortCost = long(fxpAmount * (market.getNumTicks() - fxpPrice))
+    completeSetFees = long(fxpAmount * market.getNumTicks() * fix('0.0101') / 10**18)
     shortFee = long((completeSetFees * shortCost) / (longCost + shortCost))
     longFee = completeSetFees - shortFee
 
-    print "creator escrows cash, filler pays with cash"
+    print "creator escrows ETH, filler pays with ETH"
```
MicahZoltu: There should be a reset to snapshot call before this first execute, otherwise it will run with whatever state the last test to run was in.

Consider moving the `fixture.resetToSnapshot(kitchenSinkSnapsho)` into the `execute` method to ensure that it is run before every iteration, without having to add it before each of the calls to execute.  Note: This will require adding the `kitchenSinkSnapshot` as a parameter to `execute`.

tests/trading/test_wcl_fuzzy.py
```diff
@@ -296,58 +317,58 @@ def execute_askOrder_tests(contractsFixture, universe, market, fxpAmount, fxpPri
         expectedFillerShortShares = 0,
         expectedFillerTokens = 0)
 
-def test_binary(contractsFixture, market, randomAmount, randomNormalizedPrice):
+    fixture.resetToSnapshot(kitchenSinkSnapshot)
```
MicahZoltu: No need to reset the snapshot at the end, just at the beginning (see note above about moving the resets into the `execute` function).

tests/trading/test_wcl_fuzzy.py
```diff
@@ -89,33 +94,35 @@ def acquireShortShareSet(outcome, amount, approvalAddress, sender):
...
     acquireLongShares(orderOutcome, fillerLongShares, fillOrder.address, sender = fillerKey)
     acquireShortShareSet(orderOutcome, fillerShortShares, fillOrder.address, sender = fillerKey)
     remaining = fillOrder.publicFillOrder(orderID, orderSize, 42, sender = fillerKey, value = fillerTokens)
     assert not remaining
 
     # assert final state
-    assert cash.balanceOf(creatorAddress) == expectedMakerTokens
-    assert cash.balanceOf(fillerAddress) == expectedFillerTokens
+    assert fixture.chain.head_state.get_balance(creatorAddress) == creatorOriginalBalance - creatorEthRequiredLong - creatorEthRequiredShort - creatorTokens + expectedMakerTokens
+    assert fixture.chain.head_state.get_balance(fillerAddress) == fillerOriginalBalance - fillerEthRequiredLong - fillerEthRequiredShort - fillerTokens + expectedFillerTokens
```
MicahZoltu: This change feels incorrect.  We should probably rename `expectedMakerTokens` and `expectedFillerTokens` to `expectedMakerEth` and `expectedTakerEth` but it seems like we should still be testing against the incoming expectation directly.

If this change is necessary, can you explain a bit why that is so?

adrake33: I agree that `expectedMakerTokens` and `expectedFillerTokens` should be renamed.  However, comparing the ETH balance of the creator/filler to these values in the same way we were dong with the cash balance will fail.  The ETH balance of the creator is set very high at the beginning of the test, whereas the `expectedMakerTokens` and `expectedFillerTokens` are typically set to values like `shortCost`, `shortCost - shortFee`, 0, etc.  

I agree that it seems odd to do the assert in this way, but in order to do it like we were before, I think we'd have to alter the values we're passing in for `expectedMakerTokens` and `expectedFillerTokens`.

tests/trading/test_wcl_fuzzy.py
adrake33: Sounds good.  I'll move the reset line into `execute`.

tests/trading/test_wcl_fuzzy.py
MicahZoltu: Do all of the test accounts have a bunch of ETH in them, or only a the first one?  One option would be to transfer away the excess ETH before the tests start, so the balance starts at "exactly the right amount" and ends at "exactly the right amount".  Even as it is, I would expect `creatorOriginalBalance + expectedMakerTokens` and `fillerOriginalBalance + expectedFillerTokens`.  I would not expect all of the stuff in the middle, as that suggests there is a bug in the test somewhere.

adrake33: What if we did asserts on the balance of the creator/filler after the creator acquires shares, after the order is created, after the filler acquires shares, and after the order is filled?  That way, the final asserts would only be comparing the users' current balances with their previous balance + the expected maker/filler tokens?  Plus, it would verify their balances are correct along the way.

adrake33:
I added some more asserts on the ETH balances of the tester accounts so that final asserts are clearer.  

I also reverted the original account balances back to 10**24.  However, upon doing this, I found that the `expectedMakerTokens` and `expectedFillerTokens` were off by one when executing the "creator escrows shares, filler pays with shares" test.  (`expectedMakerTokens` was one too high and `expectedFillerTokens` one too low for ask orders.  I did not see this happening for bid orders -- probably because it happened to be rounding in the correct direction.)  

This off-by-one error seems to be happening due to a rounding error when calculating `longFee`.  `longFee` was getting initialized like: `longFee = long((completeSetFees * longCost) / (longCost + shortCost))`, which was storing a decimal number into a long type variable.  This, in turn, was causing `shortFee` to get rounded (since it was being set to `shortFee = completeSetFees - longFee`.  These roundings were then also causing the following calculations to be off:

```
expectedMakerTokens = (longCost - longFee)
expectedFillerTokens = (shortCost - shortFee)
```

PR #403, opened 2017-10-23
Fix for Orders Fuzzy Test

adrake33:
Updated Orders fuzzy test to accommodate  changes to order ID type (i.e., bytes instead of integers).  Also updated test to use kitchenSinkSnapshot.


tests/conftest.py
```diff
@@ -167,7 +167,7 @@ def __init__(self):
         config_metropolis['BLOCK_GAS_LIMIT'] = 2**60
 
         for a in range(10):
-            tester.base_alloc[getattr(tester, 'a%i' % a)] = {'balance': 10**24}
+            tester.base_alloc[getattr(tester, 'a%i' % a)] = {'balance': 10**40}
```
MicahZoltu: Why the drastic increase in seeded capital?  That is more ETH than there is in existence I believe.

adrake33: I could lower it a bit, but the sum of creatorEthRequiredLong + creatorEthRequiredShort + creatorTokens is typically on the order of 10**36.  This is because the number of clicks is now 10**18.

MicahZoltu: Can we edit the test to not try to buy so many shares?  I'm not sure how it is deciding how many shares to buy, but it seems like buying 10^20 shares is incredibly silly when the shares cost 1 ether each.

joeykrug: Dang, tester is a baller

adrake33: @MicahZoltu Like I mentioned on Slack, the number of shares is getting set to a fixed point number, so it ends up being a really high number like 8000000000000000000, and the price gets also gets set to a hight number like 547868758977303680 due to the number of ticks.  I suggested dividing them by 10**18 to make them smaller, but I believe you said the price was valid (albeit silly) and that the error was probably somewhere else.  Dividing by 10^18 seems like a sensible solution.  For example, in the above example, it would result in the number of shares being 8, and the price being $0.54, which seems like reasonable.

tests/trading/test_wcl_fuzzy.py
```diff
@@ -22,100 +22,110 @@
...
     acquireShortShareSet(orderOutcome, fillerShortShares, fillOrder.address, sender = fillerKey)
-    remaining = fillOrder.publicFillOrder(orderID, orderSize, 42, sender = fillerKey, value = fillerTokens)
+    remaining = fillOrder.publicFillOrder(orderId, orderSize, 42, sender = fillerKey, value = fillerTokens)
     assert not remaining
 
     # assert final state
-    assert cash.balanceOf(creatorAddress) == expectedMakerTokens
-    assert cash.balanceOf(fillerAddress) == expectedFillerTokens
+    assert fixture.chain.head_state.get_balance(creatorAddress) == creatorOriginalBalance - creatorEthRequiredLong - creatorEthRequiredShort - creatorTokens + expectedMakerTokens
+    assert fixture.chain.head_state.get_balance(fillerAddress) == fillerOriginalBalance - fillerEthRequiredLong - fillerEthRequiredShort - fillerTokens + expectedFillerTokens
```
MicahZoltu: Same comment as on the other PR, this seems incorrect.  The assertion should be that the participants end with the expected amount of ETH.  If `expectedMakerTokens` is a "change in ETH from start" then adding original balance would be reasonable.  The other stuff shouldn't be necessary and suggests that the test has some other problem in it that is causing the balances to not work out correctly.

tests/conftest.py
MicahZoltu: The price is fine, and should be a number between `[0, 10^18]`.  However, there is no need to have the test try to buy `10^18` shares like it is at the moment if it is doing something like `fix('1')`.  If the price of one share is `10^18`, then the tests should be buying only a few shares, on the order of 1-1000.

Rather than doing `fix(1) / 10^18`, instead we should just do `1` for the number of purchased shares (or similar).  In theory, this should just be a matter of removing the `fix(...)` wrapper around share counts, though I haven't dug deeply into these tests to confirm that.  Such a change would make it so we don't need to give the test accounts 10^22 ETH.

adrake33: Ok, that's fine.  I can remove the call to `fix` when setting the number of shares.  That should be sufficient.

adrake33:
I've reverted the changes related to WCL fuzzy test from this branch.  Please let me know if the other changes look ok, and I'll go ahead and merge this to develop.

PR #404, opened 2017-10-24
Event logging

nuevoalex:
Adding multiple people to this to make sure the event data we're sending is adequate and correct.

NOTE: we went back to using a centralized Augur contract to send all logs and are sending the universe in all logs as well. This was after Jack was filled in on the previous implementation and noted that too much currently depends on the centralized Augur contract being the source of events to change that behavior. Filtering the Universe is a simple addition in comparison.


tests/reporting/test_market.py
```diff
@@ -2,20 +2,31 @@
 from ethereum.tools import tester
 from ethereum.tools.tester import TransactionFailed
 from pytest import raises
-from utils import stringToBytes
+from utils import stringToBytes, captureFilteredLogs, bytesToHexString
 
 tester.STARTGAS = long(6.7 * 10**6)
 
-def test_market_creation(contractsFixture, universe, cash, market):
+def test_markett_creation(contractsFixture, universe, cash, market):
```
epheph: `markett` ?

source/contracts/Augur.sol
```diff
@@ -3,13 +3,98 @@ pragma solidity 0.4.17;
 
 import 'Controlled.sol';
 import 'libraries/token/ERC20.sol';
+import 'reporting/IUniverse.sol';
 
 
 // AUDIT/CONSIDER: Is it better that this contract provide generic functions that are limited to whitelisted callers or for it to have many specific functions which have more limited and specific validation?
 contract Augur is Controlled {
+    event MarketCreated(address universe, address indexed market, address indexed marketCreator, uint256 marketCreationFee, string extraInfo);
```
MicahZoltu: `universe` should be indexed on all of these so the Augur Node can filter to only universes it cares about.

MicahZoltu: Note, we are limited to 4 indexes per event, so this _may_ mean dropping one of the other indexes.  If that is necessary, we should discuss with @tinybike @epheph to figure out which one to drop.

epheph: maximum 3 indexes per event, so it's even more critical

source/contracts/Augur.sol
```diff
@@ -3,13 +3,98 @@ pragma solidity 0.4.17;
...
+        OrderFilled(_universe, _shareToken, _creator, _filler, _price, _numCreatorShares, _numCreatorTokens, _numFillerShares, _numFillerTokens, _settlementFees, _tradeGroupId);
+        return true;
+    }
+
+    function logProceedsClaimed(IUniverse _universe, address _sender, address _market, uint256 _numShares, uint256 _numPayoutTokens, uint256 _finalTokenBalance) public onlyWhitelistedCallers returns (bool) {
+        ProceedsClaimed(_universe, _sender, _market, _numShares, _numPayoutTokens, _finalTokenBalance);
+        return true;
+    }
+
+    function logUniverseForked() public returns (bool) {
```
MicahZoltu: This should include `Universe` parameter.

source/contracts/Augur.sol
```diff
@@ -3,13 +3,98 @@ pragma solidity 0.4.17;
...
+        UniverseForked(msg.sender);
+    }
+
+    function logUniverseCreated(IUniverse _childUniverse) public returns (bool) {
```
MicahZoltu: This should include parentUniverse as first parameter (that is the thing that needs to be trusted).

source/contracts/Augur.sol
MicahZoltu: Hmm, I see, the universe is expected to be the caller... OK, nevermind.

source/contracts/Augur.sol
MicahZoltu: Same as above, I now see that the parent universe is the caller, so it doesn't need to be a separate parameter.

source/contracts/Augur.sol
```diff
@@ -3,13 +3,98 @@ pragma solidity 0.4.17;
...
+        return true;
+    }
+
+    function logMarketFinalized(IUniverse _universe, address _market) public returns (bool) {
+        require(_universe.isContainerForMarket(IMarket(msg.sender)));
+        MarketFinalized(_universe, _market);
+        return true;
+    }
+
+    function logOrderCanceled(IUniverse _universe, address _shareToken, address _sender, bytes32 _orderId, uint8 _orderType, uint256 _tokenRefund, uint256 _sharesRefund) public onlyWhitelistedCallers returns (bool) {
```
MicahZoltu: Bleh, I'm not a fan of mixing trust strategies here, though I suppose there isn't really any alternative option?  I'm referring to how some of these functions use `onlyWhitelistedCallers` and others use `require(universe.isContainerFor...)`.  😢

source/contracts/Augur.sol
nuevoalex: Brought this up in Slack. I think we need @jack to weigh in.

source/contracts/Augur.sol
nuevoalex: Yea, the trading contracts are centralized and whitelisted so it's a bit different.

source/contracts/Augur.sol
nuevoalex: Brought this up in Slack too, but another option might be not sending the universe at all in these logs since they are sent to `Augur` by whitelisted contracts so we know they are legitimate and shouldn't need to filter them at all.

tinybike: Hm, currently it looks like just `market` and `marketCreator` are indexed, so you should be able to also add `universe` as an index.  Market and market creator aren't super important any more (b/c augur-node indexes markets already) but they're nice to have.

nuevoalex: The issue is specifically about the `Order` logs.

tinybike: Ah, gotcha -- I think we should index `universe`, `shareToken`, and `creator` for all the orders logs.

tinybike: That's a good point re: knowing they're legitimate.  Even so, I think it is still good to have universe as an index, so that if there are ever multiple (legitimate) universes, listeners can easily filter by universe address.

source/contracts/Augur.sol
```diff
@@ -3,13 +3,98 @@ pragma solidity 0.4.17;
...
+import 'reporting/IUniverse.sol';
 
 
 // AUDIT/CONSIDER: Is it better that this contract provide generic functions that are limited to whitelisted callers or for it to have many specific functions which have more limited and specific validation?
 contract Augur is Controlled {
+    event MarketCreated(address universe, address indexed market, address indexed marketCreator, uint256 marketCreationFee, string extraInfo);
+    event DesignatedReportSubmitted(address universe, address indexed reporter, address indexed market, address stakeToken, uint256 amountStaked, uint256[] payoutNumerators);
+    event ReportSubmitted(address universe, address indexed reporter, address indexed market, address stakeToken, uint256 amountStaked, uint256[] payoutNumerators);
+    event WinningTokensRedeemed(address universe, address indexed reporter, address indexed market, address stakeToken, uint256 amountRedeemed, uint256 reportingFeesReceived, uint256[] payoutNumerators);
+    event ReportsDisputed(address universe, address indexed disputer, address indexed market, uint8 reportingPhase, uint256 disputeBondAmount);
```
MicahZoltu: Why is `reportingPhase` a `uint8` instead of a `ReportingPhase`?

source/contracts/Augur.sol
```diff
@@ -3,13 +3,98 @@ pragma solidity 0.4.17;
...
+import 'reporting/IUniverse.sol';
 
 
 // AUDIT/CONSIDER: Is it better that this contract provide generic functions that are limited to whitelisted callers or for it to have many specific functions which have more limited and specific validation?
 contract Augur is Controlled {
+    event MarketCreated(address indexed universe, address indexed market, address indexed marketCreator, uint256 marketCreationFee, string extraInfo);
+    event DesignatedReportSubmitted(address indexed universe, address indexed reporter, address indexed market, address stakeToken, uint256 amountStaked, uint256[] payoutNumerators);
+    event ReportSubmitted(address indexed universe, address indexed reporter, address indexed market, address stakeToken, uint256 amountStaked, uint256[] payoutNumerators);
+    event WinningTokensRedeemed(address indexed universe, address indexed reporter, address indexed market, address stakeToken, uint256 amountRedeemed, uint256 reportingFeesReceived, uint256[] payoutNumerators);
+    event ReportsDisputed(address indexed universe, address indexed disputer, address indexed market, uint8 reportingPhase, uint256 disputeBondAmount);
```
MicahZoltu: Why is reportingPhase a uint8 instead of a ReportingPhase?

source/contracts/trading/ClaimProceeds.sol
```diff
@@ -51,11 +56,19 @@ contract ClaimProceeds is CashAutoConverter, ReentrancyGuard, IClaimProceeds {
             if (_reporterShare > 0) {
                 require(_denominationToken.transferFrom(_market, _market.getUniverse().getNextReportingWindow(), _reporterShare));
             }
+
```
MicahZoltu: Nitpick: No indentation on blank lines.  I'm surprised your editor let you save the file like this...

nuevoalex: Hm I must not have the right setting in vscode. Will add and correct

source/contracts/trading/ClaimProceeds.sol
```diff
@@ -51,11 +56,19 @@ contract ClaimProceeds is CashAutoConverter, ReentrancyGuard, IClaimProceeds {
...
+            
+            _numShares = _numShares.add(_numberOfShares);
+            _numPayoutTokens = _numPayoutTokens.add(_shareHolderShare);
         }
 
+        logProceedsClaimed(_market, msg.sender, _numShares, _numPayoutTokens);
         return true;
     }
 
+    function logProceedsClaimed(IMarket _market, address _sender, uint256 _numShares, uint256 _numPayoutTokens) private returns (bool) {
```
MicahZoltu: Is this a separate function due to stack depth issues?  If not, why not just inline the log call like everywhere else?

source/contracts/Augur.sol
nuevoalex: Was following the types provided by @tinybike

source/contracts/trading/ClaimProceeds.sol
nuevoalex: Yuuup. Stack depth fun.

source/contracts/trading/ClaimProceeds.sol
```diff
@@ -51,11 +56,19 @@ contract ClaimProceeds is CashAutoConverter, ReentrancyGuard, IClaimProceeds {
             if (_reporterShare > 0) {
                 require(_denominationToken.transferFrom(_market, _market.getUniverse().getNextReportingWindow(), _reporterShare));
             }
+            
+            _numShares = _numShares.add(_numberOfShares);
+            _numPayoutTokens = _numPayoutTokens.add(_shareHolderShare);
         }
 
+        logProceedsClaimed(_market, msg.sender, _numShares, _numPayoutTokens);
```
MicahZoltu: It feels like this should be called for each share token type that has non-zero value.  It also feels like the event should contain the address of the share being redeemed.  cc @tinybike @epheph

source/contracts/trading/FillOrder.sol
```diff
@@ -389,7 +390,7 @@ contract FillOrder is CashAutoConverter, ReentrancyGuard, IFillOrder {
         }
 
         // AUDIT: is there a reentry risk here?  we execute all of the above code, which includes transferring tokens around, before we mark the order as filled
-        _tradeData.contracts.orders.fillOrderLog(_tradeData.order.orderId, _tradeData.filler.participantAddress, _tradeData.getMakerSharesDepleted(), _tradeData.getMakerTokensDepleted(), _tradeData.getFillerSharesDepleted(), _tradeData.getFillerTokensDepleted(), _tradeGroupId);
+        _tradeData.contracts.orders.fillOrderLog(_tradeData.order.orderId, _tradeData.filler.participantAddress, _tradeData.getMakerSharesDepleted(), _tradeData.getMakerTokensDepleted(), _tradeData.getFillerSharesDepleted(), _tradeData.getFillerTokensDepleted(), _settlementFees, _tradeGroupId);
```
MicahZoltu: Do we not have enough data available to log directly here?  Going through two different contracts to log is quite unfortunate.  :/

nuevoalex: Yea we have to pull the order data from storage from the passed in ID. Without doing that IIR it's a stack depth issue

source/contracts/reporting/Market.sol
```diff
@@ -159,6 +160,7 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
             updateTentativeWinningPayoutDistributionHash(tentativeWinningPayoutDistributionHash);
         }
         reportingWindow.updateMarketPhase();
+        Augur(controller.lookup("Augur")).logReportsDisputed(getUniverse(), msg.sender, this, uint8(ReportingState.DESIGNATED_DISPUTE), _bondAmount);
```
MicahZoltu: I wonder if we should add a method on `Controller` named `getAugur` that just does:
```
function getAugur() external view returns (Augur) {
    return lookup("Augur");
}
```
Then we wouldn't have to have this ugly thing everywhere.  Only reason I think special casing is valuable in this case is because looking up this particular contract is _soo_ common.

nuevoalex: That sounds liek a good idea to me. I was bothered by this too.

source/contracts/reporting/Universe.sol
```diff
@@ -125,6 +127,7 @@ contract Universe is DelegationTarget, ITyped, Initializable, IUniverse {
     function getOrCreateChildUniverse(bytes32 _parentPayoutDistributionHash) public returns (IUniverse) {
         if (childUniverses[_parentPayoutDistributionHash] == address(0)) {
             childUniverses[_parentPayoutDistributionHash] = UniverseFactory(controller.lookup("UniverseFactory")).createUniverse(controller, this, _parentPayoutDistributionHash);
+            Augur(controller.lookup("Augur")).logUniverseCreated(childUniverses[_parentPayoutDistributionHash]);
```
MicahZoltu: This feels like it should go in `universe.initialize`.

source/contracts/reporting/ReportingWindow.sol
```diff
@@ -61,6 +61,7 @@ contract ReportingWindow is DelegationTarget, ITyped, Initializable, IReportingW
         markets.add(_newMarket);
         firstReporterMarkets.add(_newMarket);
         designatedReportNoShows += 1;
+        Augur(controller.lookup("Augur")).logMarketCreated(universe, _newMarket, msg.sender, msg.value, _extraInfo);
```
MicahZoltu: This feels like it should go in `market.initialize`.

source/contracts/reporting/Universe.sol
nuevoalex: For these cases we can't put these in the initialize because it hasn't been added into the universe tree yet (Both for this and the market creation log)

source/contracts/Augur.sol
MicahZoltu: On the client side, sending through `ReportingPhase` will come through as a uint8 until such time as enums are ABI encodable (something that is in development by the solc team I believe).  Once that is implemented, I think we'll be better off having the actual enum come through as it will allow consumers (particularly automated rendering systems like EtherScan) to be _way_ more intelligent about what they display (`5` is meaningless, but AWAITING FOR MIGRATION is not).

cc @tinybike since you want to weigh in on every change.

MicahZoltu: Actually, maybe it will be uint256... can you find out what ABI solc spits out for an event with an enum parameter in it?

source/contracts/trading/Orders.sol
```diff
@@ -18,11 +19,8 @@ contract Orders is DelegationTarget, IOrders {
     using Order for Order.Data;
     using SafeMathUint256 for uint256;
 
-    event CancelOrder(address indexed market, address indexed sender, uint256 price, uint256 amount, bytes32 orderId, uint8 outcome, Order.TradeTypes orderType, uint256 cashRefund, uint256 sharesRefund);
     event BuyCompleteSets(address indexed sender, address indexed market, uint256 amount, uint256 numOutcomes);
     event SellCompleteSets(address indexed sender, address indexed market, uint256 amount, uint256 numOutcomes, uint256 creatorFee, uint256 reportingFee);
```
MicahZoltu: Why weren't these moved over?  Does the UI not need them?  If not, I wonder if we can just delete them.  cc @tinybike

MicahZoltu: Are there any other events anywhere else in the contracts that aren't part of Augur.sol and needed by the UI?

source/contracts/trading/Orders.sol
```diff
@@ -213,9 +217,9 @@ contract Orders is DelegationTarget, IOrders {
         return true;
     }
 
-    function fillOrderLog(bytes32 _orderId, address _filler, uint256 _creatorSharesFilled, uint256 _creatorTokensFilled, uint256 _fillerSharesFilled, uint256 _fillerTokensFilled, uint256 _tradeGroupId) public returns (bool) {
+    function fillOrderLog(bytes32 _orderId, address _filler, uint256 _creatorSharesFilled, uint256 _creatorTokensFilled, uint256 _fillerSharesFilled, uint256 _fillerTokensFilled, uint256 _settlementFees, uint256 _tradeGroupId) public returns (bool) {
         Order.Data storage _order = orders[_orderId];
-        FillOrder(_order.market, _order.outcome, _order.tradeType, _orderId, _order.price, _order.creator, _filler, _creatorSharesFilled, _creatorTokensFilled, _fillerSharesFilled, _fillerTokensFilled, _tradeGroupId);
+        Augur(controller.lookup("Augur")).logOrderFilled(_order.market.getUniverse(), _order.market.getShareToken(_order.outcome), _order.creator, _filler, _order.price, _creatorSharesFilled, _creatorTokensFilled, _fillerSharesFilled, _fillerTokensFilled, _settlementFees, _tradeGroupId);
```
MicahZoltu: I would really like to get these two log methods out of Orders.  @tinybike can we remove most of the order details off of these?  The Augur Node should already have the order details in its DB from when the order was created, so supplying them here is redundant and a waste of gas.  It also would allow us to get these logs out of Orders contract and directly into Augur.sol to align with the pattern followed everywhere else.

source/contracts/Augur.sol
```diff
@@ -3,13 +3,98 @@ pragma solidity 0.4.17;
...
+    event MarketFinalized(address indexed universe, address indexed market);
+    event UniverseForked(address indexed universe);
+    event OrderCanceled(address indexed universe, address indexed shareToken, address indexed sender, bytes32 orderId, uint8 orderType, uint256 tokenRefund, uint256 sharesRefund);
+    event OrderCreated(address indexed universe, address indexed shareToken, address indexed creator, bytes32 orderId, uint256 price, uint256 amount, uint256 numTokensEscrowed, uint256 numSharesEscrowed, uint256 tradeGroupId);
+    event OrderFilled(address indexed universe, address indexed shareToken, address indexed creator, address filler, uint256 price, uint256 numCreatorShares, uint256 numCreatorTokens, uint256 numFillerShares, uint256 numFillerTokens, uint256 settlementFees, uint256 tradeGroupId);
+    event ProceedsClaimed(address indexed universe, address indexed sender, address indexed market, uint256 numShares, uint256 numPayoutTokens, uint256 finalTokenBalance);
+    event UniverseCreated(address indexed parentUniverse, address indexed childUniverse);
```
MicahZoltu: I don't see any transfer events included in here.  @tinybike @epheph, this means that any accounting you do in the DB won't line up.  For example, if account 0xabcd buys stake tokens then transfers them to 0x1234 who redeems them, you will end up with a redeem that has no correlated buy.  The same can happen for shares and any of the other tokens, which means when the UI attempts to show trading history, P/L calculations, transaction history, etc. the balance sheets won't add up.

source/contracts/Augur.sol
nuevoalex: uint8. I feel like this is a more than reasonable change given that

source/contracts/Augur.sol
nuevoalex: The Transfer event is in our `StandardToken` and `BasicToken` implementations

MicahZoltu: Yes, but for the same reason we centralized the rest of the events in Augur.sol, we also need to centralize those transfer events.  This does mean that the Transfer event will be fired twice, once for ERC20 purposes, and once for Augur Node purposes.

source/contracts/trading/ClaimProceeds.sol
tinybike: Agreed, IMO `logProceedsClaimed` should be called once per iteration (of the loop above it).

source/contracts/Augur.sol
tinybike: Ah -- the ABI JSON files unfortunately aren't able to include things like enums, so they just display it as the smallest integer type possible.  In this case the ABI JSON display type is uint8.  I think it's fine to have an enum in the Augur.sol file though -- IIRC the example Augur.sol file I put on the proposed-events branch actually does use enums :grin:

source/contracts/Augur.sol
tinybike: I previously mentioned to @nuevoalex that it was ok to leave the proposed `TokensTransfered` event out, but thinking about it more, I think @MicahZoltu is right about this -- the extra gas cost of logging twice (once for ERC20, once for Augur) is IMO outweighed by the convenience of having all the relevant logs accessible from a single contract.  Could we add `TokensTransfered` to Augur.sol?

source/contracts/trading/Orders.sol
tinybike: Yes I think these can just be deleted.  IIRC all the relevant information (read: fees) is contained in the order filled logs.

source/contracts/Augur.sol
nuevoalex: Yea. We don't actually have to log twice either really. To implement this I was imagining giving the base token a method to emit logs which will get called during transfers, where it emits what it currently does but subclasses can override however they choose (In our case sending a request to Augur to log). Do we _want_ 2 logs?

MicahZoltu: We need 2 logs if we want our tokens to be ERC20 tokens.  For ReputationToken at least, we do want this and presumably for others as well.

source/contracts/trading/Orders.sol
tinybike: Good point!  We can get rid of `creator` and `price` here, since those are both attached to the order.  However, I think we need to log the `orderId` -- this isn't presently included in `OrderFilled` (my mistake!) but we need it in order to look the order up in the augur-node DB.

I think we need all 4 of the shares/tokens filled fields and the settlement fees field included, since those will be specific to the specific fill (not attached to the order).

MicahZoltu: I believe by dropping creator/price we can pull this log out into the proper location as all other information is available in that location.  The same should be true for the other log.  It definitely seems like all order related logs should include the order ID.

tinybike: What do you mean re: pull the log into the proper location?  It looks like the `OrderFilled` log is already on the Augur.sol contract.

MicahZoltu: The log originates from `FillOrder.sol`, but it routes through `Orders.sol` only because it needs to pickup price/creator along the way.  Without needing those two pieces of information, `FillOrder.sol` can directly call `logOrderFilled` without having to route through an intermediate contract.

tinybike: Ah I see what you mean, yes I agree that would be good.

PR #405, opened 2017-10-24
Apply new rules governing when migration results in a bonus minted in the new universe

nuevoalex:
dispute bond migration and stake token redemption gives a bonus until the fork window is over regardless of finalization

### DIFF
--- a/source/contracts/reporting/ReputationToken.sol
+++ b/source/contracts/reporting/ReputationToken.sol
@@ -45,7 +45,7 @@ contract ReputationToken is DelegationTarget, ITyped, Initializable, VariableSup
     }
 
     // AUDIT: check for reentrancy issues here, _destination will be called as contracts during validation
-    function internalMigrateOut(IReputationToken _destination, address _sender, address _reporter, uint256 _attotokens, bool _bonusIfInForkWindow) internal returns (bool) {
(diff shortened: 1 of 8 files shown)


source/contracts/reporting/ReputationToken.sol
```diff
@@ -30,34 +30,58 @@ contract ReputationToken is DelegationTarget, ITyped, Initializable, VariableSup
...
+        require(universe.isContainerForDisputeBondToken(IDisputeBond(msg.sender)));
+        return internalMigrateOut(_destination, msg.sender, _reporter, _attotokens, true);
+    }
+
     function migrateOut(IReputationToken _destination, address _reporter, uint256 _attotokens) public afterInitialized returns (bool) {
+        return internalMigrateOut(_destination, msg.sender, _reporter, _attotokens, false);
+    }
+
+    // AUDIT: check for reentrancy issues here, _destination will be called as contracts during validation
+    function internalMigrateOut(IReputationToken _destination, address _sender, address _reporter, uint256 _attotokens, bool _bonusIfInForkWindow) internal returns (bool) {
```
MicahZoltu: This should probably be `private` rather than `internal`.  In general, I'm of the school of thought that things should only be `internal` if the class/contract is intentionally engineered to be derived from.  ReputationToken is not intended to be derived from.

nuevoalex: Will change

PR #406, opened 2017-10-24
#3936 rework review reporting window

bthaile:
finally finished unit test for reporting window

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -120,7 +120,7 @@ class ContractsFixture:
             },
             'settings': {
                 # TODO: Remove 'remappings' line below and update 'sources' line above
-                'remappings': [ '=%s/' % resolveRelativePath("../source/contracts") ],
+                'remappings': [ '=%s/' % resolveRelativePath("../source/contracts"), 'TEST=%s/' % resolveRelativePath("solidity_test_helpers") ],
                 'optimizer': {
                     'enabled': True,
                     'runs': 500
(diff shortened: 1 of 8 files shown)


tests/conftest.py
```diff
@@ -151,6 +151,10 @@ def getAllDependencies(filePath, knownDependencies):
         matches = findall("import ['\"](.*?)['\"]", fileContents)
         for match in matches:
             dependencyPath = path.join(BASE_PATH, '..', 'source/contracts', match)
+            if not path.isfile(dependencyPath):
```
MicahZoltu: Instead of checking if the file exists, we could instead just check to see if the filename is prefixed with `TEST`.  This will avoid having to go to disk to probe for the file when we can quickly know the path won't be found.
```
dependencyPath = path.join(BASE_PATH, '..', 'source/contracts', match) if path.startsWith('TEST') else path.join(BASE_PATH, 'solidity_test_helpers', match).replace("TEST/", "")
```
(or however Python ternarys works)

tests/reporting/test_reporting_windows.py
```diff
@@ -41,9 +41,58 @@ def test_reporting_window_initialize(localFixture, chain, mockUniverse, mockPart
...
+    assert reportingWindow.isActive() == False
+    assert reportingWindow.isReportingActive() == False
+    assert reportingWindow.isDisputeActive() == False
     assert reportingWindow.isOver()
 
+    chain.head_state.timestamp = reportingWindow.getEndTime() + 1
+    assert reportingWindow.isActive() == False
+    assert reportingWindow.isReportingActive() == False
+    assert reportingWindow.isDisputeActive() == False
+    assert reportingWindow.isOver()
```
MicahZoltu: For consistency, recommend adding `== True` to this.  Alternatively, make all of the others be `assert not ...` and `assert ...`.

tests/conftest.py
bthaile: updated to use "TEST" in dependency python's "contains"

tests/reporting/test_reporting_windows.py
bthaile: update to add "== True" for consistency.

MicahZoltu:
As we move over to the mocks you have created, it would be great if we could get away from using the `kitchenSink` fixture.  Feel free to address this in a separate PR, or just as you are working on future tests.

PR #409, opened 2017-10-27
General cleanup

nuevoalex:
Includes:

- Final renamings
- Replacing vars with types
- Turning magic numbers to constants
- Removing redundant or un-needed requires
- Some commenting cleanup

Most of this is trivial, although reviewing the removed requires to make sure they weren't actually needed and looking at the magic number refactor for the fork REP migration payout goal would be good.


source/contracts/trading/Order.sol
```diff
@@ -22,7 +22,7 @@ library Order {
 
     uint256 constant MIN_ORDER_VALUE = 10**14;
 
-    enum TradeTypes {
+    enum OrderTypes {
```
MicahZoltu: Consider `Types` here since it will always be called as `Order.<name>` and having `Order` twice is redundant.

We use `orderType` as the variable name because `type` is a reserved word, not because it is a good name.  :)

nuevoalex: Ah yea, will rename the enum

source/contracts/libraries/Delegator.sol
```diff
@@ -11,10 +11,7 @@ contract Delegator is DelegationTarget {
         controllerLookupName = _controllerLookupName;
     }
 
-    // We currently only support a single non-array return value
-    // If we truly need to return more advanced sets of data we could follow
-    // the pattern demonstrated here: https://gist.github.com/Arachnid/4ca9da48d51e23e5cfe0f0e14dd6318f
-    // But in a way that is automated.
+    // We currently only support a single non-array return value. If we truly need to return more advanced sets of data we could follow the pattern demonstrated here: https://gist.github.com/Arachnid/4ca9da48d51e23e5cfe0f0e14dd6318f but in a way that is automated.
```
MicahZoltu: We should update this to reference the Byzantium fork/clubhouse issue for supporting array returns.

source/contracts/reporting/Universe.sol
```diff
@@ -56,6 +57,8 @@ contract Universe is DelegationTarget, ITyped, Initializable, IUniverse {
         require(isContainerForMarket(IMarket(msg.sender)));
         forkingMarket = IMarket(msg.sender);
         forkEndTime = block.timestamp + Reporting.forkDurationSeconds();
+        // We pre calculate the amount of REP needed to determine a winner in the fork REP migration contest
+        forkReputationGoal = reputationToken.totalSupply() / Reporting.forkRepMigrationVictoryDivisor();
```
MicahZoltu: This opens up an attack that Id on't have a good solution for.

1. Attacker collects 99% of all REP.
2. Attacker triggers a fork.
3. Attacker does not migrate any REP to the fork (sacrifices bonds).
4. On new universe, attacker triggers another fork.
5. In the same block as the second fork is triggered, the attacker migrates 5% of their REP from original universe to intermediate universe and then from intermediate universe to final universe.
6. Attacker migrates remaining REP to truth universe.

This causes the fork to finalize the way the attacker wants (via early finalization at 51%, but only costing them dispute bonds (for first fork) + 5% of REP (for second fork).

Even if we didn't have the 51% auto-win rule for forks we would still have the same problem as the attacker could just hold their 94% of REP (minus bonds) until after the fork finalized.

One way we could tentatively solve this is by summing the total supply of reputation in the forking universe plus all of its ancestors.  The problem here is that technically that is an unbounded operation.  Perhaps its OK as the rate of growth will be really slow and we can address it later if that becomes a problem?

nuevoalex: Isn't the system already compromised in general if a malicious party acquires 99% of all REP?

While I'm sure this is a legit vulnerability we were previously hardcoding `11 * 10 ** 6 * 10 ** 18` for this number. After even one fork that number is arbitrary and results in a completely broken forking system.

MicahZoltu: No, acquiring all REP is not considered unhealthy/unsafe.  It is only a problem if attacking the system requires you to _burn_ that REP.  The security model is built around the cost of an attack, not the up-front capital required to launch the attack.  The attack I described above is such that you need to acquire 99% of all REP, but only burn 5% of that REP.  You can then sell the remaining 94% back to the market to recoup most of your initial costs.

source/contracts/reporting/ReportingWindow.sol
```diff
@@ -394,8 +400,6 @@ contract ReportingWindow is DelegationTarget, ITyped, Initializable, IReportingW
 
     function privateAddMarket(IMarket _market) private onlyInGoodTimes afterInitialized returns (bool) {
         require(!markets.contains(_market));
-        require(!firstReporterMarkets.contains(_market));
-        require(!lastReporterMarkets.contains(_market));
```
MicahZoltu: Why were these removed?  They seem to be valid checks.

nuevoalex: I believe there should be no condition, barring potentially critical bugs in the system, where a market could be missing from the `markets` list but in one of those two.

source/contracts/reporting/Universe.sol
```diff
@@ -56,6 +57,13 @@ contract Universe is DelegationTarget, ITyped, Initializable, IUniverse {
         require(isContainerForMarket(IMarket(msg.sender)));
         forkingMarket = IMarket(msg.sender);
         forkEndTime = block.timestamp + Reporting.forkDurationSeconds();
+        // We pre calculate the amount of REP needed to determine a winner early in a fork. We assume maximum possible fork inflation in every fork so this is hard to achieve with every subsequent fork and may become impossible in some universes.
+        if (parentUniverse != IUniverse(0)) {
+            uint256 _previousForkReputationGoal = parentUniverse.getForkReputationGoal();
+            forkReputationGoal = _previousForkReputationGoal + (_previousForkReputationGoal / Reporting.forkMigrationPercentageBonusDivisor());
+        } else {
+            forkReputationGoal = reputationToken.totalSupply() / Reporting.forkRepMigrationVictoryDivisor();
```
MicahZoltu: I think this should be hard-coded to 11M, not `reputationToken.totalSupply()`.  When Augur "launches", the universe will start out with a total supply of 0 and increase as people migrate REP in from the current REP token.  It is possible that the first fork occurs before everyone migrates their REP, in which case the first fork would suffer from the same problem we are trying to avoid. with this calculation.

nuevoalex: Got it. Will add a comment to that effect too.

source/contracts/reporting/Reporting.sol
```diff
@@ -55,4 +60,9 @@ library Reporting {
     function targetDesignatedReportNoShowsDivisor() internal pure returns (uint256) { return TARGET_DESIGNATED_REPORT_NO_SHOWS_DIVISOR; }
     function targetRepMarketCapMultiplier() internal pure returns (uint256) { return TARGET_REP_MARKET_CAP_MULTIPLIER; }
     function forkMigrationPercentageBonusDivisor() internal pure returns (uint256) { return FORK_MIGRATION_PERCENTAGE_BONUS_DIVISOR; }
+    function bondPayoutMultiplier() internal pure returns (uint256) { return BOND_PAYOUT_MULTIPLIER; }
+    function forkRepMigrationVictoryDivisor() internal pure returns (uint256) { return FORK_REP_MIGRATION_VICTORY_DIVISOR; }
+    function maximumReportingFeeDivisor() internal pure returns (uint256) { return MAXIMUM_REPORTING_FEE_DIVISOR; }
+    function defaultReportingFeeDivisor() internal pure returns (uint256) { return DEFAULT_REPORTING_FEE_DIVISOR; }
+    function initialREPSupply() internal pure returns (uint256) { return INITIAL_REP_SUPPLY; }
```
MicahZoltu: For naming consistency, these should _all_ be named `get*` as that is the pattern we use for dumb getters everywhere else.

Alternatively, we should remove the `get` prefix from everything else in the repo.  ;)

nuevoalex: I'll opt for putting get on these ;p

source/contracts/trading/Trade.sol
```diff
@@ -14,7 +14,7 @@ import 'libraries/CashAutoConverter.sol';
 
 
 contract Trade is CashAutoConverter, ReentrancyGuard {
-    uint256 private constant MINIMUM_GAS_NEEDED = 300000;
+    uint256 private constant MINIMUM_GAS_NEEDED = 500000;
```
MicahZoltu: I'm assuming this is up-to-date with latest?  Is it possible to author a test that will fail if gas exceeds this amount?

nuevoalex: Yea this is closer to the latest tests but still an estimate. Will add a test for this

PR #411, opened 2017-10-27
Changes polling interval to follow an exponential backoff.

MicahZoltu:
This reduces the amount of configuration necessary to run the node while supporting both fast chains and slow chains.  It also minimizes the negative impact of spamming a node with requests by having the interval slow down over time.

Note: When working against a remote node, the round-trip time to the node is included as part of the delay between transactions, so this actually starts at `round_trip+10` then increases to `round_trip+20` on the first iteration, `round_trip+30` on the next iteration, etc.


MicahZoltu:
This was previously approved in #367, re-erview is mostly a formality.  I have tested this against Parity and Geth with 1 second block times, so I believe that it is compatible with both and will not overload either.

PR #412, opened 2017-10-27
Gets fuzzy tests passing.

MicahZoltu:
This is a rebase of #399.  Even though I'm still concerned about the assertions, I think we should just go ahead and get these merged since having them run at all (and pass) with a couple questionable assertions is better than not passing at all, and I don't believe the questionable assertions are a problem in the code, just in the way the test is setup that needs further digging/investigation.

### DIFF
--- a/tests/trading/test_wcl_fuzzy.py
+++ b/tests/trading/test_wcl_fuzzy.py
@@ -22,7 +22,7 @@ GAS_PRICE = 7
 
 # TODO: turn these into 24 parameterized tests rather than 3 tests that each execute 8 sub-tests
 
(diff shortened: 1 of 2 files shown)


tests/trading/test_wcl_fuzzy.py
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 
+from decimal import *
```
nuevoalex: Usually we're explicit. I don't mind this that much but should probably be consistent

tests/trading/test_wcl_fuzzy.py
```diff
@@ -152,14 +184,15 @@ def execute_bidOrder_tests(contractsFixture, universe, market, fxpAmount, fxpPri
         fillerTokens = 0,
         expectedMakerLongShares = 0,
         expectedMakerShortShares = 0,
-        expectedMakerTokens = shortCost - shortFee,
+        expectedMakerTokens = (shortCost - shortFee).quantize(Decimal('1.'), rounding=ROUND_UP),
```
nuevoalex: This is cool, haven't seen that utility before

tests/trading/test_wcl_fuzzy.py
MicahZoltu: Fixed.

PR #415, opened 2017-10-27
Eth accidental extraction

nuevoalex:
I ended up making separate methods as the intention of the `suicideFunds` method is to also destroy the contract.

I'm using a stored boolean for ETH extraction and a blacklist mapping for token extraction. This seemed to work pretty well but I'm open to alternatives.

### DIFF
--- a/source/contracts/Augur.sol
+++ b/source/contracts/Augur.sol
@@ -134,8 +134,7 @@ contract Augur is Controlled, Extractable {
         return true;
     }
 
-    function getProtectedTokens() internal returns (address[]) {
-        address[] memory _protectedTokens = new address[](0);
-        return _protectedTokens;
+    function getProtectedTokens() internal returns (address[] memory) {
+        return new address[](0);
     }
 }
--- a/source/contracts/Augur.sol
+++ b/source/contracts/Augur.sol
@@ -9,10 +9,11 @@ import 'reporting/IStakeToken.sol';
 import 'reporting/IReputationToken.sol';
 import 'trading/IShareToken.sol';
 import 'trading/Order.sol';
+import 'libraries/Extractable.sol';
 
 
 // Centralized approval authority and event emissions
-contract Augur is Controlled {
+contract Augur is Controlled, Extractable {
     event MarketCreated(address indexed universe, address indexed market, address indexed marketCreator, uint256 marketCreationFee, string extraInfo);
     event DesignatedReportSubmitted(address indexed universe, address indexed reporter, address indexed market, address stakeToken, uint256 amountStaked, uint256[] payoutNumerators);
     event ReportSubmitted(address indexed universe, address indexed reporter, address indexed market, address stakeToken, uint256 amountStaked, uint256[] payoutNumerators);
@@ -131,5 +132,10 @@ contract Augur is Controlled {
         require(_universe.isContainerForShareToken(IShareToken(msg.sender)));
         TokensTransferred(_universe, msg.sender, _from, _to, _value);
         return true;
+    }
+
+    function getProtectedTokens() internal returns (address[]) {
+        address[] memory _protectedTokens = new address[](0);
+        return _protectedTokens;
     }
 }
--- a/source/contracts/Controlled.sol
+++ b/source/contracts/Controlled.sol
@@ -8,6 +8,9 @@ import 'libraries/token/ERC20Basic.sol';
 
 contract Controlled is IControlled {
     IController internal controller;
+    // Inidcates whether ETH may be extracted from this contract by the controller. It should only be allowed when ETH cannot legitimately be held by the contract
+    bool internal mayExtractETH = true;
+    mapping(address => bool) tokenExtractionDisallowed;
 
     modifier onlyWhitelistedCallers {
         require(controller.assertIsWhitelisted(msg.sender));
@@ -45,13 +48,27 @@ contract Controlled is IControlled {
     function suicideFunds(address _target, ERC20Basic[] _tokens) public onlyControllerCaller returns(bool) {
         // Transfer tokens to target
         for (uint256 i = 0; i < _tokens.length; i++) {
-            ERC20Basic token = _tokens[i];
-            uint256 balance = token.balanceOf(this);
-            token.transfer(_target, balance);
+            ERC20Basic _token = _tokens[i];
+            uint256 _balance = _token.balanceOf(this);
+            _token.transfer(_target, _balance);
         }
 
         // Transfer Eth to target and terminate contract
         selfdestruct(_target);
         return true;
+    }
+
+    // Send accidentally received ETH to the controller
+    function extractETH() public onlyControllerCaller returns(bool) {
+        require(mayExtractETH);
+        require(controller.call.value(this.balance)());
+        return true;
+    }
+
+    // Send accidentally received tokens to the controller
+    function extractTokens(ERC20Basic _token) public onlyControllerCaller returns (bool) {
+        require(!tokenExtractionDisallowed[_token]);
+        uint256 _balance = _token.balanceOf(this);
+        _token.transfer(controller, _balance);
     }
 }
--- a/source/contracts/Controlled.sol
+++ b/source/contracts/Controlled.sol
@@ -47,7 +47,7 @@ contract Controlled is IControlled {
         for (uint256 i = 0; i < _tokens.length; i++) {
             ERC20Basic _token = _tokens[i];
             uint256 _balance = _token.balanceOf(this);
-            _token.transfer(_target, _balance);
+            require(_token.transfer(_target, _balance));
         }
 
         // Transfer Eth to target and terminate contract
--- a/source/contracts/Controlled.sol
+++ b/source/contracts/Controlled.sol
@@ -8,9 +8,6 @@ import 'libraries/token/ERC20Basic.sol';
 
 contract Controlled is IControlled {
     IController internal controller;
-    // Inidcates whether ETH may be extracted from this contract by the controller. It should only be allowed when ETH cannot legitimately be held by the contract
-    bool internal mayExtractETH = true;
-    mapping(address => bool) tokenExtractionDisallowed;
 
     modifier onlyWhitelistedCallers {
         require(controller.assertIsWhitelisted(msg.sender));
@@ -58,18 +55,5 @@ contract Controlled is IControlled {
         return true;
     }
 
-    // Send accidentally received ETH to the controller
-    function extractETH() public onlyControllerCaller returns(bool) {
-        require(mayExtractETH);
-        require(controller.call.value(this.balance)());
-        return true;
-    }
 
-    // Send accidentally received tokens to the controller
-    function extractTokens(ERC20Basic _token) public onlyControllerCaller returns (bool) {
-        require(!tokenExtractionDisallowed[_token]);
-        uint256 _balance = _token.balanceOf(this);
-        _token.transfer(controller, _balance);
-        return true;
-    }
 }
--- a/source/contracts/Controlled.sol
+++ b/source/contracts/Controlled.sol
@@ -54,6 +54,4 @@ contract Controlled is IControlled {
         selfdestruct(_target);
         return true;
     }
-
-
 }
--- a/source/contracts/Controlled.sol
+++ b/source/contracts/Controlled.sol
@@ -70,5 +70,6 @@ contract Controlled is IControlled {
         require(!tokenExtractionDisallowed[_token]);
         uint256 _balance = _token.balanceOf(this);
         _token.transfer(controller, _balance);
+        return true;
     }
 }
--- a/source/contracts/libraries/Extractable.sol
+++ b/source/contracts/libraries/Extractable.sol
@@ -1,0 +1,34 @@
+pragma solidity 0.4.17;
+
+import 'libraries/token/ERC20Basic.sol';
+import 'Controlled.sol';
+
+
+contract Extractable is Controlled {
+    // Send accidentally received ETH to the destination
+    function extractETH(address _destination) public onlyControllerCaller returns(bool) {
+        require(mayExtract(ERC20Basic(0)));
+        require(_destination.call.value(this.balance)());
+        return true;
+    }
+
+    // Send accidentally received tokens to the destination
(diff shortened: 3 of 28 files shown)


source/contracts/Controlled.sol
```diff
@@ -8,6 +8,9 @@ import 'libraries/token/ERC20Basic.sol';
 
 contract Controlled is IControlled {
     IController internal controller;
+    // Inidcates whether ETH may be extracted from this contract by the controller. It should only be allowed when ETH cannot legitimately be held by the contract
+    bool internal mayExtractETH = true;
```
MicahZoltu: Since `mayExtractETH` is the "less secure" option, we should probably default to `false` and then have contracts explicitly assert that ETH extraction is reasonable.  This way if we "forget" one, we don't introduce an attack vector for the controller owner.

source/contracts/Controlled.sol
```diff
@@ -8,6 +8,9 @@ import 'libraries/token/ERC20Basic.sol';
...
+    mapping(address => bool) tokenExtractionDisallowed;
```
MicahZoltu: Hmm, same argument as above, but much harder to implement with the mapping.  :/  Can we make it so that if the derived class doesn't take any action all token withdraws are disallowed, and if they take action then this blacklist is used?

MicahZoltu: EtherDelta combines the ETH and Tokens into a single mapping where ETH is address `0x0000...`.  Consider doing the same here rather than having two variables that are used in very similar ways.  This would make ETH "less special".

nuevoalex: Doing this would mean that ETH needs to be blacklisted rather then whitelisted if I'm understanding correctly. Otherwise I do like it.

nuevoalex: W/ regard to the first question that's easy enough to do, although I think it will just result in boilerplate where every subclass of controlled turns on Token withdraws through extraction won't it?

MicahZoltu: Yes (to both).

Another option is to be a bit less generalized and assert that all tokens can be extracted from all contracts by default _except_ `ReputationToken`, `Cash` and `ETH`.  Then this would become a whitelist for tokens in that set that _can_ be withdrawn.

Note: This solution will start to crumble a bit once we add support for markets denominated in other tokens.  I don't believe any of our contracts are designed to hold anything beyond those 3 right now though, so maybe it is "good enough"?

🤔 more after typing the above, Reputation is a bit tricky since there isn't just one reputation token, but there _is_ one reputation token per universe.  I wonder if we could make this somehow "do the right thing" and for contracts that shouldn't allow REP withdraw limit that constraint to only REP for the current universe?  Maybe an abstract method on this class that derivatives need to implement for `getReputationToken()`?

Or perhaps an abstract method for `getProtectedTokens() returns (address[])`?  Then every contract would be forced to make the cognitive choice as to what tokens are protected.  Many would return `[]`, but that is OK because at least we had to be explicit about it and thing about it for each contract (compiler won't let us "forget" one).

nuevoalex: I like the `getProtectedTokens() returns (address[])` option. This is actually how I first wrote this. I also like that the subclass protection is isolated in a special function then too instead of jammed into initialize.

nuevoalex: Just noting since I'm pivoting now, the abstract method route ended up being a bust since `Delegator` is a subclass of `Controlled` and there didn't seem to be a clean way to make this work because of that. In particular the nature of `Delegator` is such that since it must provide a `address _target = controller.lookup(controllerLookupName);` method that method will always be called for any delegated contracts. Adding delegation to that method to get the real contract's protected tokens value started to err into no longer being clean.

nuevoalex: Actually, thinking more on this I do like this pattern still though. I think I'll continue this route but unfortunately it won't have compiler enforcement

nuevoalex: Ugh, again nm the method at the base would still be inherited. We'd have to make a new class for everything to inherit from. Backing out to previous types of implementations using storage with the considerations we've talked about here

source/contracts/Augur.sol
```diff
@@ -132,4 +133,9 @@ contract Augur is Controlled {
         TokensTransferred(_universe, msg.sender, _from, _to, _value);
         return true;
     }
+
+    function getProtectedTokens() internal returns (address[]) {
+        address[] memory _protectedTokens = new address[](0);
+        return _protectedTokens;
```
MicahZoltu: Why two lines instead of one?  `return new address[](0);`?

source/contracts/Controlled.sol
```diff
@@ -45,13 +45,15 @@ contract Controlled is IControlled {
...
+            uint256 _balance = _token.balanceOf(this);
+            _token.transfer(_target, _balance);
         }
 
         // Transfer Eth to target and terminate contract
         selfdestruct(_target);
         return true;
     }
+
+
```
MicahZoltu: Style: No newlines between closing two curlies.

source/contracts/libraries/Extractable.sol
```diff
@@ -0,0 +1,34 @@
+pragma solidity 0.4.17;
+
+import 'libraries/token/ERC20Basic.sol';
+import 'Controlled.sol';
+
+
+contract Extractable is Controlled {
+    // Send accidentally received ETH to the destination
+    function extractETH(address _destination) public onlyControllerCaller returns(bool) {
```
MicahZoltu: Style: `Eth` (lower case last two letters)

source/contracts/Augur.sol
```diff
@@ -132,4 +133,9 @@ contract Augur is Controlled {
         TokensTransferred(_universe, msg.sender, _from, _to, _value);
         return true;
     }
+
+    function getProtectedTokens() internal returns (address[]) {
```
MicahZoltu: It would be good to specify storage in the return type: `returns (address[] memory)`

source/contracts/libraries/Extractable.sol
```diff
@@ -0,0 +1,34 @@
...
+        address[] memory _protectedTokens = getProtectedTokens();
+        for (uint8 i = 0; i < _protectedTokens.length; i++) {
+            if (_protectedTokens[i] == address(_token)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    function getProtectedTokens() internal returns (address[]);
```
MicahZoltu: Same comment as above, `returns(address[] memory)`.  Same for all overrides.  I'm a big fan of being clear/explicit about the storage location since the defaults are so complicated.

source/contracts/libraries/Extractable.sol
```diff
@@ -0,0 +1,34 @@
...
+        require(mayExtract(ERC20Basic(0)));
+        require(_destination.call.value(this.balance)());
+        return true;
+    }
+
+    // Send accidentally received tokens to the destination
+    function extractTokens(address _destination, ERC20Basic _token) public onlyControllerCaller returns (bool) {
+        require(mayExtract(_token));
+        uint256 _balance = _token.balanceOf(this);
+        _token.transfer(_destination, _balance);
```
MicahZoltu: This should be wrapped in a `require()`.  ERC20 doesn't require throw on failure, it allows return false on failure and having the function clearly fail on-chain rather than silently fail would be good.

source/contracts/libraries/Extractable.sol
MicahZoltu: Consider just calling it `Ether` in function/variable names.  I wonder if it would be good to just use `Ether` everywhere in our codebase rather than `ETH`/`Eth`/`eth`?  We aren't really saving anything by abbreviating in this case.

source/contracts/libraries/Extractable.sol
```diff
@@ -0,0 +1,34 @@
...
+        return true;
+    }
+
+    function mayExtract(ERC20Basic _token) private returns (bool) {
+        address[] memory _protectedTokens = getProtectedTokens();
+        for (uint8 i = 0; i < _protectedTokens.length; i++) {
+            if (_protectedTokens[i] == address(_token)) {
+                return false;
+            }
+        }
```
MicahZoltu: `.contains` seems like a good candidate for the `*Array` libraries.  This would allow us to unit test the `.contains` code separately from the business logic around getting protected tokens.  I know this is a minor gain, but I think generally a good abstraction.

MicahZoltu: Caveat: That would mean that all contracts that derive from this will need to compile in `AddressArray`, which also contains slice.  Due to the contract size limits, I wonder if we would be better off having a separate contract for each of the array methods?  Maybe the same for the safe math libraries?

```
contract Apple {
    using AddressArrayContains for address[];
    using Uint256SafeMathAdd for uint256;
    using Uint256SafeMathSub for uint256;

    function banana(address[] array, uint256 number) returns (uint256) {
        return array.sum().add(5).sub(7);
    }
}
````

The above pattern would allow us to shrink down some of the contracts that use a subset of SafeMath (every contract that uses SafeMath uses a subset of it) and it would allow us to be a bit more liberal with what we convert to libraries.

Note: We can put multiple contracts into the same file, so we could still have `SafeMathUint256.sol`, it would just have multiple contracts in it (one for each function).

**Comment by MicahZoltu** [comment]
Do we have a clubhouse issue to create a little script (my vote is TS, perhaps as part of the TS test suite) that looks at the ABI of every contract to make sure it has this new extension?  Lets put it into the backlog for now and decide how to schedule it later.

File: source/contracts/reporting/Market.sol [hunk 10]
```diff
@@ -680,4 +681,12 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
         }
         return _winningDestination.getUniverse().getParentPayoutDistributionHash();
     }
+
+    // Markets hold the initial fees paid by the creator in ETH and REP, so we dissallow ETH and REP extraction by the controller
```
MicahZoltu: Markets also hold all of the order escrow, which means the denomination token and all share tokens for this market.  This logic lives over in `Order.sol`.

MicahZoltu: It also escrows denomination token when complete sets are purchased in `CompleteSet.sol`.

source/contracts/Augur.sol
nuevoalex: Will change. This just worked better as a copy-paste-review format when doing all the boilerplate.

source/contracts/libraries/Extractable.sol
nuevoalex: Will make an issue to do this elsewhere at some point

source/contracts/libraries/Extractable.sol
nuevoalex: Noted in Slack too but for posterity: punting this till later and tracking as a CH issue

nuevoalex:
Primary blockers addressed and made many of the changes suggested. For anything not done I created a CH issue in the backlog, including the CI one.

source/contracts/Controlled.sol
```diff
@@ -45,9 +45,9 @@ contract Controlled is IControlled {
     function suicideFunds(address _target, ERC20Basic[] _tokens) public onlyControllerCaller returns(bool) {
         // Transfer tokens to target
         for (uint256 i = 0; i < _tokens.length; i++) {
-            ERC20Basic token = _tokens[i];
-            uint256 balance = token.balanceOf(this);
-            token.transfer(_target, balance);
+            ERC20Basic _token = _tokens[i];
+            uint256 _balance = _token.balanceOf(this);
+            _token.transfer(_target, _balance);
```
MicahZoltu: I know your PR didn't introduce this bug, but this should be wrapped in a `require()`.

nuevoalex: Good catch, will fix

source/contracts/reporting/Market.sol
```diff
@@ -680,4 +677,13 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
...
+    function getProtectedTokens() internal returns (address[] memory) {
+        address[] memory _protectedTokens = new address[](3);
+        _protectedTokens[0] = address(0);
+        _protectedTokens[1] = reportingWindow.getReputationToken();
+        _protectedTokens[0] = cash;
```
MicahZoltu: `[0]` should be `[2]`.

nuevoalex: Woops. that suggests tests are passing incorrectly too.

source/contracts/reporting/Market.sol
MicahZoltu: This is still missing all of the share tokens for the market, which are held in escrow by Market for open orders.  this means we'll need `i=3` through up to `i=8` for `_protectedTokens[i]`.

source/contracts/reporting/Market.sol
MicahZoltu: We should probably use `address(1)` instead of `address(0)` for the magic token, since `0` is the default if the collection is empty.

nuevoalex:
Share tokens restricted and using a special sentinel value for Ether protection.

Adding the share tokens pushed market over the gas limit which I addressed by consolidating the designated and first reporting implementations into one function. Probably for the best anyway.

Last reporting is different enough that it felt it should be it's own thing.

source/contracts/reporting/Market.sol
```diff
@@ -148,42 +149,36 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
...
-        IReportingWindow _newReportingWindow = getUniverse().getNextReportingWindow();
-        migrateReportingWindow(_newReportingWindow);
+        reportingWindow.getReputationToken().trustedMarketTransfer(msg.sender, _bond, _bondAmount);
+        if (_reportingState == ReportingState.DESIGNATED_DISPUTE) {
+            designatedReporterDisputeBond = _bond;
+            reportingWindow.updateMarketPhase();
+        } else {
+            firstReportersDisputeBond = _bond;
+            IReportingWindow _newReportingWindow = getUniverse().getNextReportingWindow();
+            migrateReportingWindow(_newReportingWindow);
```
MicahZoltu: No need to update market phase in this code path?  Feels like it should?

nuevoalex: migrateReportingWindow does it

PR #419, opened 2017-10-29
Switched to main pyethereum

joeykrug:


requirements.txt
```diff
@@ -7,7 +7,7 @@
 asn1crypto==0.23.0        # via coincurve
 cffi==1.11.1              # via coincurve
 coincurve==6.0.0          # via ethereum-augur-temp
-ethereum-augur-temp==2.1.10
+ethereum
```
MicahZoltu: This change should be made in `requirements.in` and then the script at the top of this file should be run to re-generate this.

MicahZoltu: Also, we should specify the version of `ethereum` we are using so a push to `ethereum` doesn't break our builds.

PR #420, opened 2017-10-29
added internal transfer method in BasicToken. Updated ReputationToken…

bthaile:
see motivation details in [ClubHouse](https://app.clubhouse.io/augur/story/4451/fixup-token-transfers).
changed ReputationToken to use new internal transfer method.

### DIFF
--- a/source/contracts/libraries/token/BasicToken.sol
+++ b/source/contracts/libraries/token/BasicToken.sol
@@ -28,6 +28,18 @@ contract BasicToken is ERC20Basic {
     }
 
     /**
+    * @dev allows internal token transfers
+    * @param _from The source address
+    * @param _to The destination address 
+    */
+    function internalTransfer(address _from, address _to, uint256 _value) internal returns (bool) {
+        balances[_from] = balances[_from].sub(_value);
+        balances[_to] = balances[_to].add(_value);
+        Transfer(_from, _to, _value);
+        return true;
+    }
+
+    /**
     * @dev Gets the balance of the specified address.
     * @param _owner The address to query the the balance of.
     * @return An uint256 representing the amount owned by the passed address.
--- a/source/contracts/libraries/token/BasicToken.sol
+++ b/source/contracts/libraries/token/BasicToken.sol
@@ -35,7 +35,7 @@ contract BasicToken is ERC20Basic {
     function internalTransfer(address _from, address _to, uint256 _value) internal returns (bool) {
         balances[_from] = balances[_from].sub(_value);
         balances[_to] = balances[_to].add(_value);
-        emitTransferLogs(_from, _to, _value);
+        Transfer(_from, _to, _value);
         return true;
     }
 
--- a/source/contracts/libraries/token/BasicToken.sol
+++ b/source/contracts/libraries/token/BasicToken.sol
@@ -35,7 +35,6 @@ contract BasicToken is ERC20Basic {
     function internalTransfer(address _from, address _to, uint256 _value) internal returns (bool) {
         balances[_from] = balances[_from].sub(_value);
         balances[_to] = balances[_to].add(_value);
-        Transfer(_from, _to, _value);
         emitTransferLogs(_from, _to, _value);
         return true;
     }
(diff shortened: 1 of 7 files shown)


source/contracts/libraries/token/BasicToken.sol
```diff
@@ -28,6 +28,18 @@ contract BasicToken is ERC20Basic {
...
 
     /**
+    * @dev allows internal token transfers
+    * @param _from The source address
+    * @param _to The destination address 
+    */
+    function internalTransfer(address _from, address _to, uint256 _value) internal returns (bool) {
+        balances[_from] = balances[_from].sub(_value);
+        balances[_to] = balances[_to].add(_value);
+        Transfer(_from, _to, _value);
```
MicahZoltu: `emitTransferLog` should be called after `Transfer` (assuming you don't rename it per CH issue).

source/contracts/libraries/token/BasicToken.sol
```diff
@@ -28,6 +28,18 @@ contract BasicToken is ERC20Basic {
     }
 
     /**
+    * @dev allows internal token transfers
+    * @param _from The source address
+    * @param _to The destination address 
+    */
+    function internalTransfer(address _from, address _to, uint256 _value) internal returns (bool) {
```
MicahZoltu: The usage of this function should be applied throughout the codebase, as I suspect there are similar bugs elsewhere (and if there aren't there is opportunity for bugs elsewhere).  This could be done as a follow-up PR if you prefer that though.

source/contracts/libraries/token/BasicToken.sol
bthaile: Would this in effect cause 2 Transfer events to happen?

source/contracts/libraries/token/BasicToken.sol
bthaile: I'd prefer a follow up PR.

source/contracts/libraries/token/BasicToken.sol
MicahZoltu: Yes, it will in some cases. One of the transfer events will be to maintain ERC20 spec, the other would be what we use in Augur Node for tracking transfers.

bthaile: Change made

source/contracts/libraries/token/BasicToken.sol
```diff
@@ -28,6 +28,19 @@ contract BasicToken is ERC20Basic {
...
+        balances[_from] = balances[_from].sub(_value);
+        balances[_to] = balances[_to].add(_value);
+        Transfer(_from, _to, _value);
+        emitTransferLogs(_from, _to, _value);
```
MicahZoltu: At the moment, `emitTransferLogs(...)` in most cases emits `Transfer` event (as well as some other event).  Either we should change this function to only call `emitTransferLogs` and not also emit the `Transfer` event, or we need to tackle one or more of the other tasks in the ClubHouse issue and fix the other call sites.

If you prefer small isolated PRs, then I think removing the `Transfer` event from this function is the right course of action (and in hindsight, perhaps what you meant when you said "fired twice") and then in a follow-up PR address the "`Transfer` can be missed" problem.

bthaile: removed the Transfer call in internalTransfer. Currently working on adding unit tests.

tests/libraries/token/test_basic_token.py
```diff
@@ -0,0 +1,66 @@
...
+    assert mockToken.balanceOf(tester.a1) == 100
+    assert mockToken.balanceOf(tester.a2) == 0
+    with raises(AttributeError, message="can not call internal method"):
+        mockToken.internalTransfer(tester.a1, tester.a2, 100)
+
+    with raises(TransactionFailed, message="can not cause negative balance"):
+        mockToken.callInternalTransfer(tester.a1, tester.a2, 150)
+
+    with raises(TransactionFailed, message="can not cause negative balance"):
+        mockToken.callInternalTransfer(tester.a1, tester.a2, 150)
```
MicahZoltu: This appears to be identical to the check immediately above.

bthaile: yes, dup. will remove

PR #421, opened 2017-10-29
#3936 reputation token tests

bthaile:
unit tests for reputation token.
slimmed down creating mock test fixture, by using existing pattern of loading mock contracts.
added more call methods to existing mocks
removed offending/bug line from ReputationToken that adds to total supply on internalTrustedTransfer

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -269,6 +269,19 @@ class ContractsFixture:
                 else:
                     self.uploadAndAddToController(path.join(directory, filename))
 
+    def uploadAllMockContracts(self):
+        for directory, _, filenames in walk(resolveRelativePath('solidity_test_helpers')):
+            for filename in filenames:
+                name = path.splitext(filename)[0]
+                extension = path.splitext(filename)[1]
+                if extension != '.sol': continue
+                if not name.startswith('Mock'): continue
+                if 'Factory' in name:
+                    self.upload(path.join(directory, filename))    
+                else:
+                    self.uploadAndAddToController(path.join(directory, filename))
+
+
     def whitelistTradingContracts(self):
         for filename in listdir(resolveRelativePath('../source/contracts/trading')):
             name = path.splitext(filename)[0]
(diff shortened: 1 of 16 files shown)


tests/conftest.py
```diff
@@ -419,6 +432,15 @@ def augurInitializedSnapshot(fixture, controllerSnapshot):
     return fixture.createSnapshot()
 
 @pytest.fixture(scope="session")
+def augurInitializedWithMocksSnapshot(fixture, augurInitializedSnapshot):
+    fixture.uploadAndAddToController("solidity_test_helpers/Constants.sol")
+    fixture.uploadAllMockContracts()
```
MicahZoltu: We have been, generally, trying to move away from having the tests "upload everything" and more towards a pattern of "use a fixture that just has what you need" in an effort to get localized test runs sped up.  That is, if you just want to run the `test_controlled.py` suite, you don't have to wait for a ton of unrelated things to get uploaded.

The question I have for you regarding this change is, do you find that there are many test fixtures that actually need all of the mocks available, or do most test fixtures just need a couple mocks?  If it is fairly common to need all (or at least the majority) of the mocks available then I think this change is fine.  If most of the time though we only need a few mocks then I would say this change is OK, but be careful to not over-use it.

bthaile: I was noticing that unit tests need all the same mocks, universe, reputation token, market , ... The ones not needed all the time are factory but they are small mock files.

tests/solidity_test_helpers/MockDisputeBondToken.sol
```diff
@@ -57,6 +57,14 @@ contract MockDisputeBondToken is ITyped, IDisputeBond {
         return _market.decreaseExtraDisputeBondRemainingToBePaidOut(_amount);
     }
 
+    function callMigrateOutDisputeBondToken(IReputationToken _reputationToken, IReputationToken  _destination, address _reporter, uint256 _attotokens) public returns(bool) {
+        return _reputationToken.migrateOutDisputeBondToken(_destination, _reporter, _attotokens);
+    }
+
+    function callMintForDisputeBondMigration(IReputationToken _reputationToken, uint256 _attotokens) public returns(bool) {
+        return _reputationToken.mintForDisputeBondMigration(_attotokens);
+    }
```
MicahZoltu: @nuevoalex has a PR out that makes `DisputeBond` not a token.  Depending on who merges first, this may need to be deleted.

tests/solidity_test_helpers/MockLegacyReputationToken.sol
```diff
@@ -0,0 +1,23 @@
...
+    using ContractExists for address;
+    uint256 private faucetAmountValue;
+
+    function getFaucetAmountValue() public returns(uint256) {
+        return faucetAmountValue;
+    }
+
+    function MockLegacyReputationToken() public {
+
+    }
```
MicahZoltu: Empty constructors aren't necessary in Solidity, you can just remove the constructor all together.

tests/solidity_test_helpers/MockDisputeBondToken.sol
bthaile: The race is on!!!

PR #424, opened 2017-10-29
Fix ordering of _tokensEscrowed & _sharesEscrowed in signature of getOrder function

adrake33:


MicahZoltu:
I think this method isn't actually used anywhere, so we can just delete it.  Lets double check with @tinybike first though to make sure the UI/AugurNode isn't using it.  If they are, we should find out which order they are expecting the parameters and use that.

tinybike:
Correct, augur-node doesn't use `getOrder`, feel free to delete!

adrake33:
@MicahZoltu It looks like we were using the `getOrder` function in some of the asserts in the Python tests. For example, to confirm if an order was canceled (although we're also checking the return value from calling `cancelOrder`). Should I still go ahead and delete this function and remove the calls to it in the Python tests?

MicahZoltu:
If the tests were just testing `getOrder`, then yeah, delete the assertion or the test.  If `getOrder` is a means to an end and the tests are testing something else, then you can probably replace it with one or more of the `Orders.get*` methods.

PR #425, opened 2017-10-30
change to use new internalTransfer internal method to consolidate tra…

bthaile:
Only found one place where transfer logic could be consolidated to new internalTransfer internal method.

### DIFF
--- a/source/contracts/libraries/token/StandardToken.sol
+++ b/source/contracts/libraries/token/StandardToken.sol
@@ -29,15 +29,10 @@ contract StandardToken is ERC20, BasicToken {
     function transferFrom(address _from, address _to, uint256 _value) public returns (bool) {
         uint256 _allowance = allowed[_from][msg.sender];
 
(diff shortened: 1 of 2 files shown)


source/contracts/libraries/token/StandardToken.sol
```diff
@@ -29,15 +29,10 @@ contract StandardToken is ERC20, BasicToken {
     function transferFrom(address _from, address _to, uint256 _value) public returns (bool) {
         uint256 _allowance = allowed[_from][msg.sender];
 
-        // Check is not needed because sub(_allowance, _value) will already throw if this condition is not met
-        // require (_value <= _allowance);
-
-        balances[_to] = balances[_to].add(_value);
-        balances[_from] = balances[_from].sub(_value);
+        internalTransfer(_from, _to, _value);
```
MicahZoltu: We should put this line after the allowance check.

bthaile: made change

PR #427, opened 2017-10-30
Add mint and burn special logging handling support and make Rep token…

nuevoalex:
… burn and mint instead of manually editing supply in migration functions

### DIFF
--- a/source/contracts/libraries/token/VariableSupplyToken.sol
+++ b/source/contracts/libraries/token/VariableSupplyToken.sol
@@ -18,7 +18,7 @@ contract VariableSupplyToken is StandardToken {
     function mint(address _target, uint256 _amount) internal returns (bool) {
         balances[_target] = balances[_target].add(_amount);
         supply = supply.add(_amount);
-        Mint(_target, _amount);
+        onMint(_target, _amount);
         return true;
     }
 
@@ -30,6 +30,18 @@ contract VariableSupplyToken is StandardToken {
     function burn(address _target, uint256 _amount) internal returns (bool) {
         balances[_target] = balances[_target].sub(_amount);
         supply = supply.sub(_amount);
+        onBurn(_target, _amount);
+        return true;
+    }
+
+    // Subclasses of this token may want to send additional logs through the centralized Augur log emitter contract
+    function onMint(address _target, uint256 _amount) internal returns (bool) {
+        Mint(_target, _amount);
+        return true;
+    }
+
+    // Subclasses of this token may want to send additional logs through the centralized Augur log emitter contract
+    function onBurn(address _target, uint256 _amount) internal returns (bool) {
         Burn(_target, _amount);
         return true;
     }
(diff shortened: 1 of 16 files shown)


source/contracts/libraries/token/VariableSupplyToken.sol
```diff
@@ -30,6 +30,18 @@ contract VariableSupplyToken is StandardToken {
     function burn(address _target, uint256 _amount) internal returns (bool) {
         balances[_target] = balances[_target].sub(_amount);
         supply = supply.sub(_amount);
+        onBurn(_target, _amount);
+        return true;
+    }
+
+    // Subclasses of this token may want to send additional logs through the centralized Augur log emitter contract
+    function onMint(address _target, uint256 _amount) internal returns (bool) {
+        Mint(_target, _amount);
```
MicahZoltu: Similar to how `Transfer` event should be fired independent of overrides, I think the `Mint`/`Burn` functions should be fired independent of overrides.  That is, we should always fire a `Mint`/`Burn` event from the token contract, as well as doing whatever else the derived contract wants to do on mint/burn (such as firing another event from Augur.sol.

The argument here is weaker, since Mint/Burn are not part of the ERC20 spec, but I think there is value in them being readily available to things like chain explorers, especially for REP which will most likely be a well known contract address.

MicahZoltu: TL;DR: move `Mint` and `Burn` out of the `onMint` and `onBurn` functions and into the `mint` and `burn` functions right before the `onMint`/`onBurn` function is called.

nuevoalex: Should we do that for `Transfer` as well? currently that is not the pattern being used

MicahZoltu:
Is `ReputationToken` our only `VariableSupplyToken` (aside from the test token faucets).

source/contracts/libraries/token/VariableSupplyToken.sol
MicahZoltu: Yes, I believe @bthaile is working on a PR to do that as part of https://app.clubhouse.io/augur/story/4451/fixup-token-transfers

MicahZoltu: I realize now that you may not have read that CH issue prior to me mentioning it just now.  😛

nuevoalex:
No, most of our tokens are `VariableSupplyToken`s actually. We can add this to all of them if desired.

nuevoalex:
Actually we'll have to add this to all `VariableSupplyTokens` with the above change since `onMint` and `onBurn` will have to become abstract methods to avoid compiler warnings.

MicahZoltu:
👍 on making it abstract and implementing it everywhere.  Per conversation in Slack, this will help us ensure that all of our tokens properly log via Augur.sol.

nuevoalex:
Changes made. I also added the mint/burn target to the logs since its relevant and might be useful at some point.

PR #428, opened 2017-10-31
Now with client-side signing!

MicahZoltu:
Adds an environment variable to integration tests for private key.  This is the key that will be used to sign all transactions submitted to the chain, rather then relying on the chain having an account unlocked.

This ended up being a pretty big refactor along the way, but isolated to integration test code.  The biggest change is the new `libraries/ContractInterfaces.ts`.  This is hand coded at the moment, but it is designed so that it can easily be auto-generated from the contract ABI output.  Authoring such a code generator should be pretty easy as its just a matter of mapping solc types around and looking at stateMutability.  The big win from this pattern over the previous pattern is that we get dev-time type checking and IDE auto-complete for all of our contracts (once code generator is authored) in TS.

Note: that contract method names ending with `_` are local calls.  Contract methods that are stateMutability of `view` or `pure` only have this overload, while contract methods that are `payable` or `nonpayable` have two copies of the method, one with `_` suffix and one without so the caller can decide whether they want to get the result or run the call on-chain.  I'm open to ideas to better ways to express this.  I originally went with a function parameter but that got kind of ugly fast.

I also abstracted out account management, connection management and configuration.  This was mainly to remove some code duplication that I was introducing as a side effect of the private key changes.

At this point we are just using the basics of ethjs, none of the more complex stuff like contract creation and call mapping.  I also removed the ABIParser class as I find the new statically typeded contract accessors to be far superior.

Finally, I tightened up a number of type definition files for some of the ethjs libraries we are using, as well as solc.  At the moment there is quite a bit of type duplication that we should take care of, but it is not really causing trouble at the moment so I have put it off once again.


source/libraries/Connector.ts
```diff
@@ -1,35 +1,24 @@
...
 
     public async waitUntilConnected(): Promise<EthjsQuery> {
-        const ethjsQuery = await this.getEthjsQuery();
         while (true) {
             try {
-                await ethjsQuery.net_version();
+                await this.ethjsQuery.net_version();
                 break;
             } catch {
                 // swallow the error, we just want to loop until the net_version call above succeeds.
```
pgebheim: I'd say check for the specific error you're expecting to swallow here, too likely that something upstream breaks and we end up swallowing an error that we're not expecting.

source/deployment/deployContracts.ts
```diff
@@ -1,45 +1,46 @@
...
 
     return contractDeployer;
 }
 
 // If this script is not being imported by another module (i.e., it is being run independently via the command line)
-if (!module.parent) {
+async function runStandalone() {
     require('source-map-support').install();
-    compileAndDeployContracts().then(() => {
+    const configuration = new Configuration(await getPort());
```
pgebheim: I would move the `await getPort()` into Configuration so it can be co-located with the logic around undefined `ETHEREUM_HOST` && using TestRPC.

It's confusing to use `getPort` in a case where the seeming use-case is connecting to an external host -- not spinning up a server yourself.

MicahZoltu: I did that originally, but `getPort` is async and you can't have an async constructor in TS, which would mean two-step construction.  What I can do though, now that I have slept and can think again, is add a `ConfigurationFactory()` method that is async that gets the port and returns a newly constructed `Configuration` object.

source/libraries/Connector.ts
MicahZoltu: Normally I would agree with you on this, but the problem with that strategy in this case is that there is a whole slew of possible errors you might get back and I'm not confident I could reliably enumerate them all.  Depending on the current state of the node at the time the request happens to land, and the node you are talking to (Geth, Parity, TestRPC), you could get anything from a socket timeout, TCP connection error, JSON-RPC error, or any other kind of transport/protocol error at any of many different layers of the stack.  `this.ethjsQuery.net_version` is picked as a very simple method that really has no failure conditions other than "remote side not on or functioning correctly", so any error we get back is indicative of either a programming error (e.g., `this.ethjsQuery` was constructed with a bad host/port) or the remote end is not on or in a bad state or is not an Ethereum node.

source/deployment/deployContracts.ts
MicahZoltu: Made the changes to have a factory function.  Check them out and let me know if that makes you feel better.

nuevoalex:
What do you think of maybe changing the `_` to `_local`? The `_` alone seems easy to miss potentially.

MicahZoltu:
The point of the `_` _is_ to be easy to miss.  :)  The hope was to avoid having already very verbose function names become even more long/verbose.  We would also have to deal with naming conflicts should we ever name a contract function `...Local` (though this seems fairly unlikely).  As a caller, what you care about is whether the function returns a result or returns a transaction hash.  I toyed with generics as another option, but you can't switch on generic types in TS so that didn't pan out.  We could do an extra required parameter and have method overloads, but that suffers the "callsites quickly get ugly" problem when half of your callsites have a `true` parameter that isn't actually part of the function.

That is my argument for using a trailing `_`, if after reading that you still think a `Local` suffix would be better we can move forward with that option.  If we take that path, should we also have a `Remote` suffix for the other set of transactions so there is symmetry?  Some functions will be just local (view/pure), the rest will be both local & remote (since all of our functions return a value).  I generally like symmetry, but I also am trying to simplify call sites here so I'm undecided.

nuevoalex:
I'm not strongly opinionated enough on it to really insist on that change; I just wanted to float the idea of being more explicit in the convention to see what thoughts were. I'll defer to others.

source/libraries/Configuration.ts
```diff
@@ -0,0 +1,24 @@
...
+    public readonly privateKey: string;
+
+    public constructor(host: string, port: number, gasPrice: BN, privateKey: string) {
+        this.httpProviderHost = host;
+        this.httpProviderPort = port;
+        this.gasPrice = gasPrice;
+        this.privateKey = privateKey;
+    }
+
+    public static create = async (): Promise<Configuration> => {
```
adrake33: Is this the syntax that has to be used for static async functions?
(As opposed to `public static async create(): Promise<Configuration> {`)?

source/libraries/ContractDeployer.ts
```diff
@@ -1,42 +1,44 @@
...
-    public testAccounts;
-    public controller;
-    public universeAddress: string;
-    public market;
-
-    public constructor(ethjsQuery: EthjsQuery, compilerOutput: CompilerOutputContracts, gasPrice: number) {
-        this.ethjsQuery = ethjsQuery;
-        this.ethjsContract = new EthjsContract(ethjsQuery);
-        this.gasPrice = gasPrice;
+    private nonce: BN | null = null;
```
adrake33: Why is this `null = null` (as opposed to just `null`)?

source/libraries/HelperFunctions.ts
```diff
@@ -47,30 +41,10 @@ export function stringTo32ByteHex(stringToEncode: string): string {
...
-    }
-
-    return await Promise.all(testAccounts);
-}
-
-export async function initializeTestRpcClientOptions(ethHttpProviderPort: number, secretKeys: string[]): Promise<RpcClientOptions> {
-    // Generate test accounts
-    const testAccounts = await generateTestAccounts(secretKeys);
-
+export function initializeTestRpcClientOptions(ethHttpProviderPort: number, privateKey: string): RpcClientOptions {
```
adrake33: It looks like `ethHttpProviderPort` is not being used, so I think this parameter can be removed.

source/libraries/HelperFunctions.ts
```diff
@@ -1,18 +1,12 @@
 import * as binascii from "binascii";
-import EthjsAccount = require("ethjs-account");
 import EthjsQuery = require('ethjs-query');
 import { TransactionReceipt } from 'ethjs-shared';
+import { Account } from 'ethjs-account';
```
adrake33: Is `Account` being used anywhere?

adrake33:
I don't really have a better suggestion regarding the trailing underscore convention, so I suppose I'm ok with it.  My biggest concern is that the convention won't be terribly obvious to someone who's new to the code.  Perhaps this could be documented somewhere?  I realize the goal is to eventually auto-generate ContractInterfaces.ts, so if we were to put a comment in there, the code that eventually generates that file would have to add the comment every time it gets generated.  I'm not sure if there's any better place to document this convention, though.

source/libraries/ContractDeployer.ts
nuevoalex: I think `BN | null` is the syntax to make it nullable

adrake33: Ah, yeah, I see.  It's assigning it to `null` after the `BN | null` part.  Never mind. :)

source/libraries/Configuration.ts
MicahZoltu: This style makes it so the function has a captured `this` so you don't have to worry about JavaScript stupid this re-binding problems.  In general, I prefer this style over `public foo(): Type {...}` for that reason.

MicahZoltu:
I applied all of your feedback @adrake33.  I also fixed some solc warnings introduced by previous PRs.

PR #429, opened 2017-10-31
completing the tasks in CH, renamed emit.. to onTokenTransfer. Create…

bthaile:
If I got this correct. renamed emit.. to onTokenTransfer on BasicToken which forces derived contracts to impl it. There they can impl any special logging or logic specific to the derived contract without messing with parent class.

### DIFF
--- a/source/contracts/trading/Cash.sol
+++ b/source/contracts/trading/Cash.sol
@@ -58,4 +58,8 @@ contract Cash is Controlled, Extractable, ITyped, VariableSupplyToken, ICash {
     function getProtectedTokens() internal returns (address[] memory) {
         return new address[](0);
     }
+
+    function onTokenTransfer(address _from, address _to, uint256 _value) internal returns (bool) {
+        return true;
+    }    
 }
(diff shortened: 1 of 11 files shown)


source/contracts/trading/Cash.sol
```diff
@@ -58,4 +58,8 @@ contract Cash is Controlled, Extractable, ITyped, VariableSupplyToken, ICash {
     function getProtectedTokens() internal returns (address[] memory) {
         return new address[](0);
     }
+
+    function onTokenTransfer(address _from, address _to, uint256 _value) internal returns (bool) {
+        return true;
```
MicahZoltu: @nuevoalex We may want to have `Cash` log via Augur, just like the other trading contracts.  @epheph?

tests/trading/test_shareToken.py
```diff
@@ -114,7 +114,7 @@ def test_approve(contractsFixture, market):
             "to": bytesToHexString(tester.a1),
             "universe": market.getUniverse(),
             "value": 7
-        },
+        }
```
MicahZoltu: Style: Remove trailing whitespace.

MicahZoltu: Also style, though not one I think we have discussed. I prefer trailing commas as it makes re-arranging things easier.

PR #431, opened 2017-11-01
Extracts test methods out of deployment code.

MicahZoltu:
In order to make this go smoothly, I greatly increased the strictness of TS, and then fixed all of the errors.  I also upgraded to TS 2.6 which gave us some new static analysis.

I still have a number of cleanup related things I would like to do, but this addresses the biggest concern for getting deployment CI working properly, which was that there was test code happening as part of deployment (creating a market).


bthaile:
Would there be any advantage to have a "Report" to verify that uploads/system is up and configured properly? Not sure of a suite of tests would do it or just a 'roll-call' so to speak would do it.

MicahZoltu:
@bthaile 

If I understand what you mean by "Report" correctly, then I think that aligns with our plans on integration tests.  The idea is that after you deploy you can then run the integration tests against that chain and they should all pass.  At the moment, the integration tests don't do anything interesting, but long term the goal would be to have them create a market, trade on the market, report on the market, dispute the market, fork the market, resolve the market, etc.  This would allow us to fully validate that the uploaded contracts are fully functional.

Such a test suite wouldn't be as thorough as the unit test suite in terms of coverage due to run time being _much_ slower and costs being _much_ higher, but it should try give us general confidence that the contracts did in fact upload correctly and everything is properly added to controller/whitelists.

Does this align with what you had in mind, or am I missing the point of your comment?

bthaile:
Nailed my point. A light integration test suite would give confidence and keep gas cost down.

MicahZoltu:
Does this mean you approve my PR?!  😁

source/tests-integration/TestFixture.ts
```diff
@@ -0,0 +1,91 @@
...
+        // TODO: just get enough REP to cover the bonds rather than over-allocating
+        const repFaucetTransactionHash = await legacyReputationToken.faucet(new BN(0));
+        const repApprovalTransactionHash = await legacyReputationToken.approve(reputationTokenAddress, new BN(2).pow(new BN(256)).sub(new BN(1)));
+        await this.connector.waitForTransactionReceipt(repFaucetTransactionHash, `Using legacy reputation faucet.`);
+        await this.connector.waitForTransactionReceipt(repApprovalTransactionHash, `Approving legacy reputation.`);
+        const repMigrationTransactionHash = await reputationToken.migrateFromLegacyReputationToken();
+        // necessary because it is used part of market creation fee calculation
+        const currentReportingWindowTransactionHash = await universe.getCurrentReportingWindow();
+        // necessary because it is used as part of market creation fee calculation
+        const previousReportingWindowTransactionHash = await universe.getPreviousReportingWindow();
```
nuevoalex: Mostly just curious about this. Would you also be able to just call `universe.getMarketCreationCost()` instead of `universe.getMarketCreationCost_()` and these prep calls?

MicahZoltu: Calling a remote function does not give you the return value.  Calling a local function does not cause it to mutate storage.  For some functions, we need to both mutate storage and get the return value, which is why we have to do _both_ a local call and a remote call.  See the giant discussion in #dev filled with @tinybike  😡 for details on why this is.  😄

nuevoalex: Ah yea that makes sense now reading it.

PR #434, opened 2017-11-03
Makes the deploy script skip uploading contracts that don't match the file name.

MicahZoltu:
This prevents us from attempting to upload helper library contracts that are included in a file with the primary contract in that file.  An alternitave would be to continue uploading them, but restructure how we do lookup keys.  I'm not actually sure what is best, but this solution was simple and fixes a bug where we were uploading a contract but then clobbering it in our internal data structures with the next (`FillOrder.sol`).

### DIFF
--- a/source/libraries/ContractDeployer.ts
+++ b/source/libraries/ContractDeployer.ts
@@ -130,6 +130,10 @@ export class ContractDeployer {
             }
 
             for (let contractName in this.compiledContracts[contractFileName]) {
+                // Filter out any contracts that don't match the file name so helper libraries are skipped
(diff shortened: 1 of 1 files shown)


source/libraries/ContractDeployer.ts
```diff
@@ -130,6 +130,10 @@ export class ContractDeployer {
             }
 
             for (let contractName in this.compiledContracts[contractFileName]) {
+                // Filter out any contracts that don't match the file name so helper libraries are skipped included
```
nuevoalex: S/skipped included/skipped

PR #436, opened 2017-11-05
Makes controller registry store contract name mapped to address, commit_hash, and file_sha.

MicahZoltu:
This change allows us to make scripts that are much smarter in how they upload contracts and also allows reviewers to easily go from on-chain contracts to source code with some basic means of validation along the way.

At the moment, the scripts just set the commit/file hashes but the next step (future PR) is to make the script look to see if a matching contract is already uploaded and if so only upload/update contracts that haven't yet been uploaded or have changed (different file sha) since they were uploaded.

I chose to change the controller registry mapping to point at a contract rather than having multiple mappings for clarity.  I also changed the `registerContract` (previously `setValue`) function to be `onlyOwnerCaller` rather than `onlyWhitelistedCallers` since that is a more appropriate constraint.


MicahZoltu:
Added @joeykrug and @tinybike as reviewers since this is a code change (to `Controller` registry).  Per our discussion in #core-dev the other day, it sounds like everyone is on board and the change is very small (just some extra metadata and an accessibility restriction) so I don't _think_ it is contentious but want to make sure no one is caught by surprise.

This change will allow us to do incremental contract deploys/upgrades which I think is important now that we are actually putting these contracts on real blockchains.

source/contracts/Controller.sol
```diff
@@ -8,9 +8,17 @@ import 'libraries/Extractable.sol';
 
 
 contract Controller is IController {
+    struct ContractDetails {
+        bytes32 name;
```
nuevoalex: Why store the name here?

source/contracts/Controller.sol
```diff
@@ -67,22 +75,27 @@ contract Controller is IController {
...
     }
 
-    function removeValue(bytes32 _key) public onlyWhitelistedCallers returns(bool) {
-        registry[_key] = address(0);
+    function getContractDetails(bytes32 _key) public view returns (address, bytes20, bytes32) {
+        ContractDetails storage _details = registry[_key];
+        return (_details.contractAddress, _details.commitHash, _details.fileHash);
+    }
+
+    function unregisterContract(bytes32 _key) public onlyWhitelistedCallers returns(bool) {
```
nuevoalex: Should this also be `onlyOwnerCaller` now?

source/contracts/Controller.sol
MicahZoltu: My thought was that now that Solidity supports returning/passing structs between functions that we may use this structure to pass contract details around in the future.  By putting the name into the struct, this allows us to have a single object that contains all of the relevant details since you may not always get one of these structures by looking it up in the mapping directly.

If you think this is too much of a premature optimization I don't mind removing it.  I don't care much about the increased gas costs since this cost is only incurred at contract upload time, where we really don't care at all about gas costs.

source/contracts/Controller.sol
MicahZoltu: Yes, will fix.

source/contracts/Controller.sol
nuevoalex: Sounds good! Was just curious on the intent since it isn't used yet.

PR #437, opened 2017-11-05
#3936 market unit tests

bthaile:
init PR for market unit tests. I wanted to break up the PRs because market has a lot of logic. The PR includes initial create and testing getters and disputes (designated, first, last).
Added needed mocks and modified existing mocks to check methods were called.


tests/reporting/test_market_unit.py
```diff
@@ -0,0 +1,527 @@
...
+        market.initialize(mockReportingWindow.address, endTime, 1, numTicks, fee, mockCash.address, tester.a1, tester.a1)
+
+    with raises(TransactionFailed, message="outcomes has to be less than 9"):
+        market.initialize(mockReportingWindow.address, endTime, 9, numTicks, fee, mockCash.address, tester.a1, tester.a1)
+
+    with raises(TransactionFailed, message="numTicks needs to be divisable by outcomes"):
+        market.initialize(mockReportingWindow.address, endTime, 7, numTicks, fee, mockCash.address, tester.a1, tester.a1)
+
+    with raises(TransactionFailed, message="fee per eth can not be greater than max fee per eth in attoEth"):
+        market.initialize(mockReportingWindow.address, endTime, 5, numTicks, 5, mockCash.address, tester.a1, tester.a1)
```
nuevoalex: I don't think this is failing because of the fee max requirement. Note the max is denominated in `ether` so it's actually 10**18 / 2. Also the test below are using this fee amount, which while fine (I think) would be wrong for the test if this was actually failing because of the fee amount

tests/reporting/test_market_unit.py
```diff
@@ -0,0 +1,527 @@
...
+    mockReputationToken.setBalanceOf(0)
+    mockUniverse.setDesignatedReportNoShowBond(100)
+    mockReportingWindow.setReputationToken(mockReputationToken.address)
+    with raises(TransactionFailed, message="reporting window reputation token does not have enough balance"):
+        market.initialize(mockReportingWindow.address, endTime, 5, numTicks, 5, mockCash.address, tester.a1, tester.a1)
+    
+    mockReputationToken.setBalanceOf(100)
+    mockUniverse.setTargetReporterGasCosts(15)
+    mockUniverse.setValidityBond(12)
+    with raises(TransactionFailed, message="refund is not over 0"):
```
nuevoalex: over or equal to 0 just to be nit picky. Really the throw is because they didn't provide sufficient ETH to pay the bond costs

tests/reporting/test_market_unit.py
```diff
@@ -0,0 +1,527 @@
...
+    mockNextReportingWindow = localFixture.upload('solidity_test_helpers/MockReportingWindow.sol', 'mockNextReportingWindow');
+    mockUniverse.setNextReportingWindow(mockNextReportingWindow.address)
+
+    mockDisputeStakeToken = set_mock_stake_token_value(localFixture, initializeMarket, [0, 0, numTicks, 0, 0], mockStakeTokenFactory)
+    mockDisputeStakeToken.setTotalSupply(1000)
+
+    # execute disputeFirstReporters on market
+    assert initializeMarket.disputeFirstReporters([0, 0, numTicks, 0, 0], 50, False, sender=tester.k4)
+
+def push_to_designated_report_state(localFixture, mockUniverse, chain, initializeMarket, mockStakeToken, payoutDesignatedNumerators, mockStakeTokenFactory, mockReportingWindow, mockReputationToken, constants):
```
nuevoalex: Since at the end of this call the market is in designated dispute I think this should be renamed `push_to_designated_dispute_state`

tests/reporting/test_market_unit.py
```diff
@@ -0,0 +1,527 @@
...
+    mockSecondStakeToken.setTotalSupply(20)
+    assert initializeMarket.getTotalStake() == 0
+    assert mockSecondStakeToken.callIncreaseTotalStake(initializeMarket.address, 20)
+    assert initializeMarket.getTotalStake() == 20
+    assert mockReportingWindow.getIncreaseTotalStakeCalled() == True
+
+    assert mockSecondStakeToken.callUpdateTentativeWinningPayoutDistributionHash(initializeMarket.address, payoutSecondHashDisputeValue)
+    assert initializeMarket.getTentativeWinningPayoutDistributionHash() == initializeMarket.derivePayoutDistributionHash(payoutDesignatedNumerators, False)
+    assert initializeMarket.getBestGuessSecondPlaceTentativeWinningPayoutDistributionHash() == initializeMarket.derivePayoutDistributionHash(payoutSecondNumeratorsDispute, False)
+    # buy a thrid stake token
```
nuevoalex: s/thrid/third

tests/reporting/test_market_unit.py
```diff
@@ -0,0 +1,527 @@
...
+    mockThridStakeToken = set_mock_stake_token_value(localFixture, initializeMarket, payoutThridNumerators, mockStakeTokenFactory)
+    mockThridStakeToken.setTotalSupply(200)    
+    # verify stake tokens have correct value
+    assert initializeMarket.getPayoutDistributionHashStake(payoutDesignatedHashValue) == 105
+    assert initializeMarket.getPayoutDistributionHashStake(payoutSecondHashDisputeValue) == 20
+    assert initializeMarket.getPayoutDistributionHashStake(payoutThridHashValue) == 200
+
+    assert mockThridStakeToken.callUpdateTentativeWinningPayoutDistributionHash(initializeMarket.address, payoutThridHashValue)
+    assert initializeMarket.getTentativeWinningPayoutDistributionHash() == initializeMarket.derivePayoutDistributionHash(payoutThridNumerators, False)
```
nuevoalex: I think rather than doing derive calls through the test you can just use the already derived hashes from above. Like in this case `payoutThirdHashValue`

tests/reporting/test_market_unit.py
bthaile: I'll make the correction to passed in onEther / 2 + 1.

tests/reporting/test_market_unit.py
bthaile: good point, will make change

tests/reporting/test_market_unit.py
bthaile: Texas accent has alway screwed up my spelling, doh!
I'll make the corrections.

PR #438, opened 2017-11-06
Makes ContractDeployer do incremental uploads.

MicahZoltu:
If the currently uploaded contract (as registered with Controller) has the same bytecode sha as the bytecode to be uploaded then skip uploading.  Similarly, skip whitelisting/initialization if it is already done for the target contract.  Contains some pretty significant refactoring along the way, recommend ignoring the changeset for `ContractDeployer` and instead just reading the code directly.

I also cleaned up some of the definition files that were duplicated.

### DIFF
--- a/source/libraries/Contracts.ts
+++ b/source/libraries/Contracts.ts
@@ -1,0 +1,53 @@
+import { Abi } from 'ethereum';
+import { CompilerOutput } from 'solc';
+
+export class Contract {
+    public readonly relativeFilePath: string;
+    public readonly contractName: string;
+    public readonly abi: Abi;
+    public readonly bytecode: Buffer;
+    public address?: string;
+
+    public constructor(relativeFilePath: string, contractName: string, abi: Abi, bytecode: Buffer) {
+        this.relativeFilePath = relativeFilePath;
+        this.contractName = contractName;
+        this.abi = abi;
+        this.bytecode = bytecode;
+    }
+}
+
+export class Contracts implements Iterable<Contract> {
+    private readonly contracts = new Map<string, Contract>();
+
+    public constructor(compilerOutput: CompilerOutput) {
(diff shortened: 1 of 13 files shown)


source/libraries/Contracts.ts
```diff
@@ -0,0 +1,53 @@
...
+    public constructor(compilerOutput: CompilerOutput) {
+        for (let relativeFilePath in compilerOutput.contracts) {
+            for (let contractName in compilerOutput.contracts[relativeFilePath]) {
+                // don't include helper libraries
+                if (!relativeFilePath.endsWith(`${contractName}.sol`)) continue;
+                const abi = compilerOutput.contracts[relativeFilePath][contractName].abi;
+                if (abi === undefined) continue;
+                const bytecodeString = compilerOutput.contracts[relativeFilePath][contractName].evm.bytecode.object;
+                if (bytecodeString === undefined) continue;
+                // don't include interfaces
```
nuevoalex: This is probably something we should be doing in the python tests as well but I think it would be nice to throw here if the contract name doesn't begin with 'I' followed by a capital character. I've been bit somewhat frequently by non-interface contracts falling into this case from not properly fulfilling abstracts/interfaces.

MicahZoltu: https://app.clubhouse.io/augur/story/4631/

Per discussion, I like this idea but would require changes to the contracts to apply naming convention for abstract contracts everywhere.  Implementation at that point is quite simple:
```
if (bytecodeString.length === 0 && !contractName.startsWith('I')) throw new Error(...);
```

Putting off to future PR.

MicahZoltu:
I'm not really worried.  It is test/deploy code, so failures with it isn't really an attack vector and it works in Parity/Geth/TestRPC so it functions "well enough" even if it does have bugs.

PR #439, opened 2017-11-08
use a script to generate the interfaces file instead of hand editing

nuevoalex:
Tests are passing and the interfaces look sane to me generally. I was able to get a relatively clean way of doing `Controlled` inheritance too.


MicahZoltu:
Per discussion is #core-dev, recommend putting the folder structure back with `libraries` containing any TS files that are used in multiple parts of the project (e.g., scripts, tests, tools) and having `scripts` contain anything meant to be run from CLI and the new `tools` folder can be things used as part of our build toolchain (like this new stuff).

source/tools/generateContractInterfaces.ts
```diff
@@ -0,0 +1,188 @@
...
+}).catch(error => {
+    console.log(error);
+    process.exit();
+});
+
+async function doWork(): Promise<void> {
+
+    const configuration = await Configuration.create();
+
+    const contractABIs: CompilerOutput = JSON.parse(await fs.readFile(configuration.contractOutputPath, "utf8"));
```
MicahZoltu: Recommend having `generateContractInterfaces.ts` call `ContractCompiler.ts` and then pass the return value from that into `contractInterfacesTemplate`.  `ContractCompiler` will correctly do caching and compile if necessary (if cache is busted).  This avoids having to read the file off disk after writing it to disk.  It also avoids running into a problem where `compile` wasn't run first and re-generating the TS files using stale compilation output.

source/tools/generateContractInterfaces.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    const fileContents: String = contractInterfacesTemplate(contractABIs.contracts);
+
+    await fs.writeFile(configuration.contractInterfacesOutputPath, fileContents);
+}
+
+function contractInterfacesTemplate(contracts: CompilerOutputContracts) {
```
MicahZoltu: Style: The pattern I have been following is to have scripts be as short/simple as possible and putting all significant business logic into a separate file that can be called either from a script or from something else (such as a test or another library).  I also personally prefer class files rather than top level functions, though I don't have a strong argument for this if you have something particularly simple.

In this case, what _I_ would do is create a library class named `ContractInterfacesGenerator` that takes in a `Configuration` and a `ContractCompiler` as constructor parameters and then a single function on it called `generateContractInterfaces`.  This would allow us to provide mocks for `ContractCompiler` and `Configuration` should we want to author tests for this script at some point in the future, and it would follow the pattern the rest of our TS stuff follows (dependency injection with classes).  `doWork` above would then just instantiate `Configuration`, instantiate `ContractCompiler`, instantiate `ContractInterfacesGenerator` and then call `generateContractInterfaces`.

I would probably also have `generateContractInterfaces` return the generated string, and also write the file to disk (in case a future test author wants to validate the string).

We don't have an agreed upon style guide/pattern set for TypeScript, so all of this is up for discussion.  The above is just the pattern I have been following elsewhere, and in lieu of any discussion consistency is the current reigning champion that drives my feedback here.

nuevoalex: That makes sense. I like putting what we can into library files anyway so I'll do this.

nuevoalex:
Changes pushed

source/libraries/AccountManager.ts
```diff
@@ -3,7 +3,7 @@ import { Account, generate, privateToAccount } from 'ethjs-account';
 import { Transaction } from 'ethjs-shared';
 import { sign } from 'ethjs-signer';
 import { Connector } from './Connector';
-import { Configuration } from './Configuration';
+import { Configuration } from '../libraries/Configuration';
```
MicahZoltu: This line shouldn't have changed.  Both are correct, but the indirection doesn't add value IMO.

source/libraries/Connector.ts
```diff
@@ -1,7 +1,7 @@
 import EthjsHttpProvider = require('ethjs-provider-http');
 import EthjsQuery = require('ethjs-query');
 import { TransactionReceipt } from 'ethjs-shared';
-import { Configuration } from './Configuration';
+import { Configuration } from '../libraries/Configuration';
```
MicahZoltu: Same comment as above, and I see it in several other places below (this will be the last comment on the topic).

source/libraries/ContractInterfacesGenerator.ts
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    ${contractInterfaces.join("\n")}
+    export function ContractFactory(connector: Connector, accountManager: AccountManager, address: string, defaultGasPrice: BN): Controlled {
+        return new Controlled(connector, accountManager, address, defaultGasPrice);
+    }
+    `;
+}
+
+    function contractInterfaceTemplate(contractName: String, contractAbi: Abi) {
```
MicahZoltu: Indentation is off from here down I believe.  This should be at the root (unless you wrap the whole thing in a class).

source/libraries/ContractInterfacesGenerator.ts
```diff
@@ -0,0 +1,181 @@
+import * as fs from "async-file";
+import { Abi, AbiFunction, Primitive } from 'ethereum';
+import { ContractCompiler } from "../libraries/ContractCompiler";
+import { CompilerOutput, CompilerOutputContracts } from "solc";
+import { Configuration } from '../libraries/Configuration';
+
+
+export async function generateContractInterfaces(compiler: ContractCompiler, configuration: Configuration): Promise<String> {
```
MicahZoltu: I would rather see this as a class (for consistency) as mentioned in the other comment.  I think I just like importing a first letter capitalized things....
```
public contract ContractInterfaceGenerator {
  constructor(configuration: Configuration, contractCompiler: ContractCompiler) {
    // TODO: assign to member variables
  }
  public function generateContractInterfaces(): Promise<string> {
    // TODO: compile contracts
    // TODO: generate string
    // TODO: write to file
    // TODO: return string
  }
  private function contractInterfacesTemplate(): string {
    // TODO
  }
}
```

nuevoalex: Makes sense! Will do

nuevoalex:
All classed up now

source/libraries/ContractInterfacesGenerator.ts
```diff
@@ -0,0 +1,191 @@
...
+
+    ${contractInterfaces.join("\n")}
+    export function ContractFactory(connector: Connector, accountManager: AccountManager, address: string, defaultGasPrice: BN): Controlled {
+        return new Controlled(connector, accountManager, address, defaultGasPrice);
+    }
+    `;
+    }
+
+    private contractInterfaceTemplate(contractName: String, contractAbi: Abi) {
+        let contractMethods: Array<String> = [];
```
MicahZoltu: Recommend `const` over `let` almost everywhere unless you actually need to re-assign to the variable.  `const` can only be assigned once at declaration time, which is what you are doing in almost all of these cases.

The one case I commonly use `let` over `const` is for the iterated variable in a `for` loop.

nuevoalex: Yea I tried to do that in general. I think I wrongly thought consts were immutable originally. Will fox these up.

source/libraries/ContractInterfacesGenerator.ts
```diff
@@ -0,0 +1,191 @@
...
+    private readonly configuration: Configuration;
+
+    public constructor(configuration: Configuration, compiler: ContractCompiler) {
+        this.compiler = compiler;
+        this.configuration = configuration;
+    }
+
+    public async generateContractInterfaces(): Promise<String> {
+
+        const contractsOutput: CompilerOutput = await this.compiler.compileContracts();
```
MicahZoltu: As part of some upcoming work I'll be doing I am going to extract out the logic that filters down to just contracts that are interesting (meaning none of the libraries or interface contracts).  Once that is done we should utilize it here so as well so we don't generate these for various things that are never instantiated like interfaces and libraries.  Just a thought to keep in mind and to comment on if you think that is a bad idea.  Nothing needs to be done in this PR though.

source/libraries/ContractInterfacesGenerator.ts
MicahZoltu: I wish.  :'(

source/libraries/ContractInterfacesGenerator.ts
```diff
@@ -0,0 +1,191 @@
...
+    }`;
+    }
+
+    private getTsTypeFromPrimitive(abiType: Primitive, forReturnValue?: boolean) {
+        switch(abiType) {
+            case 'uint8':
+            case 'uint64':
+            case 'uint256':
+            case 'int256': {
+                return forReturnValue ? 'BN' : 'BN|number';
```
MicahZoltu: I did this (BN|number) to make the interfaces easier to use.  However, it does allow the user to pass in a floating point number which causes problems.  We _could_ require the user always pass in a BN instead, just makes some call sites ugly when all you want to do is pass in `1` or something.

source/tests-integration/TestFixture.ts
```diff
@@ -38,13 +38,13 @@ export class TestFixture {
...
         const authority = this.contractDeployer.getContract('Augur');
-        const cash = <Cash>this.contractDeployer.getContract('Cash');
+        const cash = new Cash(this.connector, this.accountManager, this.contractDeployer.getContract('Cash').address, this.configuration.gasPrice);
         const transactionHash = await cash.approve(authority.address, new BN(2).pow(new BN(256)).sub(new BN(1)));
         await this.connector.waitForTransactionReceipt(transactionHash, `Approving central authority.`);
     }
 
     public async createMarket(universe: Universe, numOutcomes: number, endTime: number, feePerEthInWei: number, denominationToken: string, designatedReporter: string, numTicks: number): Promise<Market> {
-        const legacyReputationToken = <LegacyReputationToken>this.contractDeployer.getContract('LegacyReputationToken');
+        const legacyReputationToken = new LegacyReputationToken(this.connector, this.accountManager, this.contractDeployer.getContract('LegacyReputationToken').address, this.configuration.gasPrice);
```
MicahZoltu: I was thinking that we should make a `getContract` overload that is generic and somehow returns the right class:
```
const cash = factory<Cash>('Cash');
```
The `factory` would curry all of the other bits like connector, accountManager, contractDeployer, etc. and it would get the contract from `contractDeployer.getContract` and then instantiate the contract via a mapping of `string => constructorFunction` that is auto-generated by the code generator.

This is out of scope of this PR, but as we start authoring more and more tools/tests against the TypeScript I believe such a feature would be very useful.

source/tools/generateContractInterfaces.ts
```diff
@@ -0,0 +1,23 @@
...
+
+async function doWork(): Promise<void> {
+
+    const configuration = await Configuration.create();
+
+    const compiler: ContractCompiler = new ContractCompiler(configuration);
+
+    const interfacesGenerator: ContractInterfaceGenerator = new ContractInterfaceGenerator(configuration, compiler);
+
+    interfacesGenerator.generateContractInterfaces();
```
MicahZoltu: Style, for such a short function, it feels like the spaces between each line are unnecessary.  Also, the space on first line of the function doesn't align with our style elsewhere.

source/libraries/ContractInterfacesGenerator.ts
nuevoalex: That would simplify things and make this more correct so I'm in favor if you think that's alright.

source/tools/generateContractInterfaces.ts
nuevoalex: Will fix

source/libraries/ContractInterfacesGenerator.ts
nuevoalex: To clarify I mean in favor of making this just `BN`

PR #442, opened 2017-11-08
change false to true.

bthaile:
market method returns false, needs to be true


bthaile:
squashing bugs left and right!

PR #443, opened 2017-11-09
Moves filtering to ContractCompiler.

MicahZoltu:
This will greatly reduce the on-disk file, which makes it more usable for the front-end and also speeds things up marginally (no more need to read/parse a 12.5MB file JSON file).

@pgebheim `output/contracts/contracts.json` is ready to be shipped off to `augur-contracts`.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -10,6 +10,13 @@
             "problemMatcher": [
                 "$tsc-watch"
             ]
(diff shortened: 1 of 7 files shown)


MicahZoltu:
Can I get a re-read of this PR.  I ended up also implementing the address stuff (cc @pgebheim).

PR #444, opened 2017-11-09
Linter: Enable (pre-installed) security plugin

duaraghav8:
Hi guys,
Solium's [official security plugin](https://github.com/duaraghav8/solium-plugin-security) checks for security based on [recommendations from Consensys](https://consensys.github.io/smart-contract-best-practices/recommendations/).

The plugin simply needs to be enabled, no installation or other steps required.

Configuring in `.soliumrc.json` is simple:
```
{
    "plugins": ["security"],
    "rules": {
        "security/no-throw": 0
    }
}
```

[Usage Guide & List of Rules](https://github.com/duaraghav8/solium-plugin-security) (2 minute read)
[Release blog](https://medium.com/solium/soliums-official-security-plugin-4d604e708c3c)

Although not mandatory, its highly recommended that you enable the plugin. So its your decision:)

(This PR will probably fail your CI tests because all security lint rules are treated as errors by default).

### DIFF
--- a/.soliumrc.json
+++ b/.soliumrc.json
@@ -1,5 +1,6 @@
 {
   "extends": "solium:all",
+  "plugins": ["security"],
   "rules": {
     "imports-on-top": 0,
     "variable-declarations": 0,
--- a/.soliumrc.json
+++ b/.soliumrc.json
@@ -9,6 +9,11 @@
     "no-empty-blocks": 0,
     "arg-overflow": 0,
     "uppercase": 0,
-    "operator-whitespace": 0
+    "operator-whitespace": 0,
+    "security/no-low-level-calls": 0,
+    "security/no-inline-assembly": 0,
+    "security/no-call-value": 0,
+    "security/no-block-members": 0,
+    "security/enforce-explicit-visibility": 0
   }
 }


MicahZoltu:
Thanks for the PR @duaraghav8.

Our policy for linters historically has been to have all options explicitly listed (even if they are set to defaults) rather than using implicit default rules.  This makes turning them on/off easier without having to go research the syntax or rule names.

I believe the following will all fail for our repository because we intentionally disobey the rules in a couple places:
```
no-low-level-calls
no-inline-assembly
no-call-value
no-block-members
```

MicahZoltu:
Separately, (cc @nuevoalex), we should disable solium linting of `legacy_reputation_token` since that is already deployed and we have no intention of re-deploying it, so the code should remain untouched.  We may want to consider also disabling linting of the test contracts, though we could also just fix them up to be good citizens.

duaraghav8:
@MicahZoltu I believe I've made the changes according to your policy. I've additionally disabled `enforce-explicit-visibility` which you can enable should you choose to lint for that too.
Let me know if you need any other changes!

PR #445, opened 2017-11-09
Orders bug

nuevoalex:
To celebrate my failure I added a test confirming this isn't a thing. Also some Solium cleanup now that fixes are in.

### DIFF
--- a/source/contracts/reporting/Universe.sol
+++ b/source/contracts/reporting/Universe.sol
@@ -324,10 +324,20 @@ contract Universe is DelegationTarget, Extractable, ITyped, Initializable, IUniv
         // Modify the amount based on the previous amount and the number of markets fitting the failure criteria. We want the amount to be somewhere in the range of 0.5 to 2 times its previous value where ALL markets with the condition results in 2x and 0 results in 0.5x.
         if (_badMarkets <= _totalMarkets / _targetDivisor) {
             // FXP formula: previous_amount * actual_percent / (2 * target_percent) + 0.5;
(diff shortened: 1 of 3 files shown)


source/contracts/reporting/Universe.sol
```diff
@@ -324,10 +324,20 @@ contract Universe is DelegationTarget, Extractable, ITyped, Initializable, IUniv
         // Modify the amount based on the previous amount and the number of markets fitting the failure criteria. We want the amount to be somewhere in the range of 0.5 to 2 times its previous value where ALL markets with the condition results in 2x and 0 results in 0.5x.
         if (_badMarkets <= _totalMarkets / _targetDivisor) {
             // FXP formula: previous_amount * actual_percent / (2 * target_percent) + 0.5;
-            _newValue = _badMarkets.mul(_previousValue).mul(_targetDivisor).div(_totalMarkets).div(2) + _previousValue / 2; // FIXME: This is on one line due to solium bugs
+            _newValue = _badMarkets
+                .mul(_previousValue)
+                .mul(_targetDivisor)
+                .div(_totalMarkets)
+                .div(2) + _previousValue / 2;
```
MicahZoltu: Why doesn't the tail end use safe math?
```
.div(2)
.add(_previousValue
    .div(2))
```

MicahZoltu: Same question below.

nuevoalex: Stack depth! :1st_place_medal:

PR #446, opened 2017-11-09
Solium security plugin changes

nuevoalex:


.soliumrc.json
```diff
@@ -13,7 +13,6 @@
     "security/no-low-level-calls": 0,
     "security/no-inline-assembly": 0,
     "security/no-call-value": 0,
-    "security/no-block-members": 0,
-    "security/enforce-explicit-visibility": 0
+    "security/no-block-members": 0
```
MicahZoltu: Style: We should leave in settings that are default so it is explicit and doesn't change in the future if defaults change. Some goes for the other three missing settings, which I think should be included with an explicit sitting.

This allows the engineer to easily here litter settings without needing to research documentation on how to do so, and also makes it easy to reference what settings we are using.

.soliumrc.json
```diff
@@ -4,16 +4,28 @@
...
+    "no-unused-vars": "error",
+    "blank-lines": "warning",
+    "whitespace": "warning",
+    "pragma-on-top": "warning",
     "security/no-low-level-calls": 0,
     "security/no-inline-assembly": 0,
     "security/no-call-value": 0,
     "security/no-block-members": 0,
-    "security/enforce-explicit-visibility": 0
+    "security/enforce-explicit-visibility": "error"
```
MicahZoltu: I probably should have been more clear (though I do think what you added is good):
These are the new ones I was referring to that I think we should add:
```
no-throw
no-sha3
no-tx-origin
```

Source: https://github.com/duaraghav8/solium-plugin-security#list-of-rules

MicahZoltu: Right now they all default to enabled (I think?) but we should include them as explicitly enabled.

MicahZoltu: Also, can we follow the `error`/`warning` syntax for everything instead of some being `0` and `1`?

nuevoalex: +1 on turning `0` to `"off"` but what do you mean by explicitly enabling?

nuevoalex: To clarify, there is no explicit enable other than specifying an error level.

nuevoalex: Reading again I think you're just re-iterating that we should be including these rules in our config. Gonna assume that's the case

MicahZoltu: At the moment, `security/no-throw`, `security/no-sha3` and `security/no-tx-origin` don't show up anywhere in this configuration file, yet they are rules the linter has enabled because they are on-by-default.  I would like to add a line for each of those rules so they show up in our config as enabled as errors.  This is the same as you did for a bunch of other rules here, but for the security plugin.

PR #450, opened 2017-11-11
Removing unused array libraries, adding some comments, & minor cleanup

adrake33:


tests/libraries/test_arrays.py
```diff
@@ -1,55 +0,0 @@
-#!/usr/bin/env python
```
nuevoalex: Why do we delete these tests? They're testing the uint256 array library which is still being used I think

nuevoalex:
Look good but I think we should keep the array library test and test helper

tests/libraries/test_arrays.py
adrake33: What's using the uint256 array library?  The only place I saw it being used was ArrayHelper, which was only being used by test_arrays.py.

nuevoalex: not used I guess, just not deleted. As long as we're keeping it I'd prefer it stay tested.

nuevoalex:
Approving. I think either delete the uint256 array library too though or keep the tests. Either is fine with me

adrake33:
Ah, sorry about that.  I meant to delete the uint256 array library as well, but it looks like I missed it when I was telling Git to remove the other array libraries.  I'll remove it.

PR #452, opened 2017-11-13
Updating functions for getting min/max uint256 values

adrake33:
SafeMathUint256.sol and SafeMathInt256.sol were using different function names for getting their min/max values (i.e., minUint256/maxUint256 vs getInt256Min/getInt256Max) , so I changed them to be in sync and added test functions for these function in SafeMathUint256, since it did not have any.

I also noticed sandbox.py seems to have uint256 and int256 defined incorrectly: 
```
uint256 = """
library int256Math {
...
}"""
int256 = """
library uint256Math {
...
}"""
```

I have not fixed that as part of this PR, but if that is something that should be changed, let me know, and I will add it.


MicahZoltu:
No need to fix sandbox.  That is just a place to quickly iterate on code when you want to test Solidity/EVM behavior.

MicahZoltu:
Any reason to not just some new both `getMax()`?  Having the type in the gutter name seems redundant since you have to call it as a static method on the type.

adrake33:
I suppose I could call them `getMin`/`getMax` if you think that would be better, but there are already functions called `min`/`max` which compare 2 numbers and return the min/max value of those.  I just figured this might be a little less confusing.

MicahZoltu:
Ah, good point.

PR #453, opened 2017-11-13
change function names and add view getter functions to support local calls more cleanly.

nuevoalex:


source/contracts/reporting/Universe.sol
```diff
@@ -267,27 +292,37 @@ contract Universe is DelegationTarget, Extractable, ITyped, Initializable, IUniv
     }
 
     function getValidityBond() public onlyInGoodTimes returns (uint256) {
```
adrake33: Perhaps this should be called something like `setAndGetValidityBond` and `getValidityBondView` should be called `getValidityBond`?

nuevoalex: Yea that's a good idea. Will do

nuevoalex: Same with the target gas getter

PR #456, opened 2017-11-14
more tests and actually place requires in the internal methods to prevent bad values when windows don't exist

nuevoalex:


source/contracts/reporting/Universe.sol
```diff
@@ -352,6 +354,8 @@ contract Universe is DelegationTarget, Extractable, ITyped, Initializable, IUniv
     }
 
     function getDesignatedReportNoShowBondInternal(IReportingWindow _reportingWindow, IReportingWindow _previousReportingWindow) internal view onlyInGoodTimes returns (uint256) {
```
adrake33: I realize this would be kind of annoying to change at this point, but I'm wondering if we should revise the No-Show Bond functions since the term No-Show Bond" has been deprecated in favor of "First Report REP Bond".  The docs no longer use this term, so it will seem a bit odd to the API users that it's called a No-Show Bond.

nuevoalex: I'll bring that up as something to do in another PR possibly but go ahead and merge this once I fix that test

adrake33: Sounds good!

PR #458, opened 2017-11-14
Make the bond and fee calculation functions return 0 if no value can be determined in the view context

nuevoalex:


MicahZoltu:
I'm pretty strongly against changing a hard failure into "return 0".  This is moving from a "fail hard, fail fast" pattern to a "fail quiet, fail slow" pattern which tends to lead to more bugs and can potentially open up security vulnerabilities, depending on how the return values are used.  In general, I think we should be throwing anywhere we encounter an unexpected state, and we should never be returning a sentinel "failure" value.

@tinybike Since this is a public facing only change (at the moment, though future iterations may result in it having contract internal impact in the future), I suspect the request likely came from you?  Care to weigh in on why you believe this change is important enough to increase contract risk?

joeykrug:
Make it throw IMO

tinybike:
I requested a change to `return 0` for `eth_call` requests.  The reason for this is so the caller can distinguish "this thing needs to be created/cached" from "your request failed for some other reason", so that the caller knows whether to send a follow-up transaction that actually creates/caches.

I believe my suggestion at the time was to split the public-facing endpoints into getter-only and setter-only functions.  This would eliminate the problem described above, and have the ancillary benefit of removing any ambiguity about each function's purpose.  I'm still in favor of implementing this, unless it's not feasible for some reason!

nuevoalex:
It is split into getter-only and set-or-get methods. If there is a problem in the set-or-get variant with creating the windows it will presumably throw at that point.

tinybike:
@nuevoalex Is it necessary to have set-or-get methods?  Intuitively (naively?) it seems cleaner to me to have a setter-only method and a getter-only method.  I know the set-or-get method is useful internally; could this code be refactored so that its external API has 2 methods, `setThing` and `getThing` (for use with `eth_sendTransaction` and `eth_call`, respectively), while the internal method remains `getOrSetThing`?

MicahZoltu:
It needs to be a get or set for the internal calls to avoid having a conditional at every call site and reducing the number of cross-contract calls.

```
address result;
if (contract.hasThing()) {
    result = contract.getThing();
} else {
    contract.setThing();
    result = contract.getThing();
}
```
vs
```
address result = contract.getOrSetThing();
```

nuevoalex:
So in the case of `getOrCacheValidityBondInternal` for example you're suggesting we create a function that wraps that function call and returns bool? Note these aren't setting things based on user input, the function is using existing state (or creating when needed and appropriate to the call) and calculating a value based on that which it caches. If this is correct how does that wrapper function help or address the above desires to have the getter throw?

tinybike:
Right, but the external and internal APIs don't have to be the same.  There can be a `getOrSetThing` internal function that does the `.hasThing()` check, so it doesn't have to be done at the callsite:
```
function getOrSetThing() internal returns (address) {
    if (!contract.hasThing()) {
        contract.setThing();
    }
    address result = contract.getThing();
    return result;
}
```
So the internal callsite is just:
```
address result = contract.getOrSetThing();
```
whereas the `getThing` and `setThing` functions are available externally.

Am I missing something?  This seems pretty straightforward to me.

MicahZoltu:
The problem with separating the three is that `setThing()` should throw if the thing is already set, but this introduces a race condition if you are trying to mimick `getOrSet` off-chain since between the local call to `hasThing` and the time your transaction for `setThing` is called someone else may call `setThing` on-chain.  This results in the user paying gas to have a failed transaction, which is a sucky user experience.

@tinybike Can you describe the use-case you have for off-chain being able to `getOrSet`?  The fact that you need to directly use `getOrSet` suggests maybe we have some other problem and the better solution is to not have the user ever have to call `getOrSet`.  For example, in market creation we need the reporting window to exist first, so we should probably have a method on `Universe` for `createMarket` that will create the reporting window first if necessary and then create the market after that.

nuevoalex:
Going to ask that we please use these specific functions for our examples. These are not at all analogous to things like just getting and setting a reporting window and if we arrive at some conclusion based on discussing that it will not be clear what we want to implement for these. They require multiple reporting windows and this needs to be callable before the market is created for example.

tinybike:
@MicahZoltu I don't want to do `getOrSet` off-chain.  The only reason I'm calling `getOrSet` functions off-chain is that there aren't pure getter and setter functions available.

PR #459, opened 2017-11-15
Add topic to the reporting window and market created log. Also fix some TS bugs that existed before

nuevoalex:


MicahZoltu:
What is the point of this change?  We aren't using it anywhere in the contracts, feels like it should remain part of `extraInfo`.  The primary reason we created the Augur Node was so that it could do indexing for us and we could stop worrying about trying to figure out how to do on-chain indexing of things (including event logs) since that ended up being a path full of failure.

cc @tinybike

PR #462, opened 2017-11-15
Send reporterFees and marketCreatorFees seperately on orderFilled event

nuevoalex:


source/contracts/trading/FillOrder.sol
```diff
@@ -392,11 +392,15 @@ contract FillOrder is CashAutoConverter, Extractable, ReentrancyGuard, IFillOrde
...
+        logOrderFilled(_tradeData, _marketCreatorFees, _reporterFees, _tradeGroupId);
         _tradeData.contracts.orders.fillOrder(_orderId, _tradeData.getMakerSharesDepleted(), _tradeData.getMakerTokensDepleted());
         return _tradeData.filler.sharesToSell.add(_tradeData.filler.sharesToBuy);
     }
 
+    function logOrderFilled(Trade.Data _tradeData, uint256 _marketCreatorFees, uint256 _reporterFees, uint256 _tradeGroupId) private returns (bool) {
+        controller.getAugur().logOrderFilled(_tradeData.contracts.market.getUniverse(), _tradeData.contracts.market.getShareToken(_tradeData.order.outcome), _tradeData.filler.participantAddress, _tradeData.order.orderId, _tradeData.getMakerSharesDepleted(), _tradeData.getMakerTokensDepleted(), _tradeData.getFillerSharesDepleted(), _tradeData.getFillerTokensDepleted(), _marketCreatorFees, _reporterFees, _tradeGroupId);
+    }
```
adrake33: This function should have a return statement.

source/contracts/trading/FillOrder.sol
```diff
@@ -97,7 +97,7 @@ library Trade {
         }
 
         // sell complete sets
-        uint256 _settlementFees = _data.contracts.completeSets.sellCompleteSets(this, _data.contracts.market, _numberOfCompleteSets);
+        var (_marketCreatorFees, _reporterFees) = _data.contracts.completeSets.sellCompleteSets(this, _data.contracts.market, _numberOfCompleteSets);
```
MicahZoltu: I believe you can do `uint256 (_marketCreatorFees, _reporterFees) = ...` here.  `var` is only necessary (I think) if the types differ and you don't want to declare on a separate line.

PR #470, opened 2017-11-19
Add specific example case to unit tests and add creating an order to the integration tests

nuevoalex:


nuevoalex:
Closing as this branch will be used to address https://app.clubhouse.io/augur/story/4849/trading-with-escrowed-shares-is-not-tested-and-has-a-poor-ux-we-should-tests-it-and-make-explicit-user-approval-not-needed

PR #472, opened 2017-11-21
Market creation changes

nuevoalex:
A great deal of this is just modifying tests.

Whether we actually want to ultimately merge this or not I guess is being discussed currently.

### DIFF
--- a/source/contracts/reporting/Reporting.sol
+++ b/source/contracts/reporting/Reporting.sol
@@ -89,7 +89,7 @@ library Reporting {
 
     function getScalarMarketNumTicks(int256 _minPrice, int256 _maxPrice, int256 _tickShift) internal pure returns (uint256) {
         // Overflow prevention
-        require (_tickShift < 76);
+        require (_tickShift <= 76);
         uint256 _normalizedTicks = 10 ** _tickShift.abs();
         uint256 _range = uint256(_maxPrice.sub(_minPrice));
         if (_tickShift < 0) {
--- a/source/contracts/reporting/Reporting.sol
+++ b/source/contracts/reporting/Reporting.sol
@@ -87,7 +87,12 @@ library Reporting {
         if (_numOutcomes == 8) {return 10000;}
     }
 
-    function getScalarMarketNumTicks(int256 _minPrice, int256 _maxPrice) internal pure returns (uint256) {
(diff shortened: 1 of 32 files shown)


source/contracts/reporting/Reporting.sol
```diff
@@ -65,4 +73,26 @@ library Reporting {
...
+        if (_numOutcomes == 2) {return 10000;}
+        if (_numOutcomes == 3) {return 10002;}
+        if (_numOutcomes == 4) {return 10000;}
+        if (_numOutcomes == 5) {return 10000;}
+        if (_numOutcomes == 6) {return 10002;}
+        if (_numOutcomes == 7) {return 10003;}
+        if (_numOutcomes == 8) {return 10000;}
+    }
+
+    function getScalarMarketNumTicks(int256 _minPrice, int256 _maxPrice, int256 _tickShift) internal pure returns (uint256) {
```
tinybike: Hm, `_normalizedTicks` will overflow / wrap around if `_tickShift < -255*log10(2)` or `_tickShift > log10(2^255 - 1)` (i.e. minus or plus about 76.8, respectively).  Suggest adding a `require` here to make sure `76 <= _tickShift <= 76`.

Are there _practical_ limits to `_tickShift` that we should enforce on-contract?  A market with `_tickShift = 76` will be utterly unusable; should we set the bounds at some lower value?  (18 is the obvious choice, although possibly some users will have a reason to want to set `_tickShift` higher than 18...?)

nuevoalex: I make a note of this in the tests I added that if a ridiculous value was arrived at here (through overflow or otherwise) the market creation process will fail as it normally does for invalid `numTicks`. Adding sanitation here might still be a good idea but it will just be a convenience for callers that accidentally enter in a silly value that through overflow that ends up being a legitimate accepted `numTicks` later. Still, this makes sense and I like `requires` so lets add that.

Any unusable tickShift will end up with a 0 `numTicks` value which will be ultimately rejected. An obscene level of precision is allowed here but we didn't establish a limit on that and it's very subjective.

nuevoalex:
Changes made!

nuevoalex:
TODO Addition: allow the user to pass in an arbitrary ticksize (instead of just an exponent) to scalar markets.  (Categorical and binary markets should have their ticksizes automatically set)

PR #473, opened 2017-11-21
Make ShareTokens use trustedTransfer

nuevoalex:
I also made a private method on market that may need to be manually called public

### DIFF
--- a/source/contracts/trading/ShareToken.sol
+++ b/source/contracts/trading/ShareToken.sol
@@ -36,7 +36,11 @@ contract ShareToken is DelegationTarget, Extractable, ITyped, Initializable, Var
         return true;
     }
 
-    function trustedTransfer(address _source, address _destination, uint256 _attotokens) public onlyWhitelistedCallers onlyInGoodTimes afterInitialized returns (bool) {
+    function trustedOrderTransfer(address _source, address _destination, uint256 _attotokens) public onlyCaller(bytes32("CreateOrder")) onlyInGoodTimes afterInitialized returns (bool) {
+        return internalTransfer(_source, _destination, _attotokens);
+    }
+
+    function trustedFillOrderTransfer(address _source, address _destination, uint256 _attotokens) public onlyCaller(bytes32("FillOrder")) onlyInGoodTimes afterInitialized returns (bool) {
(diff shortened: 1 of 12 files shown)


source/contracts/trading/ShareToken.sol
```diff
@@ -36,6 +36,10 @@ contract ShareToken is DelegationTarget, Extractable, ITyped, Initializable, Var
         return true;
     }
 
+    function trustedTransfer(address _source, address _destination, uint256 _attotokens) public onlyWhitelistedCallers onlyInGoodTimes afterInitialized returns (bool) {
```
MicahZoltu: Bleh.  I'm generally not a fan of the whole "whitelisted caller" thing because it makes auditing _really_ hard as you have to assume that _any_ contract in the whitelist can call it.  Also, it means that the Augur dev team has the ability to transfer user assets without constraint.  It is possible that whitelisted callers already has the ability to steal all user funds, in which case this doesn't matter much but if not then this is a pretty major break in the security model (the one currently going through audits).

I would be more OK with this if we constrained it to specific contracts that are allowed to transfer funds (using something like `controller.assertOnlySpecifiedCaller`).  In general, anytime I see `onlyWhitelistedCallers` I become very apprehensive for the above reasons.

nuevoalex:
The Order and FillOrder contract need to call trustedTransfer. I'm ok with implementing a modifier that restricts the caller by string as suggested

source/contracts/trading/ShareToken.sol
```diff
@@ -36,6 +36,14 @@ contract ShareToken is DelegationTarget, Extractable, ITyped, Initializable, Var
         return true;
     }
 
+    function trustedOrderTransfer(address _source, address _destination, uint256 _attotokens) public onlyCaller(bytes32("CreateOrder")) onlyInGoodTimes afterInitialized returns (bool) {
```
MicahZoltu: The explicit `bytes32` conversion should not be necessary here.  Solidity will implicitly cast a string literal into a `bytes32` if the context requires it (in this case it does since the `onlyCaller` signature takes a `bytes32`.  What you can't do is assign a string literal to a `string` and then assign that `string` to a `bytes32`, but that caveat shouldn't apply here.

PR #474, opened 2017-11-21
Speed up testing and deployment in CI and locally

pgebheim:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "augur-core",
-  "verson": "0.1.0",
+  "version": "0.0.1",
```
nuevoalex: why the downgrade?

tinybike: Hm.  IMO we probably should only _bump_ the version, not downgrade it.

pgebheim: I didn't downgrade it, the old fake version was spelled wrong ;-)

So I am just starting at 0.0.1 -- and then once we get things rolling we can pull up a real Version.

PR #475, opened 2017-11-22
Fix reversal of min order value check

nuevoalex:


joeykrug:
What's the deal with this?

Also why were tests not broken before / after changing this?

nuevoalex:
This was a bug. There were not tests explicitly checking that orders must be greater than the minimum order amount.

PR #476, opened 2017-11-22
Attempt Travis Build Matrixing

pgebheim:


pgebheim:
Tons of little commits on this, but after all the tweeks.

Currently, the longest test of python tests are in the test/reporting folder, they account for nearly 10 minutes of test time.

By running that in parallel with the integration tests and other python tests, we cut the build time down from ~20 minutes to ~13 minutes.

If the reporting tests were split into an easily identifiable subset (e.g. two directories that make sense), we could cut this further and do a full build with full python tests in perhaps less than 10 minutes.

This version relies heavily on docker image caching, which we use to store the build between layers as well as cache between builds. This can reduce the time it takes to build the initial image from 6 minutes to as low as 2.

.travis.yml
```diff
@@ -12,25 +12,36 @@ before_install:
...
-  - provider: script
-    skip_cleanup: true
-    script: ARTIFACTS=true npm run docker:deploy
-    on:
-      tags: true
+script: if [[ "$TESTS" == "integration" ]]; then
+          npm run docker:test:integration;
+        else
+          npm run docker:test:unit:single -- $TESTS;
+        fi
```
MicahZoltu: Is it possible to move this logic into the NPM script?  It would be nice to have conditional logic not in bash inside the travis file if possible and instead somewhere else.

.travis.yml
```diff
@@ -12,25 +12,36 @@ before_install:
...
+jobs:
+  include:
+    - stage: build
+      script: docker login -u "$DOCKER_USERNAME" -p "$DOCKER_PASSWORD";
+              docker pull augurproject/augur-core:latest;
+              npm run docker:build;
+              npm run docker:push
```
MicahZoltu: Is this really valid YML?  Is semicolon in YML an escape character?

.travis.yml
```diff
@@ -12,25 +12,36 @@ before_install:
...
+      env:
+        - DOCKER_USERNAME=pgebheim
```
MicahZoltu: At some point we should get a team account rather than using your personal account.  Not a priority, but would be good to get rid of the bus factor.  We can use KeyBase for storing raw credentials.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "augur-core",
-  "version": "0.10.0",
+  "version": "0.10.9",
```
MicahZoltu: Recommend just going to 1.0.  I have never personally been a fan of starting from 0 and having minor/patch version slots mean something different from normal sem ver.

That being said, did we decide what our versioning scheme was going to be based off of?  What defines whether we bump major, minor or patch?

package.json
```diff
@@ -19,6 +19,7 @@
     "build:interfaces": "node output/tools/generateContractInterfaces.js",
     "build": "npm run build:source && npm run lint && npm run build:contracts && npm run build:interfaces",
     "test:unit": "pytest tests -vv",
+    "test:unit:single": "pytest -vv",
```
MicahZoltu: How is this different from `test:unit`?

support/artifacts/run.sh
```diff
@@ -40,9 +40,14 @@ fi
 git clone $repo_url output/augur-contracts
 current_dir=$PWD;
 cd output/augur-contracts
+
+git config user.email "team@augur.net"
+git config user.name "The Augur Developers"
```
MicahZoltu: Consider setting the name to something like `Augur Artifact Scripts` or `Augur CI` so it is clear that this isn't coming from a person (and probably hasn't been looked at by a human).

package.json
pgebheim: `test:unit` defaults to the "tests" directory, this allows us to pass in some tests. It's just there so you can iterate on a single unit test.

support/artifacts/run.sh
pgebheim: I like that!

.travis.yml
pgebheim: Indeed. I have integration accounts for everything else just hadn't done it -- I'll do it now before merge.

PR #477, opened 2017-11-22
4896 solc 0418

bthaile:
update solc version to 0.4.18. integration test and unit tests pass running in docker and pytest tests ran successfully.


source/libraries/ContractCompiler.ts
```diff
@@ -66,11 +66,6 @@ export class ContractCompiler {
                 optimizer: {
                     enabled: true,
                     runs: 500
-                },
-                outputSelection: {
-                    "*": {
-                        "*": [ "abi", "evm.bytecode.object" ]
-                    }
```
MicahZoltu: I finally got a chance to dig into this.  Can you leave this here and update the `conftest.py` block to match it, then see if everything compiles in 0.4.18?  It appears that this is actually correct for 0.4.18 and it was just a bug in `conftest.py` (we are missing the `"*": { ... }` layer in the middle).

nuevoalex:
LGTM

PR #479, opened 2017-11-27
3936 market more tests

bthaile:
more unit tests for market contract.

### DIFF
--- a/tests/reporting/test_market_unit.py
+++ b/tests/reporting/test_market_unit.py
@@ -66,45 +66,45 @@ def test_market_creation(localFixture, mockUniverse, mockReportingWindow, mockCa
     assert mockShareTokenFactory.getCreateShareToken(3) == market.getShareToken(3)
     assert mockShareTokenFactory.getCreateShareToken(4) == market.getShareToken(4)
 
-def test_market_designated_report(localFixture, constants, mockUniverse, chain, initializeMarket, mockStakeToken, mockStakeTokenFactory, mockReportingWindow, mockReputationToken):
+def test_market_designated_report(localFixture, constants, mockUniverse, chain, initializedMarket, mockStakeToken, mockStakeTokenFactory, mockReportingWindow, mockReputationToken):
     with raises(TransactionFailed, message="market is not in designated reporting state"):
-        initializeMarket.designatedReport()
+        initializedMarket.designatedReport()
 
-    chain.head_state.timestamp = initializeMarket.getDesignatedReportDueTimestamp() - 1
+    chain.head_state.timestamp = initializedMarket.getDesignatedReportDueTimestamp() - 1
     with raises(TransactionFailed, message="designated report method needs to be called from StakeToken"):
-        initializeMarket.designatedReport()
+        initializedMarket.designatedReport()
 
     with raises(TransactionFailed, message="StakeToken needs to be contained by market"):
-        mockStakeToken.callMarketDesignatedReport(initializeMarket.address)
+        mockStakeToken.callMarketDesignatedReport(initializedMarket.address)
     
     payoutDesignatedNumerators = [numTicks, 0, 0, 0, 0]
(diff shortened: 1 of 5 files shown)


tests/reporting/test_market_unit.py
```diff
@@ -451,9 +439,94 @@ def test_market_migrate_through_one_fork(localFixture, initializeMarket, constan
...
 
+def test_market_disavow_tokens(localFixture, initializeMarket, mockUniverse, mockReportingWindow, mockMarket, mockStakeTokenFactory):
+    stakePayNumeration = [0,0,0,0,numTicks]
+    with raises(TransactionFailed, message="market state has to be AWAITING_FORK_MIGRATION"):
+        initializeMarket.disavowTokens()
+
+    mockUniverse.setForkingMarket(mockMarket.address)
+    mockReportingWindow.setIsForkingMarketFinalized(False)
+    with raises(TransactionFailed, message="market needs stake tokens"):
+        initializeMarket.disavowTokens()
```
MicahZoltu: I think this is actually a contract bug.  If there are no stake tokens, disavowing should just be a noop.  cc @nuevoalex

tests/reporting/test_market_unit.py
```diff
@@ -451,9 +439,94 @@ def test_market_migrate_through_one_fork(localFixture, initializeMarket, constan
...
+    assert initializeMarket.isContainerForStakeToken(mockStakeToken.address) == False
+
+def test_market_first_reporter_comp_check(localFixture, initializeMarket, chain, constants, mockUniverse, mockStakeToken, mockReportingWindow, mockMarket, mockStakeTokenFactory, mockReputationToken):
+    stakePayNumeration = [0,0,0,0,numTicks]
+    existingBalance = chain.head_state.get_balance(tester.a1)
+    mockReputationToken.setBalanceOfValueFor(initializeMarket.address, 1000)
+    with raises(TransactionFailed, message="sender must be a stake token"):
+        initializeMarket.firstReporterCompensationCheck(tester.a1)
+    
+    mockStakeToken2 = set_mock_stake_token_value(localFixture, initializeMarket, stakePayNumeration, mockStakeTokenFactory, 10)
```
MicahZoltu: Any reason why this is named with a `2` suffix?  If not, recommend removing the suffix to avoid confusion in future readers (I stopped and looked around for `mockStakeToken1`).

tests/reporting/test_market_unit.py
```diff
@@ -451,9 +439,94 @@ def test_market_migrate_through_one_fork(localFixture, initializeMarket, constan
     assert initializeMarket.getLastReportersDisputeBond() == longToHexString(0)
     assert initializeMarket.getTentativeWinningPayoutDistributionHash() == initializeMarket.getDesignatedReportPayoutHash()
 
+def test_market_disavow_tokens(localFixture, initializeMarket, mockUniverse, mockReportingWindow, mockMarket, mockStakeTokenFactory):
```
MicahZoltu: `initializeMarket` feels like a bad name for this fixture.  Any reason we aren't just calling it `market`?  Do you perhaps mean `initializedMarket`?

tests/reporting/test_market_unit.py
bthaile: no reason specifically for the '2' only to differentiate it from existing mockStakeToken. I'll make it more clear.

tests/reporting/test_market_unit.py
bthaile: agreed, I'll change to `initializedMarket`

tests/reporting/test_market_unit.py
bthaile: code on contract is 
`require(stakeTokens.getCount() > 0);`

MicahZoltu: Yeah, that feels wrong.  It feels like it should be `if (stakeTokens.getCount == 0) return true;`

bthaile: I'll make that change and submit as a separate PR.

nuevoalex: I'm fine with that change

PR #480, opened 2017-11-27
removed require for stake tokens to exist on market when disavow tokens.

bthaile:
removed require on staketokens to exist on market for disavowing stake tokens.


source/contracts/reporting/Market.sol
```diff
@@ -351,7 +351,8 @@ contract Market is DelegationTarget, Extractable, ITyped, Initializable, Ownable
 
     function disavowTokens() public onlyInGoodTimes returns (bool) {
         require(getReportingState() == ReportingState.AWAITING_FORK_MIGRATION);
-        require(stakeTokens.getCount() > 0);
+        if (stakeTokens.getCount() == 0) return true;
```
MicahZoltu: Style, should have curlies and newline in here, despite my example in other PR.  :)

bthaile: I'll make that change, my linter has all sorts of red issues.

PR #482, opened 2017-11-27
Universe mods

nuevoalex:
Remove the getter functions for market creation costs and such in favor of the get or create variants.

Also added the minimum reporting fee divisor while I was in there.

### DIFF
--- a/source/contracts/reporting/Reporting.sol
+++ b/source/contracts/reporting/Reporting.sol
@@ -19,7 +19,7 @@ library Reporting {
     uint256 private constant BOND_PAYOUT_MULTIPLIER = 2; // Multiplier applied to the dispute bond payment to get what they should ideally be compensated with
     uint256 private constant DEFAULT_REPORTING_FEE_DIVISOR = 100; // 1% fees
     uint256 private constant MAXIMUM_REPORTING_FEE_DIVISOR = 10000; // Minimum .01% fees
-    uint256 private constant MINIMUM_REPORTING_FEE_DIVISOR = 3; // Maximum 33.3~% fees
+    uint256 private constant MINIMUM_REPORTING_FEE_DIVISOR = 3; // Maximum 33.3~% fees. Note than anything less than a value of 2 here will likely result in bugs such as divide by 0 cases.
 
     // CONSIDER: figure out approprate values for these
     uint256 private constant DESIGNATED_REPORTER_DISPUTE_BOND_AMOUNT = 11 * 10**20;
--- a/source/contracts/reporting/Reporting.sol
+++ b/source/contracts/reporting/Reporting.sol
@@ -19,6 +19,7 @@ library Reporting {
     uint256 private constant BOND_PAYOUT_MULTIPLIER = 2; // Multiplier applied to the dispute bond payment to get what they should ideally be compensated with
     uint256 private constant DEFAULT_REPORTING_FEE_DIVISOR = 100; // 1% fees
     uint256 private constant MAXIMUM_REPORTING_FEE_DIVISOR = 10000; // Minimum .01% fees
+    uint256 private constant MINIMUM_REPORTING_FEE_DIVISOR = 3; // Maximum 33.3~% fees
 
     // CONSIDER: figure out approprate values for these
     uint256 private constant DESIGNATED_REPORTER_DISPUTE_BOND_AMOUNT = 11 * 10**20;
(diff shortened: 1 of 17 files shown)


source/contracts/reporting/Reporting.sol
```diff
@@ -19,6 +19,7 @@ library Reporting {
     uint256 private constant BOND_PAYOUT_MULTIPLIER = 2; // Multiplier applied to the dispute bond payment to get what they should ideally be compensated with
     uint256 private constant DEFAULT_REPORTING_FEE_DIVISOR = 100; // 1% fees
     uint256 private constant MAXIMUM_REPORTING_FEE_DIVISOR = 10000; // Minimum .01% fees
+    uint256 private constant MINIMUM_REPORTING_FEE_DIVISOR = 3; // Maximum 33.3~% fees
```
MicahZoltu: Recommend enhancing this comment with a mention that any divisor lower than 2 _may_ result in bugs like divide-by-zero.  I don't want someone coming in here and adjusting it to 50% thinking that is a "reasonable default" and introducing a crasher.

source/contracts/reporting/ReportingWindow.sol
```diff
@@ -53,20 +53,23 @@ contract ReportingWindow is DelegationTarget, Extractable, ITyped, Initializable
...
+    function createMarket(address _sender, uint256 _endTime, uint8 _numOutcomes, uint256 _numTicks, uint256 _feePerEthInWei, ICash _denominationToken, address _designatedReporterAddress) public onlyInGoodTimes afterInitialized payable returns (IMarket _newMarket) {
         require(block.timestamp < startTime);
-        require(universe.getOrCreateReportingWindowByMarketEndTime(_endTime) == this);
-        MarketFactory _marketFactory = MarketFactory(controller.lookup("MarketFactory"));
-        getReputationToken().trustedReportingWindowTransfer(msg.sender, _marketFactory, universe.getOrCacheDesignatedReportNoShowBond());
-        _newMarket = _marketFactory.createMarket.value(msg.value)(controller, this, _endTime, _numOutcomes, _numTicks, _feePerEthInWei, _denominationToken, msg.sender, _designatedReporterAddress);
+        require(universe == IUniverse(msg.sender));
+        _newMarket = createMarketInternal(_endTime, _numOutcomes, _numTicks, _feePerEthInWei, _denominationToken, _sender, _designatedReporterAddress);
```
MicahZoltu: I'm curious if, now that we know the source of the stack-depth bug, we could get rid of the internal call here and have it inlined?  Looking at this method, I don't believe we'll have any stack depth issues if we move sender to later in the signature.

Not a necessary change, but would be nice to have fewer `internal` functions.

nuevoalex: I tried to put the parameters in the order the function receives them but that actually makes the stack depth error return. This was the magical configuration of parameters that worked,

nuevoalex: Same with inlining the function

source/contracts/reporting/Universe.sol
```diff
@@ -442,25 +397,16 @@ contract Universe is DelegationTarget, Extractable, ITyped, Initializable, IUniv
         if (_currentFeeDivisor > Reporting.getMaximumReportingFeeDivisor()) {
             _currentFeeDivisor = Reporting.getMaximumReportingFeeDivisor();
         }
+        if (_currentFeeDivisor < Reporting.getMinimumReportingFeeDivisor()) {
+            _currentFeeDivisor = Reporting.getMinimumReportingFeeDivisor();
+        }
```
MicahZoltu: Don't we have a `min` and `max` function on `uint256` that could be used here instead?  (same for other cases of taking the min or max above).

nuevoalex: We do! Will change

tests/reporting/test_universe.py
```diff
@@ -288,6 +288,62 @@ def test_universe_reporting_fee_divisor(localFixture, chain, populatedUniverse,
 
     assert populatedUniverse.getOrCacheReportingFeeDivisor() == defaultValue / 5 * multiplier
 
+''' TODO
```
MicahZoltu: What is the reason for this dangling TODO?

nuevoalex: Like why hasn't it been done or why do we still want to do it? There were unit tests for the reporting window previously that could in theory be recreated now for the universe. I delayed that in this PR but its still something that could be done.

Do you think there's value in doing it still?

MicahZoltu: I mean this TODO doesn't have a description and comments out what looks like a complete test.  Does this test fail right now?  Is it because the code is bad or the test is bad?  It sounds like the code is bad based on your description, do we believe that the code paths covered by this test are covered elsewhere (in which case we can delete this test) or do you believe that the code is currently untested, and fixing up this test would resolve that (in which case we should do that).

nuevoalex: Ah sorry. This is a unit test of the old reporting window market creation code which I moved here and commented out in it's entirety since that function no longer exists but one does for universe.

The universe market creation code is well tested but we could convert this outdated unit test into an equivalent on for the universe if desired. My thought was we would do that at some point.

PR #483, opened 2017-11-27
Upgrade Augur.js along with Augur-contracts

pgebheim:
This will update the prerelease version of augur-contracts and publish a new @dev channel pre-release version for augur.js along with augur-contracts.


support/artifacts/run.sh
```diff
@@ -1,5 +1,11 @@
 #!/bin/bash
 
+###
+# Setup the environments properly for merging and updating items in augur-contracts and augur.js
+# a lot of this is because we need to rely on Travis environment because it checks out in a
+# detached head state so we don't good branch information out of our local git repo
```
MicahZoltu: Style: don't put hard newlines in prose text.

PR #486, opened 2017-11-27
auto disavow

nuevoalex:
One major thing to note is that this change makes the stake token not do the universe based log on redeeming disavowed since we cannot actually confirm that the token is legitimately part of the universe.

We could add tracking of disavowed tokens if it is deemed necessary but I wanted to get this looked at and considered before doing something like that.

### DIFF
--- a/source/contracts/reporting/StakeToken.sol
+++ b/source/contracts/reporting/StakeToken.sol
@@ -23,6 +23,7 @@ contract StakeToken is DelegationTarget, Extractable, ITyped, Initializable, Var
     IMarket public market;
     uint256[] public payoutNumerators;
     bool private invalid;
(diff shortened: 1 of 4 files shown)


source/contracts/reporting/StakeToken.sol
```diff
@@ -226,14 +226,18 @@ contract StakeToken is DelegationTarget, Extractable, ITyped, Initializable, Var
     }
 
     function onBurn(address _target, uint256 _amount) internal returns (bool) {
+        // If the token is disavowed we cannot safely confirm that it is really a member of nay universe.
```
MicahZoltu: Typo: `nay` => `any`

PR #487, opened 2017-11-27
Cash delegated

nuevoalex:


source/contracts/trading/Cash.sol
```diff
@@ -5,13 +5,16 @@ import 'Controlled.sol';
...
 
 
 /**
  * @title Cash
  * @dev ETH wrapper contract to make it look like an ERC20 token.
  */
-contract Cash is Controlled, Extractable, ITyped, VariableSupplyToken, ICash {
+contract Cash is DelegationTarget, Extractable, ITyped, VariableSupplyToken, ICash {
+
+    //FIXME: Delegated contracts cannot currently use string values, so we will need to find a workaround if this hasn't been fixed before we release
```
MicahZoltu: We have a "fix" for this now I think, just needs some research and implementation.  @joeykrug dug into it a bit I believe?  Support was added in Byzantium.

MicahZoltu: 🤔 more on this, it is getting pretty late in the game.  We really should make a call on this ASAP since it will require work.

nuevoalex:
I'll make an issue to fix that problem in the column for requiring more discussion in CH

PR #490, opened 2017-11-28
README one liners

tinybike:
How to say "I'll be back" in shell script


MicahZoltu:
We are deleting the flag because the whole test suite takes so much longer than the fuzzy tests now that their run time is lost in the noise.  We may add it back if we ever can resolve that.

Therefore, this PR is unnecessary so I'm closing.  If anyone disagrees with this (@tinybike @nuevoalex) please re-open and comment!

PR #493, opened 2017-11-30
Market creation split into three methods. Description split out. Scalars can have min, max, and a normalizedNumTicks specified

nuevoalex:
For @tinybike the most important part to review is the new Universe API for creating markets and the market created log Event in Augur.sol to make sure this looks as expected.

### DIFF
--- a/source/contracts/Augur.sol
+++ b/source/contracts/Augur.sol
@@ -14,7 +14,7 @@ import 'libraries/Extractable.sol';
 
 // Centralized approval authority and event emissions
 contract Augur is Controlled, Extractable {
-    event MarketCreated(address indexed universe, bytes32 indexed topic, address indexed marketCreator, address market, uint256 marketCreationFee, int256 minPrice, int256 maxPrice, IMarket.MarketType marketType, bytes32 description, string extraInfo);
+    event MarketCreated(address indexed universe, bytes32 indexed topic, address indexed marketCreator, address market, uint256 marketCreationFee, int256 minPrice, int256 maxPrice, IMarket.MarketType marketType, string description, string extraInfo);
     event DesignatedReportSubmitted(address indexed universe, address indexed reporter, address indexed market, address stakeToken, uint256 amountStaked, uint256[] payoutNumerators);
     event ReportSubmitted(address indexed universe, address indexed reporter, address indexed market, address stakeToken, uint256 amountStaked, uint256[] payoutNumerators);
     event WinningTokensRedeemed(address indexed universe, address indexed reporter, address indexed market, address stakeToken, uint256 amountRedeemed, uint256 reportingFeesReceived, uint256[] payoutNumerators);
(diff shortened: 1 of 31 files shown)


source/contracts/reporting/Reporting.sol
```diff
@@ -67,4 +69,17 @@ library Reporting {
     function getMinimumReportingFeeDivisor() internal pure returns (uint256) { return MINIMUM_REPORTING_FEE_DIVISOR; }
     function getDefaultReportingFeeDivisor() internal pure returns (uint256) { return DEFAULT_REPORTING_FEE_DIVISOR; }
     function getInitialREPSupply() internal pure returns (uint256) { return INITIAL_REP_SUPPLY; }
+    function getBinaryMarketNumTicks() internal pure returns (uint256) { return BINARY_MARKET_NUM_TICKS; }
+
+    function getCategoricalMarketNumTicks(uint8 _numOutcomes) internal pure returns (uint256) {
+        require(_numOutcomes >= 2 && _numOutcomes <= 8);
+
+        if (_numOutcomes == 2) {return 10000;}
```
MicahZoltu: This is a binary market.  Recommend removing the special case logic for binary market and just retain this function only.

source/contracts/reporting/Universe.sol
```diff
@@ -420,10 +420,29 @@ contract Universe is DelegationTarget, Extractable, ITyped, Initializable, IUniv
...
+        _newMarket = _reportingWindow.createMarket.value(msg.value)(_endTime, _feePerEthInWei, _denominationToken, _designatedReporterAddress, msg.sender, _numOutcomes, Reporting.getCategoricalMarketNumTicks(_numOutcomes));
+        controller.getAugur().logMarketCreated(getOrCacheMarketCreationCost(), 0, 1, IMarket.MarketType.CATEGORICAL, _topic, _description, _extraInfo, this, _newMarket, msg.sender);
+        return _newMarket;
+    }
+
+    function createScalarMarket(uint256 _endTime, int256 _minPrice, int256 _maxPrice, uint256 _normalizedTicks, uint256 _feePerEthInWei, ICash _denominationToken, address _designatedReporterAddress, bytes32 _topic, bytes32 _description, string _extraInfo) public onlyInGoodTimes afterInitialized payable returns (IMarket _newMarket) {
+        require(_description.length > 0);
+        require(_minPrice < _maxPrice);
+        IReportingWindow _reportingWindow = getOrCreateReportingWindowByMarketEndTime(_endTime);
+        uint256 _numTicks = (uint256 (_maxPrice - _minPrice)).mul(_normalizedTicks);
```
MicahZoltu: Why not just have the user supply `numTicks` directly for scalars instead of having the contracts convert?

source/contracts/Augur.sol
```diff
@@ -14,7 +14,7 @@ import 'libraries/Extractable.sol';
 
 // Centralized approval authority and event emissions
 contract Augur is Controlled, Extractable {
-    event MarketCreated(address indexed universe, bytes32 indexed topic, address indexed marketCreator, address market, uint256 marketCreationFee, string extraInfo);
+    event MarketCreated(address indexed universe, bytes32 indexed topic, address indexed marketCreator, address market, uint256 marketCreationFee, int256 minPrice, int256 maxPrice, IMarket.MarketType marketType, bytes32 description, string extraInfo);
```
tinybike: `description` should be type `string`, not `bytes32`.

source/contracts/reporting/Universe.sol
tinybike: Hm, yes, I agree, it's simpler to just pass in `numTicks` here.  "Tick size" is simpler conceptually / harder for the creator to screw up, but it's a decimal, so we have to pass in `1/tickSize` instead, which isn't simpler than passing in `numTicks`.

nuevoalex: Will do

PR #494, opened 2017-11-30
Di time provider

nuevoalex:
In the deployment scripts it will set up Time correctly in the controller but from TS you would access the Time contract by requesting "TimeControlled"

### DIFF
--- a/source/contracts/Controller.sol
+++ b/source/contracts/Controller.sol
@@ -12,6 +12,7 @@ import 'IController.sol';
 import 'IControlled.sol';
 import 'libraries/token/ERC20Basic.sol';
 import 'libraries/Extractable.sol';
+import 'ITime.sol';
 
 
 contract Controller is IController {
@@ -177,5 +178,9 @@ contract Controller is IController {
 
     function getAugur() public view returns (Augur) {
         return Augur(lookup("Augur"));
+    }
+
+    function getTimestamp() public view returns (uint256) {
+        return ITime(lookup("Time")).getTimestamp();
     }
 }
--- a/source/contracts/ITime.sol
+++ b/source/contracts/ITime.sol
@@ -1,0 +1,16 @@
+pragma solidity 0.4.18;
+
+import 'Controlled.sol';
+import 'libraries/Initializable.sol';
+import 'libraries/ITyped.sol';
+import 'libraries/Ownable.sol';
+import 'libraries/Extractable.sol';
+
+
+contract ITime is Controlled, ITyped, Ownable, Extractable {
+    function getTimestamp() external view returns (uint256);
+
+    function getProtectedTokens() internal returns (address[] memory) {
+        return new address[](0);
+    }
+}
--- a/source/contracts/ITime.sol
+++ b/source/contracts/ITime.sol
@@ -3,11 +3,10 @@ pragma solidity 0.4.18;
 import 'Controlled.sol';
 import 'libraries/Initializable.sol';
 import 'libraries/ITyped.sol';
-import 'libraries/Ownable.sol';
 import 'libraries/Extractable.sol';
 
 
-contract ITime is Controlled, ITyped, Ownable, Extractable {
+contract ITime is Controlled, ITyped, Extractable {
     function getTimestamp() external view returns (uint256);
 
     function getProtectedTokens() internal returns (address[] memory) {
--- a/source/contracts/Time.sol
+++ b/source/contracts/Time.sol
@@ -1,0 +1,14 @@
+pragma solidity 0.4.18;
+
+import 'ITime.sol';
+
+
+contract Time is ITime {
+    function getTimestamp() external view returns (uint256) {
+        return block.timestamp;
+    }
+
+    function getTypeName() public view returns (bytes32) {
+        return "Time";
+    }
+}
--- a/source/libraries/Configuration.ts
+++ b/source/libraries/Configuration.ts
@@ -13,8 +13,9 @@ export class Configuration {
     public readonly contractInterfacesOutputPath: string;
     public readonly controllerAddress: string|undefined;
     public readonly createGenesisUniverse: boolean;
+    public readonly useControlledTime: boolean;
 
-    public constructor(host: string, port: number, gasPrice: BN, privateKey: string, contractSourceRoot: string, contractOutputRoot: string, controllerAddress: string|undefined, createGenesisUniverse: boolean=true) {
+    public constructor(host: string, port: number, gasPrice: BN, privateKey: string, contractSourceRoot: string, contractOutputRoot: string, controllerAddress: string|undefined, createGenesisUniverse: boolean=true, useControlledTime: boolean=true) {
         this.httpProviderHost = host;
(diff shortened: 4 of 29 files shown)


source/contracts/Controller.sol
```diff
@@ -178,4 +179,8 @@ contract Controller is IController {
     function getAugur() public view returns (Augur) {
         return Augur(lookup("Augur"));
     }
+
+    function getTimestamp() public view returns (uint256) {
+        return ITime(lookup("Time")).getTimestamp();
+    }
```
MicahZoltu: We should probably set some ground rules on what decides whether something gets a special getter added to controller.  I think `getTimestamp` may meet those requirements, whatever they are, but I would like to ensure that we are on the same page for what the bar is.

nuevoalex: It's difficult for me to say exactly why this felt right. In general it seems like small oft used lookups that could be classified at utilites feel like they belong here. Any more concrete proposal?

source/contracts/ITime.sol
```diff
@@ -0,0 +1,16 @@
+pragma solidity 0.4.18;
+
+import 'Controlled.sol';
+import 'libraries/Initializable.sol';
+import 'libraries/ITyped.sol';
+import 'libraries/Ownable.sol';
+import 'libraries/Extractable.sol';
+
+
+contract ITime is Controlled, ITyped, Ownable, Extractable {
```
MicahZoltu: I don't believe this contract needs to be `Controlled` or `ITyped`.  Probably not even `Ownable` either.  `Extractable` is good as a general rule, though users shouldn't _ever_ see the address for this contract so even that seems overkill.

Is there an argument for keeping all of these?

source/libraries/Configuration.ts
```diff
@@ -13,8 +13,9 @@ export class Configuration {
     public readonly contractInterfacesOutputPath: string;
     public readonly controllerAddress: string|undefined;
     public readonly createGenesisUniverse: boolean;
+    public readonly useControlledTime: boolean;
 
-    public constructor(host: string, port: number, gasPrice: BN, privateKey: string, contractSourceRoot: string, contractOutputRoot: string, controllerAddress: string|undefined, createGenesisUniverse: boolean=true) {
+    public constructor(host: string, port: number, gasPrice: BN, privateKey: string, contractSourceRoot: string, contractOutputRoot: string, controllerAddress: string|undefined, createGenesisUniverse: boolean=true, useControlledTime: boolean=true) {
```
MicahZoltu: We should probably change this flag to `isProduction` or something since there are now two things different about a deploy to production vs testnet:
1. `Time` vs `TimeControlled`
2. `LegacyReputationToken` vs `0xE94327D07Fc17907b4DB788E5aDf2ed424adDff6`

I suspect before launch there will be more things like this, though my hope is that we can keep the number as small as possible.

source/libraries/ContractInterfaces.ts
```diff
@@ -523,6 +523,13 @@ export class Controller extends Contract {
         return <Array<string>>result;
     }
 
+    public getTimestamp_ = async( options?: { sender?: string }): Promise<BN> => {
+        options = options || {};
+        const abi: AbiFunction = {"constant":true,"inputs":[],"name":"getTimestamp","outputs":[{"name":"","type":"uint256"}],"payable":false,"stateMutability":"view","type":"function"};
+        const result = await this.localCall(abi, [], options.sender);
+        return <BN>result[0];
+    }
```
MicahZoltu: I would have expected to see the addition of `Time` and `TimeControlled` contracts to this file, are they being ignored for some reason?

tests/trading/test_claimTradingProceeds.py
```diff
@@ -220,6 +220,6 @@ def test_reedem_failure(kitchenSinkFixture, cash, market):
         claimTradingProceeds.claimTradingProceeds(market.address, sender = tester.k1)
 
     # set timestamp to 3 days later (waiting period)
-    kitchenSinkFixture.chain.head_state.timestamp += long(timedelta(days = 3, seconds = 1).total_seconds())
+    kitchenSinkFixture.contracts["Time"].incrementTimestamp(long(timedelta(days = 3, seconds = 1).total_seconds()))
```
MicahZoltu: Big fan of this change in all of the python tests.  I wonder if this can simplify some things we had to do in conftest to deal with `time`?  Probably not worth the time to investigate if we are eventually moving off of python....

source/contracts/Time.sol
```diff
@@ -0,0 +1,14 @@
+pragma solidity 0.4.18;
+
+import 'ITime.sol';
```
MicahZoltu: I'm not a huge fan of leaving this to live in the root contracts folder, any reason to not put them in `libraries`?

source/contracts/ITime.sol
nuevoalex: Controlled is required for extractable. Also we've generally made any contract that is accesible via lookup controlled so it seems off we'd break that pattern. ITyped is useful to see which provider the current Time contract is. Ownable is at a minimum needed for the testing version of the time provider but we could remove it from the interface

source/libraries/Configuration.ts
nuevoalex: Yea good call, will do

source/contracts/Controller.sol
MicahZoltu: No, I was kind of hoping you had something concrete in your head that we could rally around.  :'(

source/libraries/ContractInterfaces.ts
nuevoalex: Hm good point. They should be added. Ill re run the generator

source/contracts/Time.sol
nuevoalex: Generally the libraries folder is for contracts that aren't actually uploaded. I wouldn't be opposed to a new folder. We could make a corw folder and put all the stragglers in there or a utils folder and just put these two there

tests/trading/test_claimTradingProceeds.py
nuevoalex: Ill still take a look and see. Might just save it for whenever i get time to work on test speed

source/contracts/ITime.sol
MicahZoltu: Fully convinced of `ITyped`'s value.

I would prefer to see `Ownable` pulled out to the implementation that actually needs it, rather than having it part of the interface since I don't think the interface needs it.

I don't see big value in `Extractable` (and therefore `Controlled`), but I also don't see much downside to including them either... so, meh on those.

source/contracts/Time.sol
MicahZoltu: Hmm... good point, I think we actually have code in the contract deployer that ignores the libraries folder.  I would prefer a folder for this kind of thing, possibly move `Augur`, `Controller` and `Controlled` in there too?  Those three contracts definitely feel disconnected from each other... okay, maybe the root is the best place until we start seeing enough stuff pop up to appropriately categorize.

PR #495, opened 2017-12-01
Deploy

tinybike:
Deploying the latest and greatest

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,17 +116,13 @@ pytest path/to/test_file.py -k 'name_of_test’
 
 By default the trading WCL fuzz testing is not run, if you want to run it specifically do
 ```
-export INCLUDE_FUZZY_TESTS=1
-pytest trading/test_wcl_fuzzy.py
-unset INCLUDE_FUZZY_TESTS
+INCLUDE_FUZZY_TESTS=1 pytest trading/test_wcl_fuzzy.py
 ```
 
 Or if you want to run all tests (including fuzzy):
 
 ```
-export INCLUDE_FUZZY_TESTS=1
-pytest
-unset INCLUDE_FUZZY_TESTS
+INCLUDE_FUZZY_TESTS=1 pytest
 ```
 
 When writing tests, it is highly recommended to make use of the ContractFixtures class for ‘placeholder’ variables. Python’s unit testing framework comes handy here; encapsulate tests within functions that start with "test\_", and use “assert” statements when testing for certain values. Parameterized tests are recommended as well to test various possibilities and edge cases.
--- a/README.md
+++ b/README.md
@@ -35,30 +35,32 @@ Travis CI is set up automatically deploy the contracts to the [Rinkeby testnet](
 
 Deployment can be run in two modes, direct or Docker. In direct mode, the assumption is that the entire system has been built (TypeScript and Solidity), and there is a working Node.js environment locally. Furthermore, one must possess an account on the deployment target (e.g. Rinkeby testnet) which has enough ETH to cover the costs of the deployment. For the purposes of deploying to the testnets, those with access to the augur private testnet keys can deploy and update the existing contracts. For reference, these keys are stored in an encrypted git repository within the Augur Project's keybase team. If you need access, please inquire in the Augur Discord.
 
-All deployment commands can be managed through scripts in package.json, and are accessible to `npm run` and `yarn`.
+All deployment commands can be managed through scripts in package.json, and can be executed using `npm run <command>`.
 
 ### Manual Rinkeby deployment
 
 Build and create compile artifacts:
 
 ```bash
-augur-core % npm run build
+npm run build
 ```
 
 Deploy to Rinkeby:
 
 ```bash
-augur-core % RINKEBY_PRIVATE_KEY=$(cat path/to/keys/deploy_keys/rinkeby.prv) npm run deploy:rinkeby
+RINKEBY_PRIVATE_KEY=$(cat path/to/keys/deploy_keys/rinkeby.prv) npm run deploy:rinkeby
 ```
 
 ### Updating augur-contracts
 
+(Note: commands _in this section only_ should be run from your local augur-contracts folder, not from augur-core!)
+
 Manual deployment generates artifacts in the `output/contracts` directory.  To merge these changes into augur-contracts, there are scripts located in [the augur-contracts repository](https://github.com/AugurProject/augur-contracts). **In local deployments, this is not automatic.**
 
-To merge the local changes into the augur-contracts repository:
+To merge the local changes into augur-contracts:
 
 ```bash
-augur-contracts % SOURCE="path/to/augur-core/output/contracts" BRANCH=master npm run update-contracts
+SOURCE="path/to/augur-core/output/contracts" BRANCH=master npm run update-contracts
 ```
 
 Next, update the version of augur-contracts:
@@ -67,14 +69,14 @@ Next, update the version of augur-contracts:
  - for _releases_, set the version to be the correct major/minor version; e.g., for dev build v4.7.0-10, the release version would be v4.7.0
 
 ```bash
-augur-contracts % npm version [<version>, major,minor,patch, prerelease]
-augur-contracts % git push && git push --tags
+npm version [<version>, major,minor,patch, prerelease]
+git push && git push --tags
 ```
 
 Finally, publish to NPM.  If this is a pre-release tag, deploy it to the `dev` channel.  The `dev` channel is the default for versions which are published from CI.
 
 ```bash
-augur-contracts % npm publish [--tag dev]
+npm publish [--tag dev]
 ```
 
 ## Tests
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -2,92 +2,82 @@
 
 [![Build Status](https://travis-ci.org/AugurProject/augur-core.svg)](https://travis-ci.org/AugurProject/augur-core)
 
-Ethereum contracts for a decentralized prediction market platform.
+Smart contracts for [Augur](https://augur.net), a decentralized prediction market platform on the [Ethereum](https://ethereum.org) blockchain.
 
 ## Installation
 
-You should already have a system-wide installation of Python and it should be Python 2.7.6 or greater.
+You need a system-wide installation of Python (v2.7.6 or higher).  Install the dependencies:
```
MicahZoltu: Since you seem to be on a Readme updating kick, probably worth mentioning here that you need [(Python and Node) or Docker] installed to work with this repository.  I believe we are currently all on Node 8, anything else may or may not work.

tinybike: My true passion is updating readmes

README.md
```diff
@@ -2,92 +2,82 @@
...
+```bash
 npm install npx
 npm install
-
-sudo pip install -r requirements.txt
+pip install -r requirements.txt
 ```
 
-For PyEthereum to work properly you'll also need to install the 0.4.18 Solc compiler on your machine. Instructions for installing it in your environment can be found here:
+The dependencies include [pyethereum](https://github.com/ethereum/pyethereum), which is used to unit test our Ethereum smart contracts.  If you want to run the the augur-node tests locally, you also need to install [version 0.4.18 of the Solidity compiler](https://github.com/ethereum/solidity/releases/tag/v0.4.18) (`solc`) on your machine for pyethereum to run correctly.  (On macOS, you need to use [virtualenv](https://python-guide-pt-br.readthedocs.io/en/latest/dev/virtualenvs/) or [homebrew](https://brew.sh/) Python to work around System Integrity Protection.)
```
MicahZoltu: "If you want to run the augur-node tests locally": The only other action you can take on this repo is to deploy, and you should _NOT_ be deploying locally and if you do you should not do so without running the tests first.  I recommend deleting a good chunk of this section and just including Solidity in the dependency set above (see other comment).  Also saying what is in `requirements.txt` seems a bit unnecessary, if the person wants to know what the PIP dependencies are they can look at `requirements.txt`, just like they can look at `package.json` if they want to know about our NPM dependencies.  The bit about virtualenv or homebrew to get things working on MacOS is valuable if it is non-obvious (not familiar with Mac/Python).

README.md
```diff
@@ -2,92 +2,82 @@
...
-### Build:
+### Local deployment
 
-```
-npm run docker:build
-```
+Deployment can be run in two modes, direct or Docker. In direct mode, the assumption is that the entire system has been built (TypeScript and Solidity), and there is a working Node.js environment locally. Furthermore, one must possess an account on the deployment target (e.g. Rinkeby testnet) which has enough ETH to cover the costs of the deployment. For the purposes of deploying to the testnets, those with access to the augur private testnet keys can deploy and update the existing contracts. For reference, these keys are stored in an encrypted git repository within the Augur Project's keybase team. If you need access, please inquire in the Augur Discord.
 
-### Test:
+All deployment commands can be managed through scripts in package.json, and are accessible to `npm run` and `yarn`.
```
MicahZoltu: Has anyone actually tested with `yarn`?

README.md
```diff
@@ -2,92 +2,82 @@
...
+### Manual Rinkeby deployment
 
-# Build and run all unit tests and integration tests
-npm run docker:test
+Build and create compile artifacts:
 
-# Build and run just integration test
-npm run docker:test:integration
+```bash
+augur-core % npm run build
```
MicahZoltu: Recommend removing the `augur-core % ` prefix on these commands.  It is unclear to the casual observer whether that is part of the command or just how _your_ prompt looks.  Many people have prompts that look different, and this isn't even the default prompt.

README.md
tinybike: I certainly haven't!

README.md
tinybike: Good call.

README.md
```diff
@@ -2,92 +2,82 @@
...
-- [Augur Master Plan](https://medium.com/@AugurProject/augur-master-plan-42dda65a3e3d)
-
-## Terminology
+### Updating augur-contracts
 
-[Augur Terminology](http://blog.augur.net/faq/all-terms/)
+Manual deployment generates artifacts in the `output/contracts` directory.  To merge these changes into augur-contracts, there are scripts located in [the augur-contracts repository](https://github.com/AugurProject/augur-contracts). **In local deployments, this is not automatic.**
 
-## Src Folders
+To merge the local changes into the augur-contracts repository:
```
MicahZoltu: I'm generally against including manual deployment instructions in readmes.  No one should _ever_ be doing a manual deploy, so there is no need to include instructions.  If CI is broken, fix CI, don't give up and manually deploy (dangerous).  If for some unfathomable reason we _must_ deploy manually, a person can deduce the process from the `.travisci` file.

README.md
```diff
@@ -99,94 +89,98 @@ The tests directory (augur-core/tests) contain tests and test fixtures to test t
 - test_controller.py -- tests controller functionalities.
 - test_mutex.py -- tests mutex functionalities.
 - test_helpers.py -- tests the controller, safeMath, and assertNoValue macros.
-- test_legacyRep.py -- tests the legacyRepTokens’ functionalities.
+- test_legacyRep.py -- tests the legacyRepTokens' functionalities.
```
MicahZoltu: I know this wasn't you, but this `'` should be before the `s`.

README.md
```diff
@@ -99,94 +89,98 @@ The tests directory (augur-core/tests) contain tests and test fixtures to test t
...
 
-By default the trading WCL fuzz testing is not run, if you want to run it specifically do
+```bash
+pytest path/to/test_file.py -k 'name_of_test'
 ```
-export INCLUDE_FUZZY_TESTS=1
-pytest trading/test_wcl_fuzzy.py
-unset INCLUDE_FUZZY_TESTS
+
+The unit test suite includes tests that use a fuzzer, which take somewhat longer (around 10 minutes on CI) to run than the rest of the tests.  These tests are _not_ run by default; to include these tests, set the environment variable `INCLUDE_FUZZY_TESTS=1`:
```
MicahZoltu: Again not you, but `INCLUDE_FUZZY_TESTS` is no longer necessary as the flag was removed.

README.md
```diff
@@ -99,94 +89,98 @@ The tests directory (augur-core/tests) contain tests and test fixtures to test t
...
+- [A Roadmap For Augur and What's Next](https://medium.com/@AugurProject/a-roadmap-for-augur-and-whats-next-930fe6c7f75a)
+- [Augur Master Plan](https://medium.com/@AugurProject/augur-master-plan-42dda65a3e3d)
+
+### Terminology
+
+[Augur Terminology](http://blog.augur.net/faq/all-terms/)
+
+### Security model
+
+The security model is based on the idea that reporters will want to invest in a market where honesty prevails, and will not trust fraudulent markets. For example, assume that there exist two almost identical universes: an honest universe, where the factually correct answer is the predominant answer, and a fraudulent universe, where incorrect answers are predominant in order to cheat the system. The REP in the fraudulent universe will approach 0, and the REP in the honest universe will approach the theoretical yield. This phenomenon occurs because reporters do not want to trade on markets in a dishonest universe. Therefore, they will not create markets in a dishonest universe, and the REP holders will not receive any market fees.
```
MicahZoltu: "reporters do not want to trade" should be "traders do not want to trade"

README.md
```diff
@@ -99,94 +89,98 @@ The tests directory (augur-core/tests) contain tests and test fixtures to test t
...
+
+The reentrancy guard (libraries/ReentrancyGuard.sol) and controller (Controller.sol) contracts also help uphold security. One can also attack Ethereum contracts by creating a chain of wanted contract calls through a malicious user. The reentrancy guard protects against that by preventing someone from re-entering the contracts before they return. The controller contract serves to update Augur and stores a registry of whitelisted contract addresses. The controller also can perform administrative actions, such as transferring ownership and performing emergency stop functions. It can be operated in either the developer mode, where the developer updates contracts, or the decentralized mode, when people must upload contracts to make changes and therefore, the market creators/traders must cancel orders and shift to the new contract. For bugs found involving money or REP, contracts are locked / an escape hatch is enabled which allows withdrawal of shares (and cancelling orders) for a value half-way between the bid-ask spread at the time of the lock. In addition, another escape hatch is enabled, which allows withdrawal of any locked up REP.
```
MicahZoltu: The mention of ReentrancyGuard, Controller, and Escape Hatches seem out of place here.  Consider just including these as doc-comments on the respective contracts.  I'm generally a fan of non-obvious classes (contracts) having a doc comment explaining their purpose (but not their implementation).  I think having text that describes the purpose of source code in the readme just leads to it going out of date much quicker, and also makes it non-obvious for the reader on where to find the information (it is disconnected).

README.md
```diff
@@ -99,94 +89,98 @@ The tests directory (augur-core/tests) contain tests and test fixtures to test t
...
 
-There are no floats in the EVM.
+### EVM numbers are always integers
 
-All augur-core contracts use integers. So ether values in contracts will always be represented as integers whose value is in attoeth (aka: wei, 10^-18 ether).
+There are no floating point numbers in the EVM. All augur-core contracts use integers. So ether values in contracts will always be represented in units of wei (AKA attoether AKA 10^-18 ether).
```
MicahZoltu: Consider mentioning that REP is also denominated this same way?

README.md
tinybike: Larger point: should we just remove this abbreviated section on the security model?  I'm not a huge fan of having this stuff both here _and_ in the whitepaper.

MicahZoltu: Yeah, I'm all for deleting this section.  I didn't realize this was a move, not an addition.  👍 to delete.

PR #496, opened 2017-12-01
Stake redemption convenience

nuevoalex:


source/contracts/reporting/DisputeBond.sol
```diff
@@ -51,7 +63,7 @@ contract DisputeBond is DelegationTarget, Extractable, ITyped, Initializable, Ow
         return true;
     }
 
-    function withdrawDisavowedTokens() public onlyOwner onlyInGoodTimes returns (bool) {
+    function withdrawDisavowedTokens() public onlyInGoodTimes returns (bool) {
```
nuevoalex: I took the owner requirement off as there is no harm I can perceive in someone else paying the gas to have the bond redeemed on behalf of whoever owns it when it is in disavowed state

source/contracts/reporting/DisputeBond.sol
```diff
@@ -37,6 +37,18 @@ contract DisputeBond is DelegationTarget, Extractable, ITyped, Initializable, Ow
     }
 
     function withdraw(bool forgoFees) public onlyOwner onlyInGoodTimes returns (bool) {
+        withdrawInternal(forgoFees);
+        return true;
+    }
+
+    function withdrawForHolder(address _sender, bool forgoFees) public onlyInGoodTimes returns (bool) {
+        require(market.getUniverse() == IUniverse(msg.sender));
```
MicahZoltu: Style: We haven't discussed this, but I prefer to have the left side of comparisons be the thing you are checking, and the right side to be the thing you are checking against.

nuevoalex: We do it in the style above everywhere :[

MicahZoltu: Is this your preferred style, or did you just copy it from somewhere and run with it?  If people generally prefer this style I'm 👍 on saying "this is how we do things".  However, if we kind of fell into it then I would recommend we figure out how we want to do things, say that is our style, and as we touch code fix it to match desired style.  For this I don't think there is a _huge_ downside to having mixed style while we iterate, though I do think we should be working towards a consistent style eventually.

nuevoalex: I have no preference so I'm fine with doing left side being the thing you're checking going forward. Can change this instance too.

source/contracts/reporting/StakeToken.sol
```diff
@@ -96,33 +97,55 @@ contract StakeToken is DelegationTarget, Extractable, ITyped, Initializable, Var
...
+        redeemWinningTokensInternal(_sender, forgoFees);
+        return true;
+    }
+
+    function redeemWinningTokens(bool forgoFees) public onlyInGoodTimes afterInitialized returns (bool) {
+        redeemWinningTokensInternal(msg.sender, forgoFees);
         return true;
     }
 
     // NOTE: UI should warn users about calling this before first calling `migrateLosingTokens` on all losing tokens with non-dust contents
```
MicahZoltu: This comment should probably be moved to public methods.

PR #497, opened 2017-12-02
delegator support for variable length data

nuevoalex:


source/contracts/libraries/Delegator.sol
```diff
@@ -35,15 +35,16 @@ contract Delegator is DelegationTarget {
...
+            let _retval := delegatecall(gas, _target, _calldataMemoryOffset, calldatasize, 0, 0)
             switch _retval
             case 0 {
                 // 0 == it threw, so we revert
                 revert(0,0)
             } default {
-                // If the call succeeded return the data in mem[_calldataMemoryOffset..32]
-                return(_calldataMemoryOffset, 32)
+                // If the call succeeded return the return data from the delegate call
+                returndatacopy(0, 0, returndatasize)
```
MicahZoltu: Wasn't there some problem with using offset `0` for storing the return value, which is why we instead store it at `_calldataMemoryOffset`?

source/libraries/ContractCompiler.ts
```diff
@@ -35,10 +35,15 @@ export class ContractCompiler {
         const compilerOutput: CompilerOutput = JSON.parse(compilerOutputJson);
         if (compilerOutput.errors) {
             let errors = "";
+
             for (let error of compilerOutput.errors) {
+                if (error.message.includes("instruction is only available after the Metropolis hard fork")) continue;
```
MicahZoltu: :'(

nuevoalex: Yea :[

source/contracts/libraries/Delegator.sol
nuevoalex: If so I can't recall it now. Things seem copacetic with all our use cases just using 0.

MicahZoltu: Ah, I remember now.  `_calldataMemoryOffset` is the "right" thing to do.  Also, we should be updating that pointer to point after the returndata as well, just like we do when we write the calldata to memory.  `mstore(0x40, add(_returndataMemoryOffset, returndatasize))`

MicahZoltu: Hmm, thinking more on this it is unclear to me how any of this can possibly work...

nuevoalex:
There are already at least two open issues for disabling solc compiler warnings  so I'm not sure there's value except in possibly poking on those. Will definitely add a comment noting that once we can do so we should just have the compiler ignore those.

MicahZoltu:
I think the value is in making sure that the Solidity developers are aware of the problem.  I wouldn't be surprised if the warning was added pre-Byzantium and they forgot to take it out.  I am not opposed to this being added to one of the existing issues, but I personally would create a new issue titled: "Remove byzantium-only warning." and see where it goes.

nuevoalex:
Review addressed!

PR #502, opened 2017-12-06
Assorted small issues

nuevoalex:
- add hints to the public trade functions
- make the createCategoricalMarket function take a list of outcomes instead of a number of them

### DIFF
--- a/source/contracts/reporting/Reporting.sol
+++ b/source/contracts/reporting/Reporting.sol
@@ -68,7 +68,7 @@ library Reporting {
     function getDefaultReportingFeeDivisor() internal pure returns (uint256) { return DEFAULT_REPORTING_FEE_DIVISOR; }
     function getInitialREPSupply() internal pure returns (uint256) { return INITIAL_REP_SUPPLY; }
 
-    function getCategoricalMarketNumTicks(uint256 _numOutcomes) internal pure returns (uint256) {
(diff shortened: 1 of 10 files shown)


source/contracts/reporting/Reporting.sol
```diff
@@ -68,7 +68,7 @@ library Reporting {
     function getDefaultReportingFeeDivisor() internal pure returns (uint256) { return DEFAULT_REPORTING_FEE_DIVISOR; }
     function getInitialREPSupply() internal pure returns (uint256) { return INITIAL_REP_SUPPLY; }
 
-    function getCategoricalMarketNumTicks(uint8 _numOutcomes) internal pure returns (uint256) {
+    function getCategoricalMarketNumTicks(uint256 _numOutcomes) internal pure returns (uint256) {
```
MicahZoltu: This change seems incorrect.  Valid values are in the range 2-8, which fits _well_ within uint8.  Was there some other reason this needed to be `uint256`?

nuevoalex: We can cast to uint8 in the caller so this can change back. Did that in the other use case but mustve missed it here

nuevoalex: More to the point, we actually receive a uint256 implicitly through calling length now

PR #503, opened 2017-12-07
log reporting window creation

nuevoalex:
@epheph let me know if the event data looks sufficient for augur node's needs


MicahZoltu:
👍 from me.  I'll let @epheph give the final PR approval.

PR #504, opened 2017-12-08
Test speed

nuevoalex:


tests/fuzzy/test_orders_fuzzy.py
```diff
@@ -17,45 +17,45 @@
 @mark.parametrize('orderType,numOrders,withBoundingOrders,deadOrderProbability', [
     (BID,  10, False,  0.0),
     (ASK,  10, False,  0.0),
-    (BID,  50, False,  0.0),
-    (ASK,  50, False,  0.0),
-    (BID, 100, False,  0.0),
-    (ASK, 100, False,  0.0),
+    (BID,  20, False,  0.0),
```
nuevoalex: I picked these since it's at least doubling instead of totally arbitrary and because this makes the fuzzy test speed run on part with the reporting suite so it isn't the bottleneck in CI

tests/reporting/test_dispute_bond_token_rep_redemption.py
```diff
@@ -17,117 +17,63 @@
 FIRST_REPORTERS_DISPUTE_BOND_AMOUNT = 11 * 10**21
 LAST_REPORTERS_DISPUTE_BOND_AMOUNT = 1375 * 10**20
 
-MARKET_TYPE_CATEGORICAL = 0
 MARKET_TYPE_SCALAR = 1
 
-CATEGORICAL_OUTCOME_A = [10002, 0, 0]
```
nuevoalex: I got rid of the categorical tests since these take an enormous amount of time and doing both payout types is redundant. If we end up rewriting reporting it will be extra moot too

source/libraries/ContractInterfacesGenerator.ts
```diff
@@ -169,6 +169,9 @@ ${contractMethods.join("\n\n")}
             case 'uint256[]': {
                 return 'Array<BN>';
             }
+            case 'bytes32[]': {
+                return 'Array<string>';
```
MicahZoltu: Was this added to this PR opportunistically, or is it necessary as part of this PR?  I don't mind, just want to make sure I understand how it is or isn't associated with the rest of the changes.

nuevoalex: I'm actually uncertain how tests passed without this. Noted in my description that some changes here were to get tests passing

nuevoalex:
np!

PR #509, opened 2017-12-22
Critical fixes

nuevoalex:
Audit report fixes except for the trading vulnerability.

Also updated some constants

### DIFF
--- a/source/contracts/reporting/Reporting.sol
+++ b/source/contracts/reporting/Reporting.sol
@@ -10,7 +10,7 @@ library Reporting {
     uint256 private constant INITIAL_REP_SUPPLY = 11 * 10 ** 6 * 10 ** 18; // 11 Million REP
     // CONSIDER: Should this be a constant?
     uint256 private constant DISPUTE_THRESHOLD_FOR_FORK = INITIAL_REP_SUPPLY / 80; // 1.25% of the REP total supply
-    uint256 private constant INITIAL_REPORT_MIN_VALUE = DISPUTE_THRESHOLD_FOR_FORK / 2**20; // This value will result in a maximum 20 round dispute sequence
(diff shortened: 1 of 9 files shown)


source/contracts/reporting/Reporting.sol
```diff
@@ -22,8 +22,8 @@ library Reporting {
     uint256 private constant MINIMUM_REPORTING_FEE_DIVISOR = 3; // Maximum 33.3~% fees. Note than anything less than a value of 2 here will likely result in bugs such as divide by 0 cases.
 
     // NOTE: We need to maintain this cost to roughly match the gas cost of reporting. This was last updated 10/02/2017
```
MicahZoltu: This date should be updated.

source/contracts/trading/CancelOrder.sol
```diff
@@ -56,15 +59,12 @@ contract CancelOrder is CashAutoConverter, Extractable, ReentrancyGuard, ICancel
             if (_type == Order.Types.Bid) {
                 for (uint8 _i = 0; _i < _market.getNumberOfOutcomes(); ++_i) {
                     if (_i != _outcome) {
-                        _market.getShareToken(_i).transfer(_sender, _sharesEscrowed);
+                        _market.getShareToken(_i).trustedCancelOrderTransfer(_market, _sender, _sharesEscrowed);
```
MicahZoltu: I'm assuming `_market` being unvalidated here is on the TODO list for later?  Should I assume the same for all cases of trusted inputs throughout this PR, or should I call them all out to be sure they don't slip through the cracks?

nuevoalex: Yea I'm doing all of that in another PR currently.

nuevoalex: Which i intend to sit on for a bit until we can get auditor feedback on the two options

PR #510, opened 2017-12-23
Store valid universes and use to validate trading contracts and logs. Genesis universe creation from `Augur`

nuevoalex:
Doing a PR just so it can be looked at but I'll try to sit on this until we can get feedback from the auditors still.

### DIFF
--- a/source/contracts/Augur.sol
+++ b/source/contracts/Augur.sol
@@ -46,8 +46,8 @@ contract Augur is Controlled, Extractable {
     }
 
     function createChildUniverse(bytes32 _parentPayoutDistributionHash) public returns (IUniverse) {
-        require(universes[msg.sender]);
         IUniverse _parentUniverse = IUniverse(msg.sender);
+        require(isKnownUniverse(_parentUniverse));
         return createUniverse(_parentUniverse, _parentPayoutDistributionHash);
     }
 
@@ -58,8 +58,8 @@ contract Augur is Controlled, Extractable {
         return _newUniverse;
     }
 
-    function isValidUniverse(IUniverse _universe) public view returns (bool) {
-        return universes[address(_universe)];
+    function isKnownUniverse(IUniverse _universe) public view returns (bool) {
+        return universes[_universe];
     }
 
     //
@@ -78,7 +78,7 @@ contract Augur is Controlled, Extractable {
 
     // This signature is intended for the categorical market creation. We use two signatures for the same event because of stack depth issues which can be circumvented by maintaining order of paramaters
     function logMarketCreated(bytes32 _topic, string _description, string _extraInfo, IUniverse _universe, address _market, address _marketCreator, bytes32[] _outcomes, int256 _minPrice, int256 _maxPrice, IMarket.MarketType _marketType) public returns (bool) {
-        require(isValidUniverse(_universe));
+        require(isKnownUniverse(_universe));
         require(_universe == IUniverse(msg.sender));
         MarketCreated(_topic, _description, _extraInfo, _universe, _market, _marketCreator, _outcomes, _universe.getOrCacheMarketCreationCost(), _minPrice, _maxPrice, _marketType);
         return true;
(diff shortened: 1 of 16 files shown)


source/contracts/Augur.sol
```diff
@@ -34,6 +35,33 @@ contract Augur is Controlled, Extractable {
...
+
+    function createUniverse(IUniverse _parentUniverse, bytes32 _parentPayoutDistributionHash) private returns (IUniverse) {
+        UniverseFactory _universeFactory = UniverseFactory(controller.lookup("UniverseFactory"));
+        IUniverse _newUniverse = _universeFactory.createUniverse(controller, _parentUniverse, _parentPayoutDistributionHash);
+        universes[_newUniverse] = true;
+        return _newUniverse;
+    }
+
+    function isValidUniverse(IUniverse _universe) public view returns (bool) {
+        return universes[address(_universe)];
```
MicahZoltu: Is this cast really necessary?  Feels like it should be an implicit cast since it is a downcast...

source/contracts/Augur.sol
```diff
@@ -34,6 +35,33 @@ contract Augur is Controlled, Extractable {
...
+    //
+    // Universe
+    //
+
+    function createGenesisUniverse() public returns (IUniverse) {
+        return createUniverse(IUniverse(0), bytes32(0));
+    }
+
+    function createChildUniverse(bytes32 _parentPayoutDistributionHash) public returns (IUniverse) {
+        require(universes[msg.sender]);
```
MicahZoltu: Recommend changing this to `require(isValidUniverse(msg.sender))` to make it a tad more clear what the check is doing.

source/contracts/Augur.sol
```diff
@@ -34,6 +35,33 @@ contract Augur is Controlled, Extractable {
...
+    }
+
+    function createUniverse(IUniverse _parentUniverse, bytes32 _parentPayoutDistributionHash) private returns (IUniverse) {
+        UniverseFactory _universeFactory = UniverseFactory(controller.lookup("UniverseFactory"));
+        IUniverse _newUniverse = _universeFactory.createUniverse(controller, _parentUniverse, _parentPayoutDistributionHash);
+        universes[_newUniverse] = true;
+        return _newUniverse;
+    }
+
+    function isValidUniverse(IUniverse _universe) public view returns (bool) {
```
MicahZoltu: Consider renaming this to `isKnownUniverse`?  `valid` _feels_ like the wrong word to me here as a universe can be valid but not blessed.  Maybe `Blessed` is better?

source/contracts/libraries/MarketValidator.sol
```diff
@@ -0,0 +1,14 @@
+pragma solidity 0.4.18;
+
+import 'reporting/IMarket.sol';
+import 'Controlled.sol';
+
+
+contract MarketValidator is Controlled {
+    modifier marketIsValid(IMarket _market) {
```
MicahZoltu: Another idea to follow some naming elsewhere: `marketIsLegit`

source/contracts/trading/CancelOrder.sol
```diff
@@ -21,12 +22,12 @@ import 'libraries/Extractable.sol';
...
  * @dev This allows you to cancel orders on the book.
  */
-contract CancelOrder is CashAutoConverter, Extractable, ReentrancyGuard, ICancelOrder {
+contract CancelOrder is CashAutoConverter, Extractable, ReentrancyGuard, MarketValidator, ICancelOrder {
     /**
      * @dev Cancellation: cancels an order, if a bid refunds money, if an ask returns shares
      * @return true if successful; throw on failure
      */
-    function cancelOrder(bytes32 _orderId, Order.Types _type, IMarket _market, uint8 _outcome) nonReentrant convertToAndFromCash external returns (bool) {
+    function cancelOrder(bytes32 _orderId, Order.Types _type, IMarket _market, uint8 _outcome) marketIsValid(_market) nonReentrant convertToAndFromCash external returns (bool) {
```
MicahZoltu: Did you already do a find-in-files for functions that take an `IMarket` parameter?

I wonder if we should also have a similar modifier for `IUniverse` and apply it everywhere?

source/contracts/trading/Order.sol
```diff
@@ -53,6 +53,7 @@ library Order {
     // Constructor
     //
 
+    // No validation is needed here as it is simply a librarty function for organizing data
```
MicahZoltu: While this is true if we only look one layer deep, is it still true that all functions that take an `Order.Data` properly validate the `.market` property?  My guess is not, which means we should do validation here.

nuevoalex: Its a library so even if we wanted to we could not. My point is that because this is a library it is effectively not an external contract call

source/contracts/trading/CancelOrder.sol
nuevoalex: Ill double check!

source/contracts/Augur.sol
nuevoalex: Nope! Will change

source/contracts/Augur.sol
nuevoalex: `Known` sounds good to me

source/contracts/trading/Order.sol
nuevoalex: More on this, any caller would also be passing in the used market, so they should have done validation

PR #512, opened 2017-12-28
Universe changes

nuevoalex:
Includes the change to allow child universe creation manually and the ability to update total rep supply as it pertains to deriving fork related values.

We'll need to update the docs to include instruction on how to trigger said updating

### DIFF
--- a/source/contracts/reporting/ReputationToken.sol
+++ b/source/contracts/reporting/ReputationToken.sol
@@ -23,11 +23,15 @@ contract ReputationToken is DelegationTarget, Extractable, ITyped, Initializable
     uint256 constant public decimals = 18;
     IUniverse private universe;
     uint256 private totalMigrated;
+    mapping(address => uint256) migratedToSibling;
+    uint256 private parentTotalTheoreticalSupply;
+    uint256 private totalTheoreticalSupply;
 
     function initialize(IUniverse _universe) public onlyInGoodTimes beforeInitialized returns (bool) {
         endInitialization();
         require(_universe != address(0));
         universe = _universe;
+        updateParentTotalTheoreticalSupply();
         return true;
     }
 
@@ -46,7 +50,9 @@ contract ReputationToken is DelegationTarget, Extractable, ITyped, Initializable
         totalMigrated += _attotokens;
         // Award a bonus if migration is done before the fork has resolved and update the universe tentative winner tracking
         if (!_parentUniverse.getForkingMarket().isFinalized()) {
-            mint(_reporter, _attotokens.div(Reporting.getForkMigrationPercentageBonusDivisor()));
+            uint256 _bonus = _attotokens.div(Reporting.getForkMigrationPercentageBonusDivisor());
+            mint(_reporter, _bonus);
+            totalTheoreticalSupply += _bonus;
             _parentUniverse.updateTentativeWinningChildUniverse(universe.getParentPayoutDistributionHash());
         }
         return true;
@@ -64,7 +70,9 @@ contract ReputationToken is DelegationTarget, Extractable, ITyped, Initializable
         IUniverse _parentUniverse = universe.getParentUniverse();
         IReportingParticipant _reportingParticipant = IReportingParticipant(msg.sender);
         require(_parentUniverse.isContainerForReportingParticipant(_reportingParticipant));
-        mint(_reportingParticipant, _amountMigrated / 2);
+        uint256 _bonus = _amountMigrated / 2;
+        mint(_reportingParticipant, _bonus);
+        totalTheoreticalSupply += _bonus;
         return true;
     }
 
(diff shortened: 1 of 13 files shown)


source/contracts/reporting/ReputationToken.sol
```diff
@@ -108,6 +116,36 @@ contract ReputationToken is DelegationTarget, Extractable, ITyped, Initializable
...
+        return totalMigrated;
+    }
+
+    function updateSiblingMigrationTotal(IReputationToken _token) public returns (bool) {
+        require(_token != this);
+        IUniverse _supposedUniverse = _token.getUniverse();
+        require(_token == universe.getParentUniverse().getChildUniverse(_supposedUniverse.getParentPayoutDistributionHash()).getReputationToken());
+        totalTheoreticalSupply += migratedToSibling[_token];
+        migratedToSibling[_token] = _token.getTotalMigrated();
+        totalTheoreticalSupply -= migratedToSibling[_token];
```
bthaile: totalTheoreticalSupply gets added to then subtracted from, not following reason why.

source/contracts/reporting/ReputationToken.sol
```diff
@@ -108,6 +116,36 @@ contract ReputationToken is DelegationTarget, Extractable, ITyped, Initializable
...
+        return true;
+    }
+
+    function updateParentTotalTheoreticalSupply() public returns (bool) {
+        IUniverse _parentUniverse = universe.getParentUniverse();
+        totalTheoreticalSupply -= parentTotalTheoreticalSupply;
+        if (_parentUniverse == IUniverse(0)) {
```
bthaile: Is IUniverse(0) the genesis universe?

source/contracts/reporting/ReputationToken.sol
```diff
@@ -108,6 +116,36 @@ contract ReputationToken is DelegationTarget, Extractable, ITyped, Initializable
...
+            parentTotalTheoreticalSupply = Reporting.getInitialREPSupply();
+        } else {
+            parentTotalTheoreticalSupply = _parentUniverse.getReputationToken().getTotalTheoreticalSupply();
+        }
+        totalTheoreticalSupply += parentTotalTheoreticalSupply;
```
bthaile: Still not following the subtraction then adding to totalTheoreticalSupply. This is called from the constructor, what is thought to be parentTotalTheoreticalSupply initially (0)?

source/contracts/reporting/ReputationToken.sol
nuevoalex: These functions exist to allow someone off chain to request that the theoretical supply be adjusted because of new information. In this case they are indicating that REP was migrated to a sibling universe and asking that the theoretical supply be lowered accordingly.

Since we may have already accounted for this siblings migration total in the past we need to first add back whatever we previously subtracted, then update the amount we're using in the calculation, then subtract it.

source/contracts/reporting/ReputationToken.sol
nuevoalex: IUnivers(0) is the default/null IUniverse value. This is really saying "cast the address 0 to an IUniverse interface". This line is testing to see if the universe has a parent. If the parent is IUniverse(0) then it is a genesis universe.

source/contracts/reporting/ReputationToken.sol
nuevoalex: Explained the theoretical supply calculations above. Initially it is 0. Solidity defaults to all uninitialized values being 0 or for bools false.

source/contracts/reporting/ReputationToken.sol
bthaile: ok, so the total theoretical supply is essentially rolled back then updated based on the migrated rep to sib universe.

nuevoalex: Yup! We have to do this since we can't trust the off chain provider to give us a delta. We have to recompute basedo n what the old value was and check what it actually is now.

MicahZoltu: I'm not sure if we care, but this could be done with fewer storage updates with something like:
```
uint256 newMigrations = _token.getTotalMigrated() - migratedToSibling[_token];
migratedToSibling[_token] += newMigrations;
totalTheoreticalSupply -= newMigrations;
```

source/contracts/reporting/ReputationToken.sol
```diff
@@ -108,6 +116,36 @@ contract ReputationToken is DelegationTarget, Extractable, ITyped, Initializable
         return universe;
     }
 
+    function getTotalMigrated() public view returns (uint256) {
+        return totalMigrated;
+    }
+
+    function updateSiblingMigrationTotal(IReputationToken _token) public returns (bool) {
+        require(_token != this);
+        IUniverse _supposedUniverse = _token.getUniverse();
```
MicahZoltu: Any particular reason you use `supposed` rather than `shady` here, in contrast to the usage of `shady` elsewhere for the same concept?  I'm not opposed to `supposed` over `shady`, though I do value consistency.

nuevoalex: Not really, that's a good point. Will change in my current PR

PR #515, opened 2017-12-28
Low medium issues

nuevoalex:
Assorted low -> medium issues from the zeppelin audit report


source/contracts/Augur.sol
```diff
@@ -279,6 +280,11 @@ contract Augur is Controlled, Extractable {
         return true;
     }
 
+    function logContractAddedToWhitelist(address _addition) public onlyControllerCaller returns (bool) {
+        WhitelistAddition(_addition);
+        return true;
+    }
```
MicahZoltu: While we are doing this, we should probably have a log for things getting added to the lookup mapping.

nuevoalex: Sounds reasonable. Will add

source/contracts/libraries/Delegator.sol
```diff
@@ -24,13 +24,7 @@ contract Delegator is DelegationTarget {
...
             let _calldataMemoryOffset := mload(0x40)
             // Update the pointer at 0x40 to point at new free memory location so any theoretical allocation doesn't stomp our memory in this call
-            let _size := 0
-            switch gt(calldatasize, 32)
-            case 1 {
-                _size := calldatasize
-            } default {
-                _size := 32
-            }
+            let _size := and(add(calldatasize, 0x1f), not(0x1f))
```
MicahZoltu: Wat?  `not(0x1f)` is a constant isn't it?  Or does `not` take in a memory address?  What is at memory address `0x1f` in that case?  Either way, it feels like we should have a comment indicating what the magic `0x1f` is.  Same question for `add`, does it take in two memory addresses and adds them together?  If so, does `and` then behave different from those two and instead takes a value directly (returned from `add`/`not`) or do `add`/`not` return memory addresses.  If so, do they use the 0x40 pointer to get a memory address in which case we need to consider that our pointer is moving while we do this?

source/contracts/libraries/token/BasicToken.sol
```diff
@@ -1,3 +1,28 @@
...
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
+
```
MicahZoltu: I would _much_ rather include this license in the `root/LICENSE` file and reference which code we picked up from them rather than apply it to each file.  Also, we have made modifications to it and our changes are GPL, so I feel like putting the license in the file incorrectly implies that it is MIT licensed.

source/contracts/reporting/Universe.sol
```diff
@@ -118,7 +118,7 @@ contract Universe is DelegationTarget, Extractable, ITyped, Initializable, IUniv
     }
 
     function getFeeWindowId(uint256 _timestamp) public view returns (uint256) {
-        return _timestamp / getDisputeRoundDurationInSeconds();
+        return _timestamp.div(getDisputeRoundDurationInSeconds());
```
MicahZoltu: Using these functions is more expensive than the native math.  In a lot of these cases, we can easily assert that the math is safe with a comment (which we do in a couple places elsewhere).  While none of these changes are incorrect, I do think we should make sure we are asking if they are worth the increased gas costs or not.

nuevoalex: The cases where I added safe math checks involve user input so i believe it is appropriate to do this as a validation measure.

source/contracts/libraries/token/BasicToken.sol
nuevoalex: I'm not well versed in what exactly is required for licensing. jamming a different license into our license and noting which files it applies to seems odd, although if there is precedence for this as a thing I agree that'd be nicer.

source/contracts/libraries/Delegator.sol
nuevoalex: `0x1f` is a constant. This is the padding value suggested by the audit to use.

`not`,`and`, and `add` in this line are simply modifying the size of memory we are reserving. They are not intended to be memory addresses.

source/contracts/reporting/Universe.sol
MicahZoltu: I have seen that strategy followed before (license file has multiple licenses and what they apply to), which is why I mentioned it.  The most common strategy (by a large margin) is to do what we did before and simply not give attribution when something is licensed as "open source".  😄

My general argument is simply that we should _not_ be littering code with license files.  How we achieve that is less important to me, though I do like encapsulating all of the licensing theater into a single file in the root of the project that everyone can ignore.

source/contracts/libraries/Delegator.sol
nuevoalex: I'll note in a comment that we are including padding in the size calculation

nuevoalex:
Changes done

MicahZoltu:
I'm intentionally deferring PR approval to @bthaile in preparation for my departure.  If you are blocked, let me know and I can approve.

PR #517, opened 2017-12-29
Deployment Issue Fixes

pgebheim:


support/deploy/run.sh
```diff
@@ -38,7 +38,7 @@ case $2 in
   "aura")
     host="aura.ethereum.nodes.augur.net"
     port="80"
-    privateKey="fae42052f82bed612a724fec3632f325f377120592c75bb78adfcceae6470c5a"
+    privateKey="47f49c399482f73143cadeb2db8938d3f249578bdc64cdcda4ecf1ee535a5c91"
```
MicahZoltu: Why are we changing the Aura private key?  The private key currently used by Auru is the same one used by Clique and Instantseal, and it is the one that the node should have setup to be seeded with all of the money.

tinybike: Currently the clique key is different and does not have ether.  It is actually the clique key that should be replaced with the aura key, not the other way around!

pgebheim: Actually it turns out, @tinybike that that key file is wrong and it should be the one located in the "keys" subdirectory, and clique/aura have the same key.

source/deployment/deployContracts.ts
```diff
@@ -19,9 +18,13 @@ async function doWork() {
     await contractDeployer.deploy();
 }
 
-doWork().then(() => {
-    process.exit();
-}).catch(error => {
-    console.log(error);
-    process.exit();
-});
+if (require.main === module) {
```
MicahZoltu: The original intent with files in the `deployment` folder was that they would never be included as a library, they were meant to be "dumb wrappers" around library code that did all of the work.  Wrapping in `if (require.main === module)` implies that this may be used as a library.

My recommendation would be that if you find the business logic in this file to be valuable, then extract it into a library and leave this as a "dumb wrapper".  In this case, it looks like the thing you want is the composition of the above 6 lines, perhaps we could just add a helper function to `ContractDeployer` that does this composition for you?

Alternatively, if you think the pattern of separating library code from script code is stupid, feel free to do whatever you want since I no longer own anything.  😉

source/deployment/deployNetworks.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export async function deployToNetworks(networks: Array<string>) {
+    for(let network of networks) {
+        const configuration = Configuration.network(network);
+        await deployContracts(configuration);
+    }
+}
+
+if (require.main === module) {
+    const networks: Array<string> = process.argv.slice(2);
```
MicahZoltu: I _personally_ prefer configuration via environment rather than via CLI arguments.  As mentioned above though, I have no say in anything anymore so do what feels most right to you.  I do recommend that if we are going to go the route of not using environment variable configuration, then we should go all-out and move all configuration to CLI rather than having a mix of CLI argument and environment variable configuration.

source/libraries/NetworkConfigurations.ts
```diff
@@ -0,0 +1,46 @@
...
+    },
+    kovan: {
+        host: "kovan.augur.net",
+        port: 8545,
+        privateKey: process.env.ROPSTEN_PRIVATE_KEY,
+        gasPrice: new BN(1)
+    },
+    rinkeby: {
+        host: "rinkeby.ethereum.origin.augur.net",
+        port: 8545,
```
MicahZoltu: I believe this should be port `80`.  8545 might still work, but HAP is setup to route port 80 and the others use port 80 below so we should remain consistent.  Same comment for two above.

source/libraries/NetworkConfigurations.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export const networkConfigurations: NetworkConfigurations = {
+    ropsten: {
+        host: "ropsten.augur.net",
+        port: 8545,
+        privateKey: process.env.ROPSTEN_PRIVATE_KEY,
+        gasPrice: new BN(20)
+    },
+    kovan: {
+        host: "kovan.augur.net",
```
MicahZoltu: Are we waiting on something to move Kovan and Ropsten over to the new subdomains?  I think they may already exist in CF but probably point at HAP which doesn't know about these two...

support/deploy/deploy.sh
```diff
@@ -1,10 +1,60 @@
-if [[ "${DEPLOY}" == "true" ]]; then
-  node output/deployment/deployContracts.js || exit 1
+echo "Deploying Augur to $ETHEREUM_NETWORK"
```
MicahZoltu: Hmm, I thought the point of this PR was to get rid of these shell scripts, rather than add to them?

source/deployment/deployNetworks.ts
pgebheim: The idea is that we need *some* tool that has a reasonable human interface. Configuration of the guts via environment is fine, but we've shown that the human interface for that sucks and is confusing / hard to discover.

source/libraries/NetworkConfigurations.ts
pgebheim: We're currently pointing to the origin nodes specifically because of HAP issues. When those are worked out I can switch them back.

source/libraries/NetworkConfigurations.ts
pgebheim: I just needed to check with @epheph if they are good to switch.

support/deploy/deploy.sh
pgebheim: We need this one entrypoint shell script to dispatch into Docker or Local deployment. Past that the goal will be to move everything into TS.

Reasoning being, if it was in TS then you'd need to have the TS built in order to run the code that would invoke the docker image -- thereby requiring a working dev env, which defeats the purpose of the Dockerizedness.

source/deployment/deployContracts.ts
pgebheim: It may make sense to move that into ContractDeployer... I'll take a look.

PR #526, opened 2018-01-11
Additions to testing & a fix

nuevoalex:
More tests and fixes the compilation/deployment issue where Maps are returning garbage.

### DIFF
--- a/source/tests-integration/TestFixture.ts
+++ b/source/tests-integration/TestFixture.ts
@@ -5,7 +5,7 @@ import { Connector } from '../libraries/Connector';
 import { AccountManager } from '../libraries/AccountManager';
 import { ContractCompiler } from '../libraries/ContractCompiler';
 import { ContractDeployer } from '../libraries/ContractDeployer';
-import { LegacyReputationToken, Cash, Universe, ReputationToken, Market, CreateOrder, Orders } from '../libraries/ContractInterfaces';
+import { FeeWindow, ShareToken, CompleteSets, TimeControlled, LegacyReputationToken, Cash, Universe, ReputationToken, Market, CreateOrder, Orders, Trade, CancelOrder } from '../libraries/ContractInterfaces';
 import { stringTo32ByteHex } from '../libraries/HelperFunctions';
 
 export class TestFixture {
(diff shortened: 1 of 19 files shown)


source/tests-integration/TradeAndReport.ts
```diff
@@ -11,21 +12,62 @@ describe("TradeAndReport", () => {
...
+        // Cancel the original rest of order
+        await fixture.cancelOrder(orderID);
+        const remainingAmount = await fixture.getOrderAmount(orderID);
+        expect(remainingAmount.toNumber()).to.equal(0);
+
+        // Proceed to reporting
+        let newTimestamp = await market.getEndTime_();
+        newTimestamp = newTimestamp.add(new BN(1));
+        await fixture.setTimestamp(newTimestamp);
+        let timestamp = await fixture.getTimestamp();
```
adrake33: Should this be a constant?

source/tests-integration/TradeAndReport.ts
```diff
@@ -11,21 +12,62 @@ describe("TradeAndReport", () => {
...
-        const outcome = new BN(0);
-        const numShares = new BN(10000000000000);
-        const price = new BN(2150);
+        const numTicks = await market.getNumTicks_();
+
+        // Place an order
+        let type = new BN(0); // BID
+        let outcome = new BN(0);
+        let numShares = new BN(10000000000000);
+        let price = new BN(2150);
```
adrake33: Are these variables getting modified anywhere? Why are they no longer declared as constants?

source/tests-integration/TestFixture.ts
```diff
@@ -92,6 +92,43 @@ export class TestFixture {
...
+
+        let actualPrice = price;
+        if (type == new BN(1)) {
+            const market = new Market(this.connector, this.accountManager, marketAddress, this.configuration.gasPrice);
+            const numTicks = await market.getNumTicks_();
+            actualPrice = numTicks.sub(price);
+        }
+        const ethValue = numShares.mul(actualPrice);
+
+        const bestPriceAmount = await trade.publicTakeBestOrder_(type, marketAddress, outcome, numShares, price, tradeGroupID, { attachedEth: ethValue });
```
adrake33: Why does this function call both `publicTakeBestOrder_ ` and `publicTakeBestOrder`?

nuevoalex: to get the return value. At present the EVm only support return values from eth calls, not transactions

source/tests-integration/TradeAndReport.ts
nuevoalex: At some point I intend to add more trading in here in which case they will be modified

PR #527, opened 2018-01-12
Adjust fork threshold

nuevoalex:


tests/reporting/test_fee_distribution.py
```diff
@@ -278,7 +278,7 @@ def test_multiple_contributors_crowdsourcer_fees(localFixture, universe, market,
         with EtherDelta(expectedFees / 2, tester.a1, localFixture.chain, "Redeeming didn't increase ETH correctly"):
             assert marketDisputeCrowdsourcer.redeem(tester.a1)
 
-    with TokenDelta(reputationToken, expectedRep, tester.a2, "Redeeming didn't refund REP"):
+    with TokenDelta(reputationToken, expectedRep + 1, tester.a2, "Redeeming didn't refund REP"):
```
adrake33: Why was the "+ 1" added here?

bthaile: Good question. Is it b/c of difference in python math and on contract math?

nuevoalex: Rounding errors / dust.

PR #528, opened 2018-01-12
Feature/multi network deploy

pgebheim:
New PR to identify what this really does.

These changes allow for multiple network deploys, with configuration management all handled in Typescript.

Artifact publishing is still in bash scripts.

### DIFF
--- a/source/libraries/ContractDeployer.ts
+++ b/source/libraries/ContractDeployer.ts
@@ -1,7 +1,7 @@
 import { hash } from 'crypto-promise';
 import { Repository } from 'nodegit';
 import { resolve as resolvePath } from 'path';
-import { writeFile } from "async-file";
+import { exists, readFile, writeFile } from "async-file";
 import { encodeParams } from 'ethjs-abi';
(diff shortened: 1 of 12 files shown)


source/libraries/ContractDeployer.ts
```diff
@@ -227,7 +227,10 @@ export class ContractDeployer {
     }
 
     private async generateAddressMapping(): Promise<string> {
-        const mapping: { [name: string]: string } = {};
+        type ContractAddressMapping = { [name: string]: string };
+        type NetworkAddressMapping = { [networkId: string]: ContractAddressMapping };
```
nuevoalex: Any preference on potentially putting these right after the imports? Unless the type is dynamic somehow that feels more right to me than in the function

pgebheim: I am keeping these local to this function only because they're only used within the function to generate some JSON.

I'm comfortable moving them if its a big concern, but it feels like this whole function could move and bring its internal types with it.

PR #529, opened 2018-01-13
simplify contract API and shimmy in a stopgap fix for geth spam problems

nuevoalex:


source/libraries/ContractDeployer.ts
```diff
@@ -173,7 +172,7 @@ export class ContractDeployer {
 
     private async whitelistTradingContracts(): Promise<void> {
         console.log('Whitelisting contracts...');
-        const promises: Array<Promise<TransactionReceipt>> = [];
+        const promises: Array<Promise<any>> = [];
```
bthaile: Reason for ‘any’, like lean toward strongly typed.

PR #530, opened 2018-01-14
get rid of stopgap retry and replace with deployment serialization stopgap

nuevoalex:


source/libraries/ContractInterfaces.ts
```diff
@@ -39,23 +39,12 @@
...
-                    }
-                    return;
-                } catch (e) {
-                    console.log(`Retrying tx. Got error: ${e}`);
-                    err = e;
-                    continue;
-                }
+            const transactionHash = await this.connector.ethjsQuery.sendRawTransaction(signedTransaction);
+            const txReceipt = await this.connector.waitForTransactionReceipt(transactionHash, `Waiting on receipt for tx: ${txName}`);
+            if (txReceipt.status != 1) {
```
MicahZoltu: Is `txReceipt.status` a number?  I would have expected it to be a `string` or a `BN`.

source/libraries/ContractInterfaces.ts
```diff
@@ -39,23 +39,12 @@
...
+                throw new Error(`Tx ${txName} failed: ${txReceipt}`);
             }
-            throw err;
+            return;
```
MicahZoltu: Can delete this line, it is the last line of the function thus `return` is implicit.

source/libraries/ContractInterfaces.ts
MicahZoltu: Note: this comment and the one below should actually be on the contract generator.

nuevoalex: technically it's a `string | number`. In practice this works as expected

PR #531, opened 2018-01-16
fudge gas estimate to prevent incorrect lower estimates from geth making stuff fail

nuevoalex:


source/libraries/ContractInterfaces.ts
```diff
@@ -903,6 +903,27 @@
     }
 
 
+    export class GethContract extends Contract {
```
MicahZoltu: Feels like this should not be included (just for testing)?

source/libraries/ContractDeployer.ts
```diff
@@ -34,7 +34,7 @@ export class ContractDeployer {
         await this.uploadAllContracts();
         await this.initializeAllContracts();
         await this.whitelistTradingContracts();
-
+
```
MicahZoltu: Surprised your editor let you add this indent, I believe the editorconfig settings are "no whitespace on blank lines"?

source/libraries/ContractInterfaces.ts
nuevoalex: Ah this shouldn't be there.

source/libraries/ContractInterfaces.ts
```diff
@@ -34,8 +34,9 @@
         protected async remoteCall(abi: AbiFunction, parameters: Array<any>, txName: String, sender?: string, gasPrice?: BN, attachedEth?: BN): Promise<void> {
             const from = sender || this.accountManager.defaultAddress;
             const data = encodeMethod(abi, parameters);
-            // TODO: remove `gas` property once https://github.com/ethereumjs/testrpc/issues/411 is fixed
-            const gas = await this.connector.ethjsQuery.estimateGas(Object.assign({ to: this.address, from: from, data: data }, attachedEth ? { value: attachedEth } : {} ));
+            let gas = await this.connector.ethjsQuery.estimateGas(Object.assign({ to: this.address, from: from, data: data }, attachedEth ? { value: attachedEth } : {} ));
+            // This is to address an observed bug while running against geth where occasionally gas estimates are lower than required
```
MicahZoltu: Once we have a repro case in a Geth GitHub issue, we should link to it in this comment so a future reader can know whether it is safe to delete this terrible code.

source/libraries/ContractDeployer.ts
nuevoalex: I'll add something

PR #532, opened 2018-01-17
adding generation of upload-block-numbers.json

epheph:


source/libraries/ContractDeployer.ts
```diff
@@ -28,7 +28,13 @@ export class ContractDeployer {
         this.contracts = new Contracts(compilerOutput);
     }
 
+    public async getBlockNumber(): Promise<number> {
+        return this.connector.ethjsQuery.getBlockByNumber('latest', false).then( (block) => block.number.toNumber());
+    }
+
     public async deploy(): Promise<void> {
+        console.log("Saving contract start block");
+        const blockNumber = await this.getBlockNumber();
```
MicahZoltu: It would be more accurate to get the block number from the controller upload transaction receipt, rather than here.  This block number will actually always be off by at _least_ 1, since you fetch it before starting the upload.

source/libraries/ContractDeployer.ts
```diff
@@ -28,7 +28,13 @@ export class ContractDeployer {
         this.contracts = new Contracts(compilerOutput);
     }
 
+    public async getBlockNumber(): Promise<number> {
+        return this.connector.ethjsQuery.getBlockByNumber('latest', false).then( (block) => block.number.toNumber());
+    }
+
     public async deploy(): Promise<void> {
+        console.log("Saving contract start block");
```
MicahZoltu: Recommend removing this log line.  If it is desirable to know when `deploy` starts, then text should be changed to `Starting contract deployment.` or something (the start block thing is fairly minor relative to the rest of the process).

source/libraries/ContractDeployer.ts
epheph: For posterity, from Discord chat: This is just a hint for log loaders to know when to start, so a block or 2 (or 100) behind isn't a big deal, and the implementation was a little simpler.

PR #536, opened 2018-01-19
add logging for timestamp modification

nuevoalex:


bthaile:
Not critical, we can modify MockAugur as we need to, your test has coverage

PR #538, opened 2018-01-22
Feature/run canned markets

pgebheim:
Run canned markets after multi-network deploy, and don't rely on augur-contracts at all

### DIFF
--- a/source/libraries/ContractInterfacesGenerator.ts
+++ b/source/libraries/ContractInterfacesGenerator.ts
@@ -74,7 +74,7 @@ export class ContractInterfaceGenerator {
             const data = encodeMethod(abi, parameters);
             let gas = await this.connector.ethjsQuery.estimateGas(Object.assign({ to: this.address, from: from, data: data }, attachedEth ? { value: attachedEth } : {} ));
             // This is to address an observed bug while running against geth where occasionally gas estimates are lower than required: https://github.com/ethereum/go-ethereum/issues/15896
-            gas = gas.add(gas.div(new BN(10)));
+            gas = BN.min(new BN(6950000), gas.add(gas.div(new BN(10))));
             gasPrice = gasPrice || this.defaultGasPrice;
             const transaction = Object.assign({ from: from, to: this.address, data: data, gasPrice: gasPrice, gas: gas }, attachedEth ? { value: attachedEth } : {});
             const signedTransaction = await this.accountManager.signTransaction(transaction);
--- a/source/libraries/NetworkConfigurations.ts
+++ b/source/libraries/NetworkConfigurations.ts
@@ -32,7 +32,7 @@ export const networkConfigurations: NetworkConfigurations = {
         host: "rinkeby.ethereum.nodes.augur.net",
         port: 80,
         privateKey: process.env.RINKEBY_PRIVATE_KEY,
-        gasPrice: new BN(30*1000000)
+        gasPrice: new BN(31*1000000)
     },
     clique: {
         isProduction: false,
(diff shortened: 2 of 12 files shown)


support/artifacts/run.sh
```diff
@@ -50,33 +49,19 @@ else
...
-
 ###
 # Update the augur.js repository
 
 git clone $augurjs_repo_url output/augur.js
 current_dir=$PWD
 cd output/augur.js
 
-AUTOCOMMIT=true BRANCH=$branch COMMIT=$commit TAG=$tag npm run update-contracts
+echo "!!!!!!! TEMPORARY !!!!!!!!!!!!!!!!!!"
```
pgebheim: Gotta remove this

source/libraries/ContractInterfacesGenerator.ts
```diff
@@ -74,7 +74,7 @@ export class ContractInterfaceGenerator {
             const data = encodeMethod(abi, parameters);
             let gas = await this.connector.ethjsQuery.estimateGas(Object.assign({ to: this.address, from: from, data: data }, attachedEth ? { value: attachedEth } : {} ));
             // This is to address an observed bug while running against geth where occasionally gas estimates are lower than required: https://github.com/ethereum/go-ethereum/issues/15896
-            gas = gas.add(gas.div(new BN(10)));
+            gas = BN.min(new BN(6950000), gas.add(gas.div(new BN(10))));
```
epheph: Minor: might be nice if this was a constant, perhaps it could come out of the the configuration object, since it's already involved with the network and gasPrice.

source/libraries/NetworkConfigurations.ts
```diff
@@ -18,7 +18,7 @@ export const networkConfigurations: NetworkConfigurations = {
         host: "ropsten.ethereum.nodes.augur.net",
         port: 80,
         privateKey: process.env.ROPSTEN_PRIVATE_KEY,
-        gasPrice: new BN(20)
```
epheph: should be multiplied by 10^9 instead of 10^6?

support/artifacts/run.sh
```diff
@@ -37,7 +37,6 @@ else
   augurjs_repo_url=${AUGURJS_REPO_URL:-https://github.com/AugurProject/augur.js}
   echo "AUGURJS_REPO_URL=${augurjs_repo_url}, set environment to override"
 
-  rm -rf output/augur-contracts
```
epheph: Good night, sweet prince...

source/libraries/ContractInterfacesGenerator.ts
pgebheim: That's a good idea, I'll do it there so its less magic and more expicit

source/libraries/NetworkConfigurations.ts
pgebheim: Same error as with Ropsten ,good call.

support/artifacts/run.sh
pgebheim: 🤴

source/libraries/NetworkConfigurations.ts
pgebheim: this has been updated in a later patch

source/libraries/NetworkConfigurations.ts
```diff
@@ -18,7 +18,7 @@ export const networkConfigurations: NetworkConfigurations = {
...
+        gasPrice: new BN(20*1000000)
```
pgebheim: See here

source/libraries/NetworkConfigurations.ts
epheph: In a later patch that is forthcoming? I see same issue on master

PR #540, opened 2018-01-23
Deploy to aura clique and rinkeby by default in docker

pgebheim:


pgebheim:
Super dumb PR to just deploy to all endpoints

PR #542, opened 2018-01-23
Modify forking bonus to be awarded anytime before the fork period ends

nuevoalex:


source/contracts/reporting/ReputationToken.sol
```diff
@@ -48,11 +48,14 @@ contract ReputationToken is DelegationTarget, Extractable, ITyped, Initializable
         require(ReputationToken(msg.sender) == _parentUniverse.getReputationToken());
         mint(_reporter, _attotokens);
         totalMigrated += _attotokens;
-        // Award a bonus if migration is done before the fork has resolved and update the universe tentative winner tracking
-        if (!_parentUniverse.getForkingMarket().isFinalized()) {
+        // Award a bonus if migration is done before the fork period is over, even if it has finalized
+        if (controller.getTimestamp() < _parentUniverse.getForkEndTime()) {
             uint256 _bonus = _attotokens.div(Reporting.getForkMigrationPercentageBonusDivisor());
```
bthaile: Is this the 5% bonus?

tests/reporting_utils.py
```diff
@@ -96,26 +96,36 @@ def finalizeFork(fixture, market, universe, finalizeByMigration = True):
 
     if (finalizeByMigration):
         # Tester 0 moves more than 50% of REP
```
bthaile: Is there a test for users to migrate after fork? They should not get the bonus.

source/contracts/reporting/ReputationToken.sol
nuevoalex: Yup!

tests/reporting_utils.py
nuevoalex: Yea! That's one of the tests I added in this. Note that if finalizeByMigration isn't true when finalizing a fork it will not add the bonus in the final migration balance check

PR #546, opened 2018-01-26
test upgrading contract

nuevoalex:
Felt like adding an explicit test for upgrading contracts since one didn't seem to be in yet.

### DIFF
--- a/tests/test_upgrading.py
+++ b/tests/test_upgrading.py
@@ -1,0 +1,44 @@
+#!/usr/bin/env python
+
+from ethereum.tools import tester
+from ethereum.tools.tester import TransactionFailed
+from pytest import raises, fixture as pytest_fixture
+from utils import garbageBytes20, garbageBytes32, stringToBytes
+
+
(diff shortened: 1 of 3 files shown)


tests/test_upgrading.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env python
+
+from ethereum.tools import tester
+from ethereum.tools.tester import TransactionFailed
+from pytest import raises, fixture as pytest_fixture
+from utils import garbageBytes20, garbageBytes32, stringToBytes
+
+
+def test_contract_upgrading(localFixture, universe, controller):
```
bthaile: where is universe coming from, I would expect to see a universe pytest_fixture

nuevoalex: Its in `conftest.py`. There are a lot we pre-make available when using the Kitchen Sink

bthaile: makes sense. looks good

PR #550, opened 2018-02-02
add in invariant assertions and update gas cost estimates

nuevoalex:
Leaving these assertions in does increase gas costs as expected, although probably not enough to take them out. I updated the gas estimate tests with the new numbers.

Some logic and tracking had to be added to keep track of the total amount escrowed in the market for open orders but otherwise this is largely just assertions on existing data.


nuevoalex:
Putting this on hold for a bit till there's a larger discussion about this. Since we can't safely check for exact parity between the balance and what is expected (since it can be artificially increased by an attacker or just someone being silly) I am not sure of the value of this actually going in the code proper and remaining there. We could check that the balances are simply _greater_ than or equal to the expected amount but that seems wrong to me as it is potentially game-able I am betting.

PR #551, opened 2018-02-03
Make proper augur-core NPM package

pgebheim:
Augur-core is now structured suitably to be able to be depended on as an NPM package.

This version contains both node and browser compatible versions of augur-core -- such that browser targeting tools like browserify and webpack can properly include only the necessary symbols. At this time the *only* symbol that meets that critera is the exported ABI json.

This library is now depended on by augur.js

We also remove auto-deploy support from Travis, but still build all the docker files as we did previously.


source/libraries/CompilerConfiguration.ts
```diff
@@ -0,0 +1,24 @@
...
+
+    public constructor(contractSourceRoot: string, outputRoot: string) {
+        this.contractSourceRoot = contractSourceRoot;
+        this.outputRoot = outputRoot;
+        this.contractInterfacesOutputPath = path.join(contractSourceRoot, '../libraries', 'ContractInterfaces.ts');
+        this.abiOutputPath = path.join(outputRoot, 'abi.json');
+        this.contractOutputPath = path.join(outputRoot, 'contracts.json');
+    }
+
+    public static create(): CompilerConfiguration {
```
MicahZoltu: Recommend removing this method and moving its contents to the constructor.  The reason there was a factory previously was because some of the configuration options were created asynchronously and you can't have an async constructor.

PR #553, opened 2018-02-11
Refactor so there is a method we can call to get the actual value to be gained by calling this

nuevoalex:


nuevoalex:
Its tested implicitly in the escape hatch tests already :]

PR #554, opened 2018-02-14
Add pre-push hook to run lint

justinbarry:


nuevoalex:
Shouldn't be needed anymore since CI runs linting already

PR #555, opened 2018-02-15
Add extra data to token transfers, make specific logs for stake redemption, add log for initial reporter transfer

nuevoalex:


epheph:
Splitting generalized "winning" report into crowdsource and initial. interesting, makes sense. 👍

epheph:
What's the difference between FeeToken and FeeWindow token? FeeWindow seems like it would be the participation token, FeeToken is initial report?

nuevoalex:
FeeWindow tokens are participation tokens yup. FeeTokens are only used internally for accounting purposes so that stake holders can get fees for every fee window they have stake locked up.

PR #559, opened 2018-02-23
Fixes

nuevoalex:


package-lock.json
```diff
@@ -388,7 +388,6 @@
             "requires": {
```
adrake33: Why did `fsevents` and `nan` get removed? We're not using them for anything anymore?

nuevoalex: I think we haven't been using them for some time and an npm install just updated it

PR #563, opened 2018-02-27
uint8 -> uint256

nuevoalex:
While the suggestion was to simply change all for loops to be uint256 this essentially means we have to change everywhere. Based on their assertion that this is essentially wasted computation anyway this is the right thing to do I believe sadly.

### DIFF
--- a/source/contracts/factories/MarketFactory.sol
+++ b/source/contracts/factories/MarketFactory.sol
@@ -9,7 +9,7 @@ import 'IController.sol';
 
 
 contract MarketFactory {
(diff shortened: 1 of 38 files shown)


nuevoalex:
This shouldn't change any downstream consumers

MicahZoltu:
I'm pretty strongly against this change and I believe it does more harm than good.  Using uint8 provided a very strong hint to the reader what to expect in the variable.  The purpose of using it wasn't as a performance gain, but rather so the reader knew what to expect and so the compiler would give us an error if someone tried to assign a big number to uint8.  Now, someone can assign 2^255 to a variable that should never be large and we won't get any compiler errors.

nuevoalex:
That was my initial feeling as well but uint8 still allowed larger numbers than we usually want anyway and we implement explicit checks for that.

PR #567, opened 2018-02-27
Remove the Extractable contract and feature

nuevoalex:


source/contracts/trading/ICash.sol
```diff
@@ -9,5 +9,4 @@ contract ICash is ERC20 {
     function depositEtherFor(address _to) external payable returns(bool);
     function withdrawEther(uint256 _amount) external returns(bool);
     function withdrawEtherTo(address _to, uint256 _amount) external returns(bool);
-    function getTypeName() public view returns (bytes32);
```
adrake33: Why was this removed? Is it no longer needed?

nuevoalex: Cash already inherits from this so it was redundant

PR #568, opened 2018-02-27
Various cleanup

nuevoalex:
Just some small various cleanup issues suggested in the audit. More to come but this seemed like a good PR size


source/contracts/trading/CancelOrder.sol
```diff
@@ -20,7 +19,7 @@ import 'trading/IOrders.sol';
  * @title CancelOrder
  * @dev This allows you to cancel orders on the book.
  */
-contract CancelOrder is CashAutoConverter, ReentrancyGuard, MarketValidator, ICancelOrder {
+contract CancelOrder is CashAutoConverter, ReentrancyGuard, ICancelOrder {
```
adrake33: Why is MarketValidator being removed here?

nuevoalex: It's never actually used here. Its only included in order to gain access to the market validation modifer which we don't need in this contract.

PR #569, opened 2018-02-28
More cleanup issues

nuevoalex:


source/contracts/trading/Cash.sol
```diff
@@ -30,16 +30,18 @@ contract Cash is DelegationTarget, ITyped, VariableSupplyToken, ICash {
     }
 
     function withdrawEther(uint256 _amount) external returns(bool) {
```
adrake33: Are `withdrawEther` and `withdrawEtherTo` intended to be called publicly? These aren't currently documented, but since users don't interact directly with the Cash contract, this should be ok, correct?

nuevoalex: They can be but its not really intended yea.

PR #571, opened 2018-02-28
update token contracts to include the incremental approval methods

nuevoalex:


source/contracts/libraries/token/StandardToken.sol
```diff
@@ -56,4 +56,41 @@ contract StandardToken is ERC20, BasicToken {
     function allowance(address _owner, address _spender) public view returns (uint256 remaining) {
         return allowed[_owner][_spender];
     }
+
+   /**
+   * @dev Increase the amount of tokens that an owner allowed to a spender.
+   *
+   * approve should be called when allowed[_spender] == 0. To increment
+   * allowed value is better to use this function to avoid 2 calls (and wait until
+   * the first transaction is mined)
```
MicahZoltu: Style: No hard newlines in comments.

source/contracts/libraries/token/StandardToken.sol
```diff
@@ -56,4 +56,41 @@ contract StandardToken is ERC20, BasicToken {
...
+   * From MonolithDAO Token.sol
```
MicahZoltu: Recommend getting this attribution out of code and into our license file(s).  Attribution in-code just clutters the code and if you follow it as a standard practice, you end up with code that is nearly unreadable due to attributions everywhere and you'll probably end up getting some of them wrong.

source/contracts/libraries/token/StandardToken.sol
```diff
@@ -56,4 +56,41 @@ contract StandardToken is ERC20, BasicToken {
...
+        return true;
+    }
+
+  /**
+   * @dev Decrease the amount of tokens that an owner allowed to a spender.
+   *
+   * approve should be called when allowed[_spender] == 0. To decrement
+   * allowed value is better to use this function to avoid 2 calls (and wait until
+   * the first transaction is mined)
+   * From MonolithDAO Token.sol
```
MicahZoltu: Same comments as above.

MicahZoltu:
I have a strong suspicion that no one will ever use these, but adding them doesn't _hurt_ anything other than contract upload size.

PR #578, opened 2018-03-04
implement escape hatch via upgrade

nuevoalex:
This would be a basic implementation of turning the escape hatch on via contract upload instead of a specific function call.

This now means the escape hatch is only pull-able while dev mode is on.


MicahZoltu:
It is unclear to me what problem this solves?

PR #579, opened 2018-03-05
escape hatch through update

nuevoalex:


source/contracts/Controller.sol
```diff
@@ -127,26 +124,11 @@ contract Controller is IController {
         return true;
     }
 
-    function switchModeSoOnlyEmergencyStopsAndEscapeHatchesCanBeUsed() public devModeOwnerOnly returns (bool) {
+    function switchToCentralizedMode() public devModeOwnerOnly returns (bool) {
```
tinybike: Should this be `switchToDecentralizedMode`?  Or, a better name might be something like, `switchOffDevMode`.

source/contracts/Controller.sol
```diff
@@ -127,26 +124,11 @@ contract Controller is IController {
...
         whitelist[owner] = false;
```
tinybike: Should this use `Controller.removeFromWhitelist` instead of editing `whitelist` directly?

nuevoalex: Yea that makes sense. Will change

source/contracts/Controller.sol
```diff
@@ -120,30 +117,15 @@ contract Controller is IController {
     function transferOwnership(address _newOwner) public onlyOwnerCaller returns (bool) {
         // if in dev mode, blacklist old owner and whitelist new owner
         if (whitelist[owner]) {
-            whitelist[owner] = false;
+            removeFromWhitelist(owner);
             whitelist[_newOwner] = true;
```
tinybike: Suggest using `addToWhitelist(_newOwner)` instead of editing `whitelist` directly.

PR #580, opened 2018-03-07
low level calls -> transfer. Wrap ERC20 calls in requires

nuevoalex:


source/contracts/reporting/Mailbox.sol
```diff
@@ -29,7 +29,7 @@ contract Mailbox is DelegationTarget, Ownable, Initializable, IMailbox {
         }
         // Withdraw any ETH balance
         if (this.balance > 0) {
-            require(owner.call.value(this.balance)());
+            owner.transfer(this.balance);
```
tinybike: Should this be wrapped in `require`?

source/contracts/reporting/Market.sol
```diff
@@ -87,7 +87,7 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
         // If the value was not at least equal to the sum of these fees this will throw. The addition here cannot overflow as these fees are capped
         uint256 _refund = msg.value.sub(reporterGasCostsFeeAttoeth + validityBondAttoeth);
         if (_refund > 0) {
-            require(owner.call.value(_refund)());
+            owner.transfer(_refund);
```
tinybike: Should this be wrapped in `require`?

source/contracts/factories/MarketFactory.sol
```diff
@@ -13,7 +13,7 @@ contract MarketFactory {
         Delegator _delegator = new Delegator(_controller, "Market");
         _market = IMarket(_delegator);
         IReputationToken _reputationToken = _universe.getReputationToken();
-        _reputationToken.transfer(_market, _reputationToken.balanceOf(this));
+        require(_reputationToken.transfer(_market, _reputationToken.balanceOf(this)));
```
MicahZoltu: Per comments previously, this is wasteful.  Why do we wrap this one function call in a `require` while every other cross contract function call is not wrapped in a require statement?  What makes this one special?  This goes against our existing coding standards and implies to the user that this contract call is somehow special/different from the others even though it is not.  It also is a waste of gas.

MicahZoltu: (same comment for many other additions in this PR)

nuevoalex: I tend to agree. The argument being made is one of adhering to standards without special consideration since that can expose vulnerabilities from absentnt-mindely not using a requires in a case where an external token is used.

PR #586, opened 2018-03-20
Controller modifications

nuevoalex:
Removed centralized aspects of the platform with the exception of the Controller contract owner


source/contracts/Controller.sol
```diff
@@ -24,48 +24,43 @@ contract Controller is IController {
     address public owner;
     mapping(address => bool) public whitelist;
     mapping(bytes32 => ContractDetails) public registry;
+    bool public stopped = false;
```
MicahZoltu: I prefer this way, but I thought the coding style required private variables with public getters?

source/contracts/Controller.sol
```diff
@@ -98,32 +90,18 @@ contract Controller is IController {
...
-        }
-        owner = _newOwner;
+    function emergencyStop() public onlyOwnerCaller onlyInGoodTimes returns (bool) {
+        stopped = true;
         return true;
     }
 
-    function switchOffDevMode() public devModeOwnerOnly returns (bool) {
-        whitelist[owner] = false;
+    function release() public onlyOwnerCaller onlyInBadTimes returns (bool) {
```
MicahZoltu: Not a huge fan of this name, though I don't have any great recommendations.  Would be nice to come up with a name that is somehow a mirror of `emergencyStop`.  Perhaps just renaming both to `start` and `stop`?  Or `emergencyStop` and `emergencyStart`?  I recognize the problems with both of these, but at least they are more clearly paired.

source/contracts/reporting/Universe.sol
```diff
@@ -384,31 +384,9 @@ contract Universe is DelegationTarget, ITyped, Initializable, IUniverse {
...
-
-        _currentFeeDivisor = _currentFeeDivisor
-            .max(Reporting.getMinimumReportingFeeDivisor())
-            .min(Reporting.getMaximumReportingFeeDivisor());
-
-        shareSettlementFeeDivisor[_feeWindow] = _currentFeeDivisor;
-        return _currentFeeDivisor;
+    // Hardcode the reporting fee until we implement the auction feature or some other decentralized REP price oracle
+    function getOrCacheReportingFeeDivisor() public view returns (uint256) {
+        return Reporting.getMaximumReportingFeeDivisor();
```
MicahZoltu: I have been thinking about this, I wonder if we should bump the min reporting fee to 0.1% or 1% now that we are hard-coding it?  This will let us run safely for longer before we _need_ to upgrade.

source/contracts/Controller.sol
nuevoalex: This is just a return to how it was previously implemented. It didn't ever follow our coding style but I don't feel strongly about making it conform.

source/contracts/reporting/Universe.sol
nuevoalex: This is a good point. The default fee actually starts at 1% in the old code, so maybe we could go with that. Should probably be discussed more in chat

source/contracts/Controller.sol
nuevoalex: I'm super ambivalent about this point (despite generally agreeing) especially since it will likely only ever be used by me in development and shouldn't ever be called outside of that context.

PR #588, opened 2018-03-21
add script for oyente

nuevoalex:
Really basic initial script to get oyente output. There is a stub that just prints json where we can do CI stuff later if desired. Going to link to a gist with the full run output in the issue to do an oyente scan separately.

### DIFF
--- a/source/tools/runOyente.py
+++ b/source/tools/runOyente.py
@@ -1,0 +1,74 @@
+from oyente.oyente import run_solidity_analysis
+from oyente.input_helper import InputHelper
+from oyente.source_map import SourceMap
+from os import path, walk, makedirs, listdir
+from pprint import pprint
+
+import argparse
+import json
+import logging
+import sys
+
+BASE_PATH = path.dirname(path.abspath(__file__))
+def resolveRelativePath(relativeFilePath):
+    return path.abspath(path.join(BASE_PATH, relativeFilePath))
+
+def main():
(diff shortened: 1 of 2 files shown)


source/tools/runOyente.py
```diff
@@ -0,0 +1,74 @@
+from oyente.oyente import run_solidity_analysis
+from oyente.input_helper import InputHelper
+from oyente.source_map import SourceMap
+from os import path, walk, makedirs, listdir
+from pprint import pprint
```
adrake33: makedirs, listdir, and pprint don't appear to be getting used.

source/tools/runOyente.py
```diff
@@ -0,0 +1,74 @@
...
+
+import argparse
+import json
```
adrake33: Same goes for json.

source/tools/runOyente.py
```diff
@@ -0,0 +1,74 @@
...
+import sys
+
+BASE_PATH = path.dirname(path.abspath(__file__))
+def resolveRelativePath(relativeFilePath):
+    return path.abspath(path.join(BASE_PATH, relativeFilePath))
+
+def main():
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument("-a", "--analyze", help="Do some sort of programattic analysis. Right now nothing interesting. Later!", action="store_true")
```
adrake33: Small typo: programattic.

PR #590, opened 2018-03-26
add log for when a market migrates

nuevoalex:
Add a log for when a market migrates


source/contracts/Augur.sol
```diff
@@ -34,6 +34,7 @@ contract Augur is Controlled, IAugur {
     event DisputeCrowdsourcerRedeemed(address indexed universe, address indexed reporter, address indexed market, address disputeCrowdsourcer, uint256 amountRedeemed, uint256 repReceived, uint256 reportingFeesReceived, uint256[] payoutNumerators);
     event FeeWindowRedeemed(address indexed universe, address indexed reporter, address indexed feeWindow, uint256 amountRedeemed, uint256 reportingFeesReceived);
     event MarketFinalized(address indexed universe, address indexed market);
+    event MarketMigrated(address indexed market, address indexed originalUniverse, address indexed newUniverse);
```
adrake33: Ah yeah, good idea to add the original universe to the log. :)

PR #592, opened 2018-03-27
Only check if we need to upload a contract if we've had an existing

pgebheim:
controller passed in

### DIFF
--- a/source/libraries/ContractDeployer.ts
+++ b/source/libraries/ContractDeployer.ts
@@ -148,7 +148,7 @@ Deploying to: ${networkConfiguration.networkName}
         if (contract.relativeFilePath.startsWith('legacy_reputation/')) return;
         if (contractName !== 'Map' && contract.relativeFilePath.startsWith('libraries/')) return;
         // Check to see if we have already uploded this version of the contract
-        if (await this.shouldUpgradeContract(contract, contractsToDelegate[contractName])) {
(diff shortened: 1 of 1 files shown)


source/libraries/ContractDeployer.ts
```diff
@@ -148,7 +148,7 @@ Deploying to: ${networkConfiguration.networkName}
         if (contract.relativeFilePath.startsWith('legacy_reputation/')) return;
         if (contractName !== 'Map' && contract.relativeFilePath.startsWith('libraries/')) return;
         // Check to see if we have already uploded this version of the contract
-        if (await this.shouldUpgradeContract(contract, contractsToDelegate[contractName])) {
+        if (typeof this.configuration.controllerAddress !== "undefined" && await this.shouldUpgradeContract(contract, contractsToDelegate[contractName])) {
```
nuevoalex: Not related to this change but it actually looks like this is doing the opposite of what it says if I'm reading correctly. looks like if `shouldUpgradeContract` returns true it tries to reuse the existing contract. otherwise looks fine though

pgebheim: That's the intended behavior.

"Upgrade" means "only upload contracts that have changed", or else we just upload everything.

pgebheim: Totally. That is a terribly named function!

PR #593, opened 2018-03-29
Create .gitattributes

joeykrug:


MicahZoltu:
In my experience this doesn't actually work.  😢 I tried to set it up on my personal repos with no luck.

joeykrug:
It works for me!

PR #594, opened 2018-03-30
make logging of universe creation occur for genesis creation as well

nuevoalex:


epheph:
@nuevoalex Why is parent payout is logged with the child universe?

nuevoalex:
We use the parent payout in AN to correlate a child universe with the payout that it represents for its parent.This is used currently in getting fork REP migration totals.

epheph:
But we already emit the parentUniverse address, that's enough for the correlation? Where is the data available for the payout that universe represents?

PR #598, opened 2018-04-02
0.13.0

epheph:


epheph:
For bounty, 0.13.0 minor version

nuevoalex:
Gonna close this actually since we're bumping now to a higher ver

epheph:
@nuevoalex yes, this is stale (I can't direct push to master, but making PR takes a long time before tests pass!). It's fine, we have the v0.13.0 tag, so if a version wasn't created that conflicted with it, we all good

PR #599, opened 2018-04-03
Auto rep migration support

nuevoalex:
When a genesis universe is created it will assume the legacy REP contract is frozen and until all balances are migrated it will be frozen except in migrating balances and allowances.

I added a test net rep contract that we can use so we still have faucet capabilities on test nets.

### DIFF
--- a/source/libraries/ContractDeployer.ts
+++ b/source/libraries/ContractDeployer.ts
@@ -1,3 +1,4 @@
+import BN = require('bn.js');
 import { hash } from 'crypto-promise';
 import { exists, readFile, writeFile } from "async-file";
 import { exec } from 'child_process';
@@ -8,7 +9,7 @@ import { CompilerOutput } from "solc";
 import { Abi, AbiFunction } from 'ethereum';
 import { DeployerConfiguration } from './DeployerConfiguration';
 import { Connector } from './Connector';
-import { Augur, ContractFactory, Controller, Controlled, Universe } from './ContractInterfaces';
+import { Augur, ContractFactory, Controller, Controlled, Universe, ReputationToken, LegacyReputationToken } from './ContractInterfaces';
 import { NetworkConfiguration } from './NetworkConfiguration';
 import { AccountManager } from './AccountManager';
 import { Contracts, Contract } from './Contracts';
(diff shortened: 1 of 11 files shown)


source/libraries/ContractDeployer.ts
```diff
@@ -57,7 +58,15 @@ Deploying to: ${networkConfiguration.networkName}
...
 
         if(this.configuration.createGenesisUniverse) {
+            if (!this.configuration.isProduction) {
+                this.initializeLegacyRep();
+            }
+
             this.universe = await this.createGenesisUniverse();
+
+            if (!this.configuration.isProduction) {
+                this.migrateFromlegacyRep();
```
adrake33: I'm not sure I understand how this is intended to work on production. I thought the idea was that holders of legacy REP would all be migrated to regular REP automatically.  How will this be done?

source/libraries/ContractDeployer.ts
```diff
@@ -258,9 +278,24 @@ Deploying to: ${networkConfiguration.networkName}
         if (await universe.getTypeName_() !== stringTo32ByteHex("Universe")) {
             throw new Error("Unable to create genesis universe. Get type name failed");
         }
+
         return universe;
     }
 
+    private async migrateFromlegacyRep(): Promise<void> {
```
adrake33: Nit: shouldn't the 'l' in `migrateFromlegacyRep` be capitalized here?

source/libraries/ContractDeployer.ts
nuevoalex: In a non production environment there is only one Legacy REP holder, the deployer of the contracts. See the part right above thise where in a non production environment we mint 11 million Legacy REP for the default address

source/libraries/ContractDeployer.ts
nuevoalex: Will fix!

source/libraries/ContractDeployer.ts
adrake33: Right, but isn't the intent to auto-migrate legacy REP users on production?  Will they have to call `migrateBalancesFromLegacyRep` and `migrateAllowancesFromLegacyRep` manually, or how will these be called on production?  Just trying to make sure I understand the intended behavior.

nuevoalex: These _won't_ be called in production. The reason we're calling them in dev is so that everything is ready to go for development and on testnets. In production we'll need to run a script that migrates balances and allowances before the system functions

adrake33: Ok, gotcha. That's the part I was trying to clarify.

PR #603, opened 2018-04-10
remove code that can only make gas cost increase

nuevoalex:


nuevoalex:
Since we have the 3 day waiting period requirement right after this it would be impossible for a tx to finalize the market here and then succeed

MicahZoltu:
Hmm, I see.  Since the long-term intent is to delete the 3-day delay, I would recommend adding a comment in that reminds future coder to add back in the check.  Once the 3-day delay is removed, I believe that check is the only thing stopping someone from withdrawing trading proceeds from an un-finalized market.  It _may_ fail due to some other check in the call stack, but I would be very wary of relying on that.

PR #604, opened 2018-04-10
go back to rep price oracle owned input

nuevoalex:


nuevoalex:
Decided it would be best to have a stopgap solution in for the initial release rather than feel rushed to do provide a new set of contracts

PR #606, opened 2018-04-20
Coverage report generation script and docs

nuevoalex:
Made the following changes to tests: 

- All txs send a stupid amount of gas now. If we want to find out how much stuff costs we can do so explicitly still. The coverage contracts cost far higher than the limit though because of all the additional logging so this is the simplest way. Also it was kind of stupid combating this on a case by case basis in the first place.
- All log testing is done with the `AssertLog` method. This will scan the logs for the desired entry and confirm it exists. We have to do this to make tests pass when we're firing hundreds of logs in coverage mode. Also it is just far cleaner than testing precise log index values.

The `Trade.sol` contract change is to accommodate coverage test runs but is functionally the same but with a trivially higher gas cost due to replacing a constant with a function call. When running in coverage mode the gas cost of the function is much higher so the min gas needs to be higher as well. We make this change and during python setup we use a subclass which uses a much higher min gas value.

Everything else should be fairly straightforward!

### DIFF
--- a/source/tools/generateCoverageReport.js
+++ b/source/tools/generateCoverageReport.js
@@ -3,7 +3,6 @@
 // NOTE: Make sure to run with --max-old-space-size=12288. We're loading a massive file into memory during report generation
 
 const App = require('solidity-coverage/lib/app.js');
-const reqCwd = require('req-cwd');
 const death = require('death');
 const { execSync } = require('child_process');
 const copydir = require('copy-dir');
--- a/source/tools/generateCoverageReport.js
+++ b/source/tools/generateCoverageReport.js
@@ -1,0 +1,53 @@
+#!/usr/bin/env node
+
+// NOTE: Make sure to run with --max-old-space-size=12288. We're loading a massive file into memory during report generation
+
+const App = require('solidity-coverage/lib/app.js');
+const reqCwd = require('req-cwd');
+const death = require('death');
+const { execSync } = require('child_process');
+const copydir = require('copy-dir');
+const replace = require("replace");
+const rimraf = require('rimraf');
+const fs = require('fs');
+
+const config = {
+    dir: './source',
+    skipFiles: [],
+    copyNodeModules: false,
+}
+
+const app = new App(config);
+app.postProcessPure = function () {};
+
+death((signal, err) => app.cleanUp(err));
+
+app.generateCoverageEnvironment();
+
+app.instrumentTarget();
+
+rimraf.sync('./coverageEnv/solidity_test_helpers');
+fs.mkdirSync('./coverageEnv/solidity_test_helpers')
+
+copydir.sync('./tests/solidity_test_helpers', './coverageEnv/solidity_test_helpers')
+
+replace({
+    regex: " view | pure ",
+    replacement: " ",
+    paths: fs.readdirSync('./coverageEnv/solidity_test_helpers').map(filename => './coverageEnv/solidity_test_helpers/' + filename),
+    silent: false,
+})
+
+try {
+    execSync('pytest --cover', {stdio:[0,1,2]});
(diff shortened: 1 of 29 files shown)


tests/conftest.py
```diff
@@ -6,15 +6,40 @@
...
 
+from ethereum.slogging import get_logger
+
+# Make TXs free.
+ethereum.opcodes.GCONTRACTBYTE = 0
+ethereum.opcodes.GTXDATAZERO = 0
+ethereum.opcodes.GTXDATANONZERO = 0
+
+# Monkeypatch to bypass the contract size limit
+def monkey_post_spurious_dragon_hardfork():
```
adrake33: Awesome function name. :)

tests/reporting/test_market.py
```diff
@@ -2,23 +2,22 @@
 from ethereum.tools import tester
 from ethereum.tools.tester import TransactionFailed
 from pytest import raises
-from utils import stringToBytes, captureFilteredLogs, bytesToHexString
+from utils import stringToBytes, AssertLog, bytesToHexString, AssertLog
```
adrake33: AssertLog only needs to be imported once here.

tests/trading/test_fillOrder.py
```diff
@@ -154,28 +118,12 @@ def test_fill_order_with_shares_escrowed_sell_with_shares(contractsFixture, cash
...
-    assert log1["filler"] == bytesToHexString(tester.a2)
-    assert log1["numCreatorShares"] == fix(1)
-    assert log1["numCreatorTokens"] == 0
-    assert log1["numFillerShares"] == fix(1)
-    assert log1["numFillerTokens"] == 0
-    assert log1["marketCreatorFees"] == fix(1, 10000) / market.getMarketCreatorSettlementFeeDivisor()
-    assert log1["reporterFees"] == fix(1, 10000) / universe.getOrCacheReportingFeeDivisor()
-    assert log1["shareToken"] == market.getShareToken(YES)
-    assert log1["tradeGroupId"] == stringToBytes("43")
-
```
adrake33: I don't see any new log code to replace this and some of the removed code below. Should all of this be getting removed?

source/tools/generateCoverageReport.js
```diff
@@ -0,0 +1,53 @@
...
+const fs = require('fs');
+
+const config = {
+    dir: './source',
+    skipFiles: [],
+    copyNodeModules: false,
+}
+
+const app = new App(config);
+app.postProcessPure = function () {};
```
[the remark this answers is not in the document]
nuevoalex: Yea I'm overwriting an existing function

tests/conftest.py
```diff
@@ -6,15 +6,40 @@
...
-from os import path, walk, makedirs, listdir
+from json import dump as json_dump, load as json_load, dumps as json_dumps
+from os import path, walk, makedirs, listdir, remove as remove_file
 import pytest
 from re import findall
 from solc import compile_standard
 from utils import bytesToHexString, bytesToLong, longToHexString, stringToBytes, garbageBytes20, garbageBytes32, twentyZeros, thirtyTwoZeros
 from copy import deepcopy
 
+from ethereum.slogging import get_logger
```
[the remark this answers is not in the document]
nuevoalex: Will remove

tests/conftest.py
nuevoalex: haha thanks

tests/trading/test_fillOrder.py
nuevoalex: I'm not sure why it was ever being done this way to begin with since we confirm the on chain results anyway. We only need to confirm logging works once IMO. It's also really expensive to test time wise.

PR #615, opened 2018-04-26
Make setTimestamp a real transaction, not an off-chain call

epheph:


epheph:
It's really easy to miss! 😬

PR #622, opened 2018-05-03
log when markets and market mailboxes are transferred

nuevoalex:


source/contracts/reporting/IMarket.sol
```diff
@@ -36,6 +37,7 @@ contract IMarket is ITyped, IOwnable {
     function getWinningPayoutNumerator(uint256 _outcome) public view returns (uint256);
     function getReputationToken() public view returns (IReputationToken);
     function getFinalizationTime() public view returns (uint256);
+    function getInitialReporterAddress() public view returns (address);
```
nuevoalex: This is to avoid cyclical dependency problems with making this just use getInitialReporter

Issue #609, opened 2018-04-21
Passing Tests w/ Maian


tomkysar:
Submit a PR with tests passing with Maian.

This may require interacting with the project maintainers to get it working 100% with Augur's contracts. We suggest working with them on bug reports, fail cases, etc. Please join the #dev or #bounties channel of our [Discord](http://augur.net/invite) with any questions. 

Bounty posted with [Gitcoin](https://gitcoin.co). 

[1] https://github.com/MAIAN-tool/MAIAN

gitcoinbot:
__This issue now has a funding of 1000.0 DAI (1000.0 USD @ $1.0/DAI) attached to it.__

 * If you would like to work on this issue you can claim it [here](https://gitcoin.co/issue/AugurProject/augur-core/609).
 * If you've completed this issue and want to claim the bounty you can do so [here](https://gitcoin.co/issue/AugurProject/augur-core/609)
 * Questions? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $8364.97 more Funded OSS Work Available at: https://gitcoin.co/explorer

gitcoinbot:
__Work has been started on the 1000.0 DAI (1000.0 USD @ $1.0/DAI) funding by__: 
 1. [@rofiqgilang](https://gitcoin.co/profile/rofiqgilang) 

 __Please work together__ and coordinate delivery of the issue scope. Gitcoin doesn't know enough about everyones skillsets / free time to say who should work on what, but we trust that the community is smart and well-intentioned enough to work together.  As a general rule; if you start work first, youll be at the top of the above list ^^, and should have 'dibs' as long as you follow through. 

 On the above list? Please leave a comment to let the funder (@tomkysar) and the other parties involved what you're working, with respect to this issue and your plans to resolve it.  If you don't leave a comment, the funder may expire your submission at their discretion. 

 * Learn more [on the gitcoin issue page](https://gitcoin.co/issue/AugurProject/augur-core/609)
 * Questions? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $6372.23 more Funded OSS Work Available at: https://gitcoin.co/explorer

vs77bb:
Hi @rofiqgilang, Vivek from Gitcoin here. Would you mind letting us know your intended process here?

rofiqgilang:
https://bitcointalk.org/index.php?topic=3185579.0
bounty process like this

[...]

PR #624, opened 2018-05-03
feature/maian-tests-ci

cryptomental:
This PR solves #609 - running security analysis of Augur contracts on CI against MAIAN tool.

MAIAN tool evaluates each Augur contract against three possible vulnerability types: greedy, prodigal and suicidal.

Full details on the approach are available in the "Finding The Greedy, Prodigal, and Suicidal Contracts at Scale" paper: https://arxiv.org/pdf/1802.06038.pdf

augur-maian-ci Docker image is based on GitHub repo https://github.com/cryptomental/maian-augur-ci and pushed to Docker Hub using automated build: https://hub.docker.com/r/cryptomental/maian-augur-ci/builds/ . The image contains all MAIAN dependencies, the test runnner and Solidity compiler 0.4.20 used for Augur 1.0.0 release.

Augur contracts are compiled with augurproject/augur-code image using Augur compiler libraries and contracts bytecode is extracted by the test runner to bin files for further evaulation. Then the test runner executes MAIAN tool three times for each Augur contract and provides a summary after test run finishes.

Refs: #609

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,8 @@ script: docker pull $IMAGE_BUILD;
         elif [[ "$TESTS" == "security:mythril" ]]; then
         npm run docker:run:test:security:mythril;
         cat source/contracts/test-results.log;
+        elif [[ "$TESTS" == "security:maian" ]]; then
+        npm run docker:run:test:security:maian;
         else
           npm run docker:run:test:unit -- $TESTS;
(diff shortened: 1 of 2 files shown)


joeykrug:
Looks good, the greedy stuff shows a lot of false positives in their paper too so that's as expected

PR #640, opened 2018-05-20
Min gas costs

nuevoalex:
Create Market cost could be reduced a lot but the worst case fill for an 8 outcome cat is still around the level we upped it to :[


epheph:
new commit looks good

Issue #607, opened 2018-04-21
Passing Tests w/ Solidity SMTChecker


tomkysar:
Submit a PR with tests passing with Solidity SMTChecker pragma enabled and in CI.

This may require interacting with the project maintainers to get it working 100% with Augur's contracts. We suggest working with them on bug reports, fail cases, etc. Please join the #dev or #bounties channel of our [Discord](http://augur.net/invite) with any questions. 

Bounty posted with [Gitcoin](https://gitcoin.co). 

[1] https://github.com/ethereum/solidity

gitcoinbot:
__This issue now has a funding of 1000.0 DAI (1000.0 USD @ $1.0/DAI) attached to it.__

 * If you would like to work on this issue you can claim it [here](https://gitcoin.co/issue/AugurProject/augur-core/607).
 * If you've completed this issue and want to claim the bounty you can do so [here](https://gitcoin.co/issue/AugurProject/augur-core/607)
 * Questions? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $6364.97 more Funded OSS Work Available at: https://gitcoin.co/explorer

gitcoinbot:
__Work has been started on the 1000.0 DAI (1000.0 USD @ $1.0/DAI) funding by__: 
 1. [@bakaoh](https://gitcoin.co/profile/bakaoh) 

 __Please work together__ and coordinate delivery of the issue scope. Gitcoin doesn't know enough about everyones skillsets / free time to say who should work on what, but we trust that the community is smart and well-intentioned enough to work together.  As a general rule; if you start work first, youll be at the top of the above list ^^, and should have 'dibs' as long as you follow through. 

 On the above list? Please leave a comment to let the funder (@tomkysar) and the other parties involved what you're working, with respect to this issue and your plans to resolve it.  If you don't leave a comment, the funder may expire your submission at their discretion. 

 * Learn more [on the gitcoin issue page](https://gitcoin.co/issue/AugurProject/augur-core/607)
 * Questions? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $8364.66 more Funded OSS Work Available at: https://gitcoin.co/explorer

vs77bb:
@bakaoh This is Vivek from Gitcoin - great to see you here! Please keep us posted with questions as you give this one a go.

bakaoh:
this issue may take some time to finish,  i'm still working on it

[...]

PR #642, opened 2018-05-21
Enable SMT Checker

cryptomental:
This pull request solves #607 .
    
* SMT checker runs a compiler-time tests and is enabled by adding compiler directive "pragma experimental SMTChecker;" to Solidity contracts.
* To enable SMTChecker, Solidity compiler with SMT support must be used. SMT support is currently available only for Ubuntu and Debian based Linux distributions. Z3 theorem prover from Microsoft is required to run and a dependency libz3-dev must be installed to run SMT checks.
* ContractCompiler.ts is patched to use native solc compiler instead of solc.js and SMT errors and warnings are displayed to the console without stopping the build.
    
Notes:
    
* There is a number of contracts that crashes solc compiler with "Internal error" message. The contracts that cause crashes are disabled for SMT check by putting them into quarantine. The quarantined contracts will be reported to Solidity team to improve SMT checker.
    
* Solc 0.4.20 is not available in PPA due to a glitch in the release. This is mentioned in https://github.com/ethereum/solidity/issues/3593. Due to this, the current Dockerfile uses fork https://github.com/cryptomental/solidity/releases/tag/v0.4.20-SMT. For solc versions newer than v0.4.20, original PPA releases can be used.


vs77bb:
Awesome! @tomkysar Looks like this one might be good for payout to @cryptomental [here](https://gitcoin.co/issue/AugurProject/augur-core/607/283).

Great work, @cryptomental 🙂

cryptomental:
Thank you!

PR #643, opened 2018-05-22
no self trading restriction

nuevoalex:
DO NOT MERGE YET


nuevoalex:
Gonna merge this in once CI finishes so that the branch in augur.js with updates can  reference the latest augur-core with this in it and work correctly.

PR #645, opened 2018-05-24
force cat 3,6,7 into 10000 numticks. Fix OI leak and fork bug

nuevoalex:


source/contracts/reporting/Market.sol
```diff
@@ -470,7 +477,11 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
             require(!_invalid || _value == _previousValue);
```
epheph: Oh i didn't you changed the lines below, but this is the line that i was concerned about

nuevoalex: That should be fine. we still want the payouts to equal. They wont sum to numTicks though in the invalid case.

For the case of a 3 outcome categorical for example all payouts are 3333.

epheph: Oh i see, so we'll just [very very] minorly short! I gotcha now, thx for explanation.

epheph: So this will trap dust amounts of Cash on market, in the case of a specifically non-divisible numerator set that ends up invalid

epheph: This is only a real problem if someone does something crazy like use numTicks of 3 for binary 👍

adrake33:
Closing, since these changes were included in https://github.com/AugurProject/augur-core/pull/646.

PR #646, opened 2018-05-24
Achapman/ch11509/allow a 0 market creator fee

nuevoalex:


epheph:
@nuevoalex what about `derivePayoutDistributionHash` ?

nuevoalex:
It was modified to cater to this case. Do you mean with that modification?

PR #647, opened 2018-05-24
Achapman/ch11508/trade functions with no min gas requirements

nuevoalex:
This is for future proofing so that if gas costs decrease we have some recourse when still doing on chain trading to take advantage of it.

### DIFF
--- a/source/contracts/trading/Trade.sol
+++ b/source/contracts/trading/Trade.sol
@@ -42,7 +42,6 @@ contract Trade is CashAutoConverter, ReentrancyGuard, MarketValidator {
         return _result;
     }
 
-    // CONSIDER: We may want to return multiple values here to indicate success and the order id seperately.
     function trade(address _sender, Order.TradeDirections _direction, IMarket _market, uint256 _outcome, uint256 _fxpAmount, uint256 _price, bytes32 _betterOrderId, bytes32 _worseOrderId, bytes32 _tradeGroupId) internal returns (bytes32) {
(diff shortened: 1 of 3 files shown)


source/contracts/trading/Trade.sol
```diff
@@ -76,6 +75,62 @@ contract Trade is CashAutoConverter, ReentrancyGuard, MarketValidator {
         return _bestFxpAmount;
     }
 
+    function publicBuyWithLimit(IMarket _market, uint256 _outcome, uint256 _fxpAmount, uint256 _price, bytes32 _betterOrderId, bytes32 _worseOrderId, bytes32 _tradeGroupId, uint256 _loopLimit) external payable marketIsLegit(_market) convertToAndFromCash onlyInGoodTimes returns (bytes32) {
```
adrake33: Isn't `_tradeGroupId` supposed to be optional, since it's only intended to be used by the UI? If so, would it be better to make it the last parameter in these functions?

nuevoalex: You can't skip arguments anyway so that's only useful in the pyethereum tests where it will pick a default null value.

More importantly (and very oddly) the way stack memory works in Solidity it is important that the order of arguments remains the same when passing arguments into other functions since reordering increases the stack size and we hit the limit easier. In this case we send the tradeGroupId to a logging function but not the loopLimit so we wouldnt want to place it before the tradeGroupId

epheph: There's no such thing as optional in solidity

PR #650, opened 2018-05-26
gas optimizations for order fills

nuevoalex:
A few relatively small optimizations that will save us a good bit of gas in trading

### DIFF
--- a/source/contracts/Augur.sol
+++ b/source/contracts/Augur.sol
@@ -60,6 +60,7 @@ contract Augur is Controlled, IAugur {
 
     mapping(address => bool) private universes;
     mapping(address => bool) private crowdsourcers;
+    mapping(address => bool) private shareTokens;
 
     //
     // Universe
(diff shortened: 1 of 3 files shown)


nuevoalex:
This isn't necessary btw at all, but will trim off about 130k gas from binary fills and around 510k gas from cat 8 fills.

If we'd rather just not make contract changes that's ok too.

bthaile:
I think now is the time. Ether will probably go up in price so any savings is a no-brainer.

source/contracts/Augur.sol
```diff
@@ -103,6 +104,21 @@ contract Augur is Controlled, IAugur {
         return true;
     }
 
+    //
+    // Share Tokens
+    //
+
+    function recordMarketShareTokens(IMarket _market) public view returns (bool) {
```
epheph: Drop view?

epheph:
Excellent, reducing gas in trading is worthwhile.

source/contracts/Augur.sol
nuevoalex: bah! Thanks!

PR #651, opened 2018-05-27
don't create orders when there are still any on the book

nuevoalex:
That match the given price and type. Overloading the meaning of the return value in the internal fill function somewhat here to avoid stack overflow :[ :[


epheph:
OK, i see, so basically, if that loop ends "prematurely" as in orders were not exhausted (since an orderId remained) (such as if they used a small "loop limit"), then don't return how much remained they wish to fill, return 0, to influence the caller to NOT create an order that is likely to be in cross

PR #652, opened 2018-05-27
Dont adjust OI on Set actions

nuevoalex:
We recently changed to a new OI scheme of immediately reducing Universe OI by all of a market's outstanding complete sets at finalization.

This is needed to prevent leakage from the failure to sell outstanding shares and leakage from invalid payouts since the claimed amount for a set will be less than what was paid (even before fees)

In doing this however I forgot to restrict further OI modification for that market, which is needed since otherwise someone could buy sets then claim proceeds which would make Universe OI increase but never decrease from that action.

### DIFF
--- a/source/contracts/trading/CompleteSets.sol
+++ b/source/contracts/trading/CompleteSets.sol
@@ -40,7 +40,9 @@ contract CompleteSets is Controlled, CashAutoConverter, ReentrancyGuard, MarketV
             _market.getShareToken(_outcome).createShares(_sender, _amount);
         }
 
-        _market.getUniverse().incrementOpenInterest(_cost);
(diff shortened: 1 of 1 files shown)


epheph:
What if someone bought complete sets before finalization, but turned them in after finalization, instead of claiming proceeds? Not sure why they would, but it seems we would leak that way?

nuevoalex:
At finalization all outstanding complete sets OI is decremented, so theyd be accounted for then

nuevoalex:
Claiming proceeds never affects OI and since this is past finalization buying or selling complete sets wont either

epheph:
Ahh, got it, so you're basically freezing any impact a finalize market has on OI. 🦊

PR #654, opened 2018-05-29
add a more resiliant transfer to order creator when filling orders

nuevoalex:
We were using a transfer here, which means if the order creator failed to receive their ETH the fill will fail. There is a malicious/failure case where this means the orderbook is effectively broken for a market with such an order creator.

This will instead _attempt_ to send them ETH but if that fails simply transfer Cash to them. There is still a recourse then for them at least, and everyone else gets to get the good UX of no Cash.

### DIFF
--- a/source/contracts/trading/Cash.sol
+++ b/source/contracts/trading/Cash.sol
@@ -47,6 +47,17 @@ contract Cash is DelegationTarget, ITyped, VariableSupplyToken, ICash {
         return true;
     }
 
+    function withdrawEtherToIfPossible(address _to, uint256 _amount) external returns (bool) {
+        require(_amount > 0 && _amount <= balances[msg.sender]);
+        if (_to.send(_amount)) {
+            burn(msg.sender, _amount);
+        } else {
+            internalTransfer(msg.sender, _to, _amount);
+        }
+        assert(this.balance >= totalSupply());
+        return true;
+    }
+
     function getTypeName() public view returns (bytes32) {
         return "Cash";
     }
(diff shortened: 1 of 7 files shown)


source/contracts/trading/Cash.sol
```diff
@@ -47,6 +47,17 @@ contract Cash is DelegationTarget, ITyped, VariableSupplyToken, ICash {
         return true;
     }
 
+    function withdrawEtherToIfPossible(address _to, uint256 _amount) external returns (bool) {
+        require(_amount > 0 && _amount <= balances[msg.sender]);
+        if (_to.send(_amount)) {
```
epheph: This change generally makes me a little uneasy, given that we are attempting to send value twice. It does make me feel a tad better to think about `.transfer()` as `require(.send())`, so if someone could influence the boolean response while still accepting the funds, there would be consequences everywhere.

source/contracts/trading/FillOrder.sol
```diff
@@ -368,7 +368,7 @@ contract FillOrder is CashAutoConverter, ReentrancyGuard, IFillOrder {
         uint256 _creatorCashBalance = _tradeData.contracts.denominationToken.balanceOf(_tradeData.creator.participantAddress);
         if (_creatorCashBalance > 0) {
             _tradeData.contracts.augur.trustedTransfer(_tradeData.contracts.denominationToken, _tradeData.creator.participantAddress, this, _creatorCashBalance);
-            _tradeData.contracts.denominationToken.withdrawEtherTo(_tradeData.creator.participantAddress, _creatorCashBalance);
+            _tradeData.contracts.denominationToken.withdrawEtherToIfPossible(_tradeData.creator.participantAddress, _creatorCashBalance);
```
epheph: Never noticed this here before, so denomination HAS to be cash, for this version

PR #665, opened 2018-06-03
Updated ContractInterfaces.ts

tinybike:
Because it's there


epheph:
It doesn't hurt to commit this early/manually, but this file being generated and added to git is part of the deploy process

https://github.com/AugurProject/augur-core/blob/master/package.json#L23

PR #667, opened 2018-06-06
binary -> yesNo

nuevoalex:
find replace


source/libraries/ContractInterfaces.ts
```diff
@@ -6561,4 +6561,3 @@
     export function ContractFactory(connector: Connector, accountManager: AccountManager, address: string, defaultGasPrice: BN): Controlled {
         return new Controlled(connector, accountManager, address, defaultGasPrice);
     }
-
```
adrake33: Nit: should probably keep extra line at the end of file since the linter expects this.

adrake33:
@nuevoalex Did you want to wait to merge this until tomorrow, when all the other renaming gets done?

nuevoalex:
I think this has to go in first actually, otherwise yesNo market creation won't work.

PR #672, opened 2018-06-16
Change docker:test to use geth or parity

adrake33:


adrake33:
@bthaile I didn't see them used anywhere else in the code, but I could just re-add docker:test and make it default to geth.

bthaile:
That would be the safest way to go. I approve of this message.

PR #673, opened 2018-06-19
add version of tests that occurs in a child universe

nuevoalex:
Wanted to have our whole python test suite also run under the context of a child universe


joeykrug:
Such good parenting!

PR #674, opened 2018-06-19
upgrade solc to 0.4.24

nuevoalex:
simple update, no problems


nuevoalex:
Turns out more changes needed! workin on it now

package.json
```diff
@@ -37,8 +37,8 @@
     "docker:push": "docker push augurproject/augur-core:latest",
     "docker:run:npm": "docker run --rm augurproject/augur-core:latest",
     "docker:run:shell": "docker run --rm -it --entrypoint=bash augurproject/augur-core:latest",
-    "docker:run:test:integration:geth": "docker-compose -f ./source/support/test/integration/docker-compose-parity.yml up --abort-on-container-exit --force-recreate",
-    "docker:run:test:integration:parity": "docker-compose -f ./source/support/test/integration/docker-compose-geth.yml up --abort-on-container-exit --force-recreate",
+    "docker:run:test:integration:geth": "docker-compose -f ./source/support/test/integration/docker-compose-geth.yml up --abort-on-container-exit --force-recreate",
+    "docker:run:test:integration:parity": "docker-compose -f ./source/support/test/integration/docker-compose-parity.yml up --abort-on-container-exit --force-recreate",
```
adrake33: Whoops, good catch!

nuevoalex:
Closing this out. The increased byte size of the Market contract pushes it past the imposed limit. By reducing the compiler optimization runs to 1 we get it closer but this means actual TXs are less optimized and will cost more.

Even then it is still too large so I started experimenting with trivial optimizations but the safe ones I found were not enough to get Geth to accept it. Since this seems to just be causing pain and will now need non-semantic contract changes I'm closing this for now. Perhaps something to revisit when we make upgraded contracts.

PR #675, opened 2018-06-20
Local integration tests working

nuevoalex:


package.json
```diff
@@ -82,7 +82,7 @@
     "chai-as-promised": "7.1.1",
     "copy-dir": "0.3.0",
     "crypto-promise": "^2.0.0",
-    "ethereumjs-testrpc": "6.0.3",
+    "ganache-core": "git+https://gitlab.com/seeseplusplus/ganache-core.git#sdb-1.0.0-rc.1",
```
epheph: Will this be released as a package. Pinning to a branch is always concerning

nuevoalex: Yea, it'll just take some time. The author of the branch works there now but says it will still take some time to get in.

PR #688, opened 2018-06-26
oyente in CI

nuevoalex:


joeykrug:
Build failing with `ERROR: file not found: security:oyente`

PR #696, opened 2018-06-29
Flatten all contracts prior to compiling so that swarm hashes match exactly

epheph:
So we can upload to Etherscan. Had to create our own fork of solidity-flattener with a few of their PRs merged, and a small change I made.

https://pypi.org/project/augur-solidity-flattener/

Before running build:contracts, you need to have the solidity_flattener in your path, so I updated package.json to take care of fetching and activating it.

Also, it is python3, so we can't mix up with the test stuff. Also, integration tests will fail until we get a docker image that supports both python2 and python3.

Lastly, solcjs has bugs. the version we were using (0.4.20) hung for a very long time, then fails with
```
abort() at Error
    at jsStackTrace (/Users/epheph/augur/augur-core/node_modules/solc/soljson.js:1:19656)
    at stackTrace (/Users/epheph/augur/augur-core/node_modules/solc/soljson.js:1:19839)
    at Object.abort (/Users/epheph/augur/augur-core/node_modules/solc/soljson.js:19:10926)
    at _abort (/Users/epheph/augur/augur-core/node_modules/solc/soljson.js:1:955278)

```
Newest 0.4.24 fails with:
```
TypeError: Runtime.functionPointers[index] is not a function
    at jsCall_vi (/Users/epheph/augur/augur-core/node_modules/solc/soljson.js:1:1114636)
    at Array.V9a (/Users/epheph/augur/augur-core/node_modules/solc/soljson.js:12:99349)
    at dd (/Users/epheph/augur/augur-core/node_modules/solc/soljson.js:6:7027)
    at ccallFunc (/Users/epheph/augur/augur-core/node_modules/solc/soljson.js:1:8258)
    at /Users/epheph/augur/augur-core/node_modules/solc/soljson.js:1:8564
    at runWithReadCallback (/Users/epheph/augur/augur-core/node_modules/solc/wrapper.js:49:26)
    at compileStandard (/Users/epheph/augur/augur-core/node_modules/solc/wrapper.js:65:16)
```

ONLY when passing it standard json input of flattened solidity. Could be size? It is significantly larger (both input and output)


epheph:
@nuevoalex smt applies a patch? What's up with that. Everything is passing except that patch fails to apply

epheph:
(probably worth a re-review, i changed a lot since you last checked)

PR #697, opened 2018-06-29
Make migration script more reliable and log more output

nuevoalex:


source/libraries/LegacyRepMigrator.ts
```diff
@@ -98,11 +102,12 @@ export class LegacyRepMigrator {
...
-        const promises: Array<Promise<any>> = [];
-        for (let balancesChunk of this.chunkedBalances) {
-            promises.push(this.migrateOwners(balancesChunk));
-        }
-        await resolveAll(promises);
+        let i = 0;
+        await Parallel.each(this.chunkedBalances, async balancesChunk => {
+            await this.migrateOwners(balancesChunk);
+            i+=this.chunkSize;
+            if (i % 100 == 0) console.log(`Migrated ${i} balances`);
```
epheph: How many batches do you suppose there will be? Randomly printing 1% will work well enough?

epheph: Ahh, looks like plenty!

PR #707, opened 2018-07-05
remove gas bond

nuevoalex:
Please review thoroughly. Its entirely removal but still.


source/contracts/reporting/Market.sol
```diff
@@ -230,10 +228,8 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
         // If the designated reporter showed up return the no show bond to the market creator. Otherwise it will be used as stake in the first report.
         if (_reporter == _initialReporter.getDesignatedReporter()) {
             require(_reputationToken.transfer(owner, _repBalance));
```
epheph: Since there's only one transfer here now, this logic could be simplified, but i agree with what you've done here, changing as few lines as possible

source/contracts/reporting/Reporting.sol
```diff
@@ -15,10 +15,6 @@ library Reporting {
     uint256 private constant MAXIMUM_REPORTING_FEE_DIVISOR = 10000; // Minimum .01% fees
     uint256 private constant MINIMUM_REPORTING_FEE_DIVISOR = 3; // Maximum 33.3~% fees. Note than anything less than a value of 2 here will likely result in bugs such as divide by 0 cases.
 
-    // NOTE: We need to maintain this cost to roughly match the gas cost of reporting. This was last updated 12/22/2017
-    uint256 private constant GAS_TO_REPORT = 1500000;
-    uint256 private constant DEFAULT_REPORTING_GAS_PRICE = 5000000000;
```
epheph: I'm glad to see these lines in red.

source/contracts/reporting/Market.sol
```diff
@@ -84,8 +83,8 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
             shareTokens.push(createShareToken(_outcome));
         }
         approveSpenders();
-        // If the value was not at least equal to the sum of these fees this will throw. The addition here cannot overflow as these fees are capped
-        uint256 _refund = msg.value.sub(reporterGasCostsFeeAttoeth + validityBondAttoeth);
+        // If the value was not at least equal to this fee this will throw. The addition here cannot overflow as these fees are capped
```
maraoz: remove 'The addition here cannot overflow as these fees are capped'

Issue #689, opened 2018-06-27
Passing Tests w/ Oyente


tomkysar:
Submit a PR with tests passing with Oyente.

This may require interacting with the project maintainers to get it working 100% with Augur's contracts. We suggest working with them on bug reports, fail cases, etc. Please join the #dev or #bounties channel of our Discord with any questions.

Almost all of the errors are Oyente bugs, there are some failed assertions, but it's failing on spots where Augur doesn't even have an assertion.

Bounty posted with Gitcoin.

[1] https://github.com/melonproject/oyente

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 250.0 DAI (250.0 USD @ $1.0/DAI) attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/AugurProject/augur-core/689/697).
 * Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $28,748.35 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 316 years, 10 months from now. Please review their questions below:


1. [cryptomental](https://gitcoin.co/profile/cryptomental) has started work. 	
 * Q: Hi! I will try to tackle this one.

cryptomental:
It seems that the work was already done and submitted https://github.com/AugurProject/augur-core/pull/688 ,  https://travis-ci.com/AugurProject/augur-core/jobs/131758314 . I will then stop working on it.

joeykrug:
It wasn't  actually working [e.g. the oyente stuff throws a bunch of errors]

tomkysar:
cc: @cryptomental ^^

cryptomental:
Hi @joeykrug @tomkysar ok I get it now. I parsed Travis logs and I found quite a few overflows and a few assertion failures:

```
contract /app/source/contracts/trading/Orders.sol:Orders:
	============ Results ===========
	  EVM Code Coverage: 			 74.5%
	  Integer Underflow: 			 True
	  Integer Overflow: 			 True
[...]

PR #715, opened 2018-07-20
Feature/ignore selected oyente issues

cryptomental:
This PR allows to treat selected Oyente issue types as warnings instead of errors.

Reference and reason of the PR: https://github.com/AugurProject/augur-core/issues/689

### DIFF
--- a/source/support/test/oyente/Dockerfile
+++ b/source/support/test/oyente/Dockerfile
@@ -16,12 +16,8 @@ RUN wget --quiet https://github.com/Z3Prover/z3/archive/z3-4.5.0.zip \
 	&& cd /tmp/z3-z3-4.5.0 \
 	&& python scripts/mk_make.py --python \
 	&& cd build \
-	&& make \
+	&& make -j8 \
 	&& make install
-
-RUN wget --quiet https://github.com/melonproject/oyente/archive/master.zip \
-	&& unzip master.zip -d /tmp \
-	&& mv /tmp/oyente-master/oyente /usr/local/lib/python2.7/site-packages
 
 RUN wget --quiet https://gethstore.blob.core.windows.net/builds/geth-alltools-linux-amd64-1.8.2-b8b9f7f4.tar.gz \
 	&& tar -xvf geth-alltools-linux-amd64-1.8.2-b8b9f7f4.tar.gz \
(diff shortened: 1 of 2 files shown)


cryptomental:
I have no idea why Travis errored, the build is fine locally:

```
 $  git clone https://github.com/cryptomental/augur-core.git
 $  cd augur-core/
 $  git co feature/ignore-selected-oyente-issues
 $  npm run docker:run:test:security:oyente
```

nuevoalex:
@cryptomental I'm seeing this generate Assertion Failure Positive hits and still pass.

It looks like the config provided will ignore Assertion failures, but this is one of the two error types we actually want to be correct and fail on.

The two we care about are:

- Assertion Failures
- Re-Entrancy Vulnerability

Even then it appears to just be ignoring any failures too since I see Integer under/overflow hits that must be getting ignored (This is fine but I would expect it to have to be specified in the config)

I guess I have these two questions about the work done here:

- What will this actually fail on?
- Which false positives are fixed in this PR?

cryptomental:
Hi @nuevoalex , I have to split the answer in two parts. First, the reason of the behavior that you observe is that Oyente does not treat Integer underflow/overflow hits as errors but only reports them as warnings.

I hope that it will be easier to understand when you have a look at https://github.com/melonproject/oyente/blob/master/oyente/symExec.py#L2337

Oyente by default returns non-zero return code if any of [callstack, money_concurrency, time_dependency, reentrancy] vulnerabilities are found.

Additionally by specifying CHECK_ASSERTIONS , the assertions and parity multisig bug are analysed and if any assertion failure/parity bug is found, Oyente returns a non-zero return code.

A few lines above https://github.com/melonproject/oyente/blob/master/oyente/symExec.py#L2326 it can be seen, that integer_underflow and integer_overflow are logged, but they are not taken into account when returning a non-zero return code. 

Therefore, to answer your first question, after the change Oyente will fail if any of vulnerabilities are found, excluding integer overflow/underflow, AssertionFailure, and MoneyConcurrency.

Second, no false positives are fixed. All errors are shown, but assertions are not considered as errors.
False positives may have been fixed by Microsoft's Z3 solver, but as reported here https://github.com/Z3Prover/z3/issues/1766 the solver segfaults in the middle of analysis.

Please let me know if this is satisfactory, if not the all the work for the bounty will need to be cancelled and I will void the pull request, as fixing would be on Z3 side.

nuevoalex:
@cryptomental Are you fairly certain that once the Z3 problems are fixed that we could update this and start failing on assertion errors safely?

cryptomental:
@nuevoalex yes. We can leave this PR in a limbo until there is a response from Z3 side. Perhaps I can also have a look at their dev branch, try to build Z3 from there and dig deeper on what causes the segfault. I will play around and try to answer till Thursday the latest.

nuevoalex:
Sounds good! Thank you!

pgebheim:
@nuevoalex Can this be closed now or does this apply to V2/monorepo still?

PR #718, opened 2018-07-23
Mkr price feed spike

nuevoalex:
Includes a `RepPriceBridge.sol` which simply lets us `poke` to set the REP price from the MKR medianizer.

Also Includes `PriceFeedWrapper.sol` which can be used to both set the price of a feed and poke our Rep Price Bridge in one tx to simplify the process for everyone.


nuevoalex:
Thought more about the owner suggestion, but when looking at it I'd end up effectively recreating the owner authority with their authority system which is more fine tuned but also more complicated, so I'm not sure its worth it. I may be missing something though.

PR #719, opened 2018-07-24
add ipc to NetworkConfiguration

nuevoalex:


nuevoalex:
Some stuff was left out of the other PR

PR #728, opened 2018-08-11
Fixed market cap calc

tinybike:
(For v2)


nuevoalex:
It's just confusing because the labels are not accurate as well. This change does what you describe

PR #734, opened 2018-08-21
Solc 0.4.24

nuevoalex:
Reduced Market contract size by some removal of things that are redundant or won't be relevant in v2


source/contracts/reporting/Market.sol
```diff
@@ -216,7 +214,7 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
         for (uint256 j = 0; j < participants.length; j++) {
             _reportingParticipant = participants[j];
             if (_reportingParticipant.getPayoutDistributionHash() == winningPayoutDistributionHash) {
-                require(_reputationToken.transfer(_reportingParticipant, _reportingParticipant.getSize().div(2)));
+                require(_reputationToken.transfer(_reportingParticipant, _reportingParticipant.getSize() / 2));
```
epheph: For posterity: Dropping safe math because /0 is an exception and we don't have variable divisor anyway.

source/support/base/Dockerfile
```diff
@@ -9,7 +9,7 @@ FROM augurproject/python2-and-3:latest
 RUN curl -sL https://deb.nodesource.com/setup_8.x | bash - \
 	&& apt-get install -y nodejs
 
-RUN wget --quiet --output-document /usr/local/bin/solc https://github.com/ethereum/solidity/releases/download/v0.4.20/solc-static-linux \
```
epheph: Unrelated to this PR, but i wonder if we should vendor it or checksum?

pgebheim: I have a checksum for it in the monorepo base docker, can copy that over here.

nuevoalex:
The MKR stuff is no longer needed due to the reporting fee bug. Would just be a lot of unneeded complication

PR #739, opened 2018-08-28
Achapman/ch15701/reduce dispute times

nuevoalex:


source/contracts/reporting/Reporting.sol
```diff
@@ -2,7 +2,7 @@ pragma solidity 0.4.24;
 
 
 library Reporting {
-    uint256 private constant DESIGNATED_REPORTING_DURATION_SECONDS = 3 days;
+    uint256 private constant DESIGNATED_REPORTING_DURATION_SECONDS = 1 days;
```
epheph: Users will now have 24h instead of 72h from market endTime to submit a designated report. The impact here is that a market is almost 30% more likely to hit an earlier feeWindow for its first dispute round if designated reporter fails to show.

source/contracts/trading/ClaimTradingProceeds.sol
```diff
@@ -21,7 +21,6 @@ contract ClaimTradingProceeds is CashAutoConverter, ReentrancyGuard, MarketValid
 
     function claimTradingProceeds(IMarket _market, address _shareHolder) external marketIsLegit(_market) nonReentrant returns(bool) {
         require(_market.isFinalized());
-        require(controller.getTimestamp() > _market.getFinalizationTime().add(Reporting.getClaimTradingProceedsWaitTime()));
```
epheph: Removing the original waiting period for claiming proceeds, useless since there is no such thing as a BadTime any more.

source/contracts/reporting/Market.sol
```diff
@@ -144,7 +144,8 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
     }
 
     function contribute(uint256[] _payoutNumerators, bool _invalid, uint256 _amount) public returns (bool) {
-        require(feeWindow.isActive());
+        require(getInitialReporter().getReportTimestamp() != 0);
```
epheph: I believe this line has no functional effect. A market is not assigned a FeeWindow until it is InitialReport'd on (open or designated). This check will pass any time after initial report (when there is a non-zero FeeWindow). Calling isActive on FeeWindow(0) would also fail.

This is more explicit, and better behavior than making a call to FeeWindow(0)

source/contracts/reporting/Market.sol
```diff
@@ -144,7 +144,8 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
...
+        require(!feeWindow.isOver());
```
epheph: The change here is from:
```isActive()```
to
```!isOver()```

The new code is more permissive, and allows this code to proceed when FeeWindow has not started yet. This would allow crowdsourcers to be contributed to before the corresponding FeeWindow starts, such as immediately following an InitialReport?

Comment (nuevoalex) [reply source/contracts/reporting/Market.sol]: That's exactly correct. The idea being that any number of dispute rounds can occur in any time period but for the market to resolve it must last an entire fee window with no completed dispute.

File: source/contracts/reporting/Market.sol [hunk 5]
```diff
@@ -163,10 +164,13 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
         if (IDisputeCrowdsourcer(_reportingParticipant).getSize() >= universe.getDisputeThresholdForFork()) {
             universe.fork();
         } else {
+            IFeeWindow _originalFeeWindow = feeWindow;
             feeWindow = universe.getOrCreateNextFeeWindow();
-            // Participants is implicitly bounded by the floor of the initial report REP cost to be no more than 21
-            for (uint256 i = 0; i < participants.length; i++) {
-                participants[i].migrate();
+            if (feeWindow != _originalFeeWindow) {
```
epheph: BaseReportingParticipant.migrate() is a function which mints fee tokens for a reporting participant, rewarding those who have exposure over longer periods of time due to dispute process. Only when a market's feeWindow changes do we want to run this process, since less than a feeWindow's time has elapsed.
Additionally, since this logic would be getting their balance on a FeeWindow's token and doubling to it, migrate()ing every time would double their tokens every dispute round (instead of multiplying it by dispute round, across different FeeWindows), significantly rewarding early reporting participants.

PR #740, opened 2018-08-29
dispute pacing threshold

nuevoalex:


source/contracts/reporting/Market.sol
```diff
@@ -51,6 +51,7 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
     IMailbox private marketCreatorMailbox;
     uint256 private finalizationTime;
     uint256 private noShowBond;
+    bool private disputePacingOn;
```
epheph: disputePacingOn is a flag which dictates if a dispute round should wait for a new fee window or allow multiple disputes to take place within the same fee window. the idea here is that while early rounds can go through several rounds of dispute very quickly, it is best to give a bit of time to before getting closer to a fork

source/contracts/reporting/Market.sol
```diff
@@ -145,7 +146,11 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
 
     function contribute(uint256[] _payoutNumerators, bool _invalid, uint256 _amount) public returns (bool) {
         require(getInitialReporter().getReportTimestamp() != 0);
-        require(!feeWindow.isOver());
+        if (disputePacingOn) {
+            require(feeWindow.isActive());
+        } else {
+            require(!feeWindow.isOver());
+        }
```
epheph: If we have triggered disputePacing for this market, require that the feeWindow is currently active. successfully disputing pushes a market to the next fee window, ensuring this require will fail until that feeWindow's time has come.

If not dispute pacing, just make sure we haven't expired the current window

source/contracts/reporting/Market.sol
```diff
@@ -161,9 +166,11 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
     function finishedCrowdsourcingDisputeBond(IReportingParticipant _reportingParticipant) private returns (bool) {
         participants.push(_reportingParticipant);
         crowdsourcers = MapFactory(controller.lookup("MapFactory")).createMap(controller, this); // disavow other crowdsourcers
-        if (IDisputeCrowdsourcer(_reportingParticipant).getSize() >= universe.getDisputeThresholdForFork()) {
+        uint256 _crowdsourcerSize = IDisputeCrowdsourcer(_reportingParticipant).getSize();
+        if (_crowdsourcerSize >= universe.getDisputeThresholdForFork()) {
```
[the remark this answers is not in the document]
nuevoalex: Yea it is. Nice catch. Will fix

source/contracts/reporting/Universe.sol
```diff
@@ -65,6 +66,7 @@ contract Universe is DelegationTarget, ITyped, Initializable, IUniverse {
         forkReputationGoal = _totalRepSupply.div(2); // 50% of REP migrating results in a victory in a fork
         disputeThresholdForFork = _totalRepSupply.div(40); // 2.5% of the total rep supply
         initialReportMinValue = disputeThresholdForFork.div(3).div(2**18).add(1); // This value will result in a maximum 20 round dispute sequence
+        disputeThresholdForDisputePacing = disputeThresholdForFork.div(2**5); // Disputes begin normal pacing once there are 4 rounds remaining. The "last" round is the one that causes a fork and requires no time so the exponent here is 5 to provide for that many rounds actually occuring.
```
epheph: Pedantic, but this comment could be more clear. `Disputes begin normal pacing no later than 4 rounds remaining.` ?

PR #742, opened 2018-09-05
Achapman/ch15702/use or lose rep

nuevoalex:
Use or Lose Implementation.

 - No more fork bonus
 - Migrating REP is disallowed past the fork window
 - Market disavowal now returns the No Show Bond / Initial Report Bond
 - Migration to a new Universe requires a No Show Bond / Initial Report be placed

### DIFF
--- a/source/contracts/legacy_reputation/OldLegacyRepToken.sol
+++ b/source/contracts/legacy_reputation/OldLegacyRepToken.sol
@@ -82,7 +82,7 @@ contract OldLegacyReputationToken is DelegationTarget, ITyped, Initializable, Va
         totalMigrated += _attotokens;
         // Award a bonus if migration is done before the fork period is over, even if it has finalized
         if (controller.getTimestamp() < _parentUniverse.getForkEndTime()) {
-            uint256 _bonus = _attotokens.div(Reporting.getForkMigrationPercentageBonusDivisor());
+            uint256 _bonus = _attotokens.div(20);
             mint(_reporter, _bonus);
             totalTheoreticalSupply += _bonus;
         }
--- a/source/contracts/reporting/InitialReporter.sol
+++ b/source/contracts/reporting/InitialReporter.sol
@@ -59,24 +59,17 @@ contract InitialReporter is DelegationTarget, Ownable, BaseReportingParticipant,
         return true;
     }
 
-    function resetReportTimestamp() public returns (bool) {
+    function returnRepFromDisavow() public returns (bool) {
         require(IMarket(msg.sender) == market);
-        if (reportTimestamp == 0) {
-            return;
-        }
-        reportTimestamp = controller.getTimestamp();
+        require(reputationToken.transfer(owner, reputationToken.balanceOf(this)));
+        reportTimestamp = 0;
         return true;
     }
 
-    function migrateREP() public returns (bool) {
+    function migrateToNewUniverse(address _designatedReporter) public returns (bool) {
         require(IMarket(msg.sender) == market);
-        IUniverse _newUniverse = market.getUniverse();
-        IReputationToken _newReputationToken = _newUniverse.getReputationToken();
-        uint256 _balance = reputationToken.balanceOf(this);
-        if (_balance > 0) {
-            reputationToken.migrateOut(_newReputationToken, _balance);
-        }
-        reputationToken = _newReputationToken;
+        designatedReporter = _designatedReporter;
+        reputationToken = market.getUniverse().getReputationToken();
         return true;
     }
 
--- a/source/contracts/reporting/Market.sol
+++ b/source/contracts/reporting/Market.sol
@@ -310,9 +310,11 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
         // Pay the No Show REP bond
         noShowBond = universe.getOrCacheDesignatedReportStake();
         noShowBondOwner = msg.sender;
+        getReputationToken().trustedMarketTransfer(noShowBondOwner, this, noShowBond);
+
+        // Update the Initial Reporter
         IInitialReporter _initialReporter = getInitialReporter();
         _initialReporter.migrateToNewUniverse(msg.sender);
-        getReputationToken().trustedMarketTransfer(noShowBondOwner, this, noShowBond);
 
         // If the market is past expiration use the reporting data to make an initial report
         uint256 _timestamp = controller.getTimestamp();
--- a/source/contracts/reporting/Market.sol
+++ b/source/contracts/reporting/Market.sol
@@ -52,6 +52,7 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
     uint256 private finalizationTime;
     uint256 private noShowBond;
     bool private disputePacingOn;
+    address private noShowBondOwner;
 
     // Collections
     IReportingParticipant[] public participants;
@@ -69,6 +70,7 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
         universe = _universe;
         require(!universe.isForking());
         owner = _creator;
+        noShowBondOwner = owner;
         assessFees();
         endTime = _endTime;
         numOutcomes = _numOutcomes;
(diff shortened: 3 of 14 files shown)


source/contracts/reporting/Market.sol
```diff
@@ -273,12 +281,14 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
         return _crowdsourcer;
     }
 
-    function migrateThroughOneFork() public returns (bool) {
+    function migrateThroughOneFork(uint256[] _payoutNumerators, bool _invalid, string _description) public returns (bool) {
```
epheph: These arguments are ignored when migrating a market that has not ended?

nuevoalex: Yea

source/contracts/reporting/Market.sol
```diff
@@ -324,8 +331,21 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
...
         IInitialReporter _initialParticipant = getInitialReporter();
+        // Early out if already disavowed or nothing to disavow
+        if (_initialParticipant.getReportTimestamp() == 0) {
+            return true;
+        }
         delete participants;
         participants.push(_initialParticipant);
+        // Send REP from the no show bond back to the address that placed it. If a report has been made tell the InitialReporter to return that REP and reset
+        IReputationToken _reputationToken = getReputationToken();
+        uint256 _balance = _reputationToken.balanceOf(this);
```
epheph: If someone transfers 1 attorep to the market after initial report, it would prevent the original inital reporter from getting their rep back? And whoever performed the migrateThroughOneFork would get all the original reporter's staked rep?

source/contracts/legacy_reputation/OldLegacyRepToken.sol
```diff
@@ -82,7 +82,7 @@ contract OldLegacyReputationToken is DelegationTarget, ITyped, Initializable, Va
         totalMigrated += _attotokens;
         // Award a bonus if migration is done before the fork period is over, even if it has finalized
         if (controller.getTimestamp() < _parentUniverse.getForkEndTime()) {
-            uint256 _bonus = _attotokens.div(Reporting.getForkMigrationPercentageBonusDivisor());
+            uint256 _bonus = _attotokens.div(20);
```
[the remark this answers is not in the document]
nuevoalex: This is the Old Legacy REP Token. Since the bonus is no longer a thing in the real contracts I removed the constant which this was referencing.

source/contracts/reporting/Market.sol
```diff
@@ -324,8 +332,20 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
         require(_forkingMarket != this);
         require(!isFinalized());
         IInitialReporter _initialParticipant = getInitialReporter();
+        // Early out if already disavowed or nothing to disavow
+        if (_initialParticipant.getReportTimestamp() == 0) {
```
epheph: Note, not feedback: If an initial reporter has not been been reported on, you can both keep the original InitialReporter (keeping the designated reporter/using the open-reporting rules) and know that no disputeCrowdsourcers exist.

source/contracts/reporting/Market.sol
epheph: Worth a comment? Optional required parameters are a bit unusual

source/contracts/reporting/ReputationToken.sol
```diff
@@ -53,14 +53,9 @@ contract ReputationToken is DelegationTarget, ITyped, Initializable, VariableSup
     function migrateIn(address _reporter, uint256 _attotokens) public afterInitialized returns (bool) {
         IUniverse _parentUniverse = universe.getParentUniverse();
         require(ReputationToken(msg.sender) == _parentUniverse.getReputationToken());
+        require(controller.getTimestamp() < _parentUniverse.getForkEndTime());
```
epheph: Note, not feedback: Instead of checking that current time is before forkEnd time to provide bonus, simply require it. All rep not moved before this time will cause a tx revert, locking REP in parent universe.

source/contracts/reporting/Market.sol
```diff
@@ -115,28 +117,35 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
     }
 
     function doInitialReport(uint256[] _payoutNumerators, bool _invalid, string _description) public returns (bool) {
+        doInitialReportInternal(msg.sender, _payoutNumerators, _invalid, _description);
+        return true;
+    }
+
+    function doInitialReportInternal(address _reporter, uint256[] _payoutNumerators, bool _invalid, string _description) public returns (bool) {
```
epheph: Should this be `public`?

source/contracts/reporting/Market.sol
```diff
@@ -324,8 +332,20 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
...
         require(!isFinalized());
         IInitialReporter _initialParticipant = getInitialReporter();
+        // Early out if already disavowed or nothing to disavow
+        if (_initialParticipant.getReportTimestamp() == 0) {
+            return true;
+        }
         delete participants;
         participants.push(_initialParticipant);
+        // Send REP from the no show bond back to the address that placed it. If a report has been made tell the InitialReporter to return that REP and reset
+        IReputationToken _reputationToken = getReputationToken();
```
epheph: Woldl save gas and maybe be more clear if this moved inside the `if`

source/contracts/reporting/Market.sol
```diff
@@ -69,6 +70,7 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
         universe = _universe;
         require(!universe.isForking());
         owner = _creator;
+        noShowBondOwner = owner;
```
epheph: Note: When a [non-forking] market is migrated to a new [winning] universe, if the market has not yet been initially reported on, the original owner of the market (who will be the noShowBondOwner here) will get their no-show bond back, and whoever is performing the migration will place a new bond (which could be a different size based on universe). We need a variable to track when the owner and the last user to perform the migrateThroughOneFork differ

source/contracts/reporting/Market.sol
```diff
@@ -324,8 +332,20 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
...
         IInitialReporter _initialParticipant = getInitialReporter();
+        // Early out if already disavowed or nothing to disavow
+        if (_initialParticipant.getReportTimestamp() == 0) {
+            return true;
+        }
         delete participants;
         participants.push(_initialParticipant);
+        // Send REP from the no show bond back to the address that placed it. If a report has been made tell the InitialReporter to return that REP and reset
+        IReputationToken _reputationToken = getReputationToken();
+        if (noShowBond > 0) {
```
epheph: Note: If there's a no-show bond present on the market, no initial report has occurred, and we return it to the noShowBondOwner (who is either the original owner, or the noShowBondOwner if this market has already been migrated previously)).

If the initial report HAS occurred, and we are disavowing, send back the rep to the initial reporter (not necessary for the crowdsourcers, since they recognize their disavowal via non-membership to market)). This initialReporter will be re-used in the next universe

source/contracts/reporting/Market.sol
```diff
@@ -293,28 +304,25 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
...
         _destinationUniverse.addMarketTo();
         _currentUniverse.removeMarketFrom();
-        IReputationToken _oldReputationToken = getReputationToken();
         universe = _destinationUniverse;
 
         universe.incrementOpenInterestFromMarket(_marketOI);
 
-        // reset state back to Initial Reporter
-        IInitialReporter _initialParticipant = getInitialReporter();
+        // Pay the No Show REP bond
```
epheph: Due to disavowal, there is NO rep associated with this market or its participants any more.  The user that is performing this migrateThroughOneFork is will to either stake 
1.) NoShowBond (with no benefit to themselves) if the market has not reached endTime. 
2.) The InitialReport if the market HAS reached endTime. To do this, it does take a trip through the noShowBond, but it really has no effect, code was just structured this way to keep conditional logic clean

The "re-used" initialReport gets updated with the sender's address and starts using the new universe's rep.

If market has not reached endtime, the arguments to this function are completely unused (and noShowBond remains on the market). If it HAS reached endtime, it performs an initial report, using the noShowBond that was JUST put in place.

source/contracts/reporting/Market.sol
```diff
@@ -324,8 +332,20 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
         require(_forkingMarket != this);
```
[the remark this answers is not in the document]
epheph: I figured out why you might want to leave this public. If a user has rep staked, they need to migrate the REP in order to use that same rep for the noShowBond after migration.

source/contracts/reporting/Market.sol
```diff
@@ -115,28 +117,35 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
     }
 
     function doInitialReport(uint256[] _payoutNumerators, bool _invalid, string _description) public returns (bool) {
+        doInitialReportInternal(msg.sender, _payoutNumerators, _invalid, _description);
+        return true;
+    }
+
+    function doInitialReportInternal(address _reporter, uint256[] _payoutNumerators, bool _invalid, string _description) public returns (bool) {
+        require(!universe.isForking());
```
epheph: Note: you can't report when a universe is forking or else anyone can come along and un-do your initial report.

source/contracts/reporting/Market.sol
```diff
@@ -115,28 +117,35 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
...
     }
 
     function distributeInitialReportingRep(address _reporter, IInitialReporter _initialReporter) private returns (uint256) {
         IReputationToken _reputationToken = getReputationToken();
         uint256 _initialReportStake = noShowBond;
-        // If the designated reporter showed up return the no show bond to the market creator. Otherwise it will be used as stake in the first report.
+        // If the designated reporter showed up return the no show bond to the bond owner. Otherwise it will be used as stake in the first report.
         if (_reporter == _initialReporter.getDesignatedReporter()) {
-            require(_reputationToken.transfer(owner, _initialReportStake));
+            require(_reputationToken.transfer(noShowBondOwner, _initialReportStake));
```
epheph: Note: yup, goes to the noShowBondOwner, which could differ from owner/creator

source/contracts/reporting/InitialReporter.sol
```diff
@@ -59,24 +59,17 @@ contract InitialReporter is DelegationTarget, Ownable, BaseReportingParticipant,
...
     }
 
-    function resetReportTimestamp() public returns (bool) {
+    function returnRepFromDisavow() public returns (bool) {
         require(IMarket(msg.sender) == market);
-        if (reportTimestamp == 0) {
-            return;
-        }
-        reportTimestamp = controller.getTimestamp();
+        require(reputationToken.transfer(owner, reputationToken.balanceOf(this)));
```
epheph: Owner is the original creator of the initialReporter. since we re-use initialReporters, won't this reward the wrong person if say, a market was first migrated to a winning universe, then inside THAT universe, there was a "returnRepFromDisavow" ?

source/contracts/reporting/Market.sol
nuevoalex: It should not. Will fix

source/contracts/reporting/InitialReporter.sol
nuevoalex: owner gets set whenever a report occurs so any legitimate REP in the InitialReport will be associated with the last account to do a report.

If for example this was migrated to a winning universe then either it is not altered at all because there was no report, or the owner is changed to the account that migrates when they do the report. Then if inside of that universe `returnRepFromDisavow` is called the user that migrated and paid for the report will get their REP back.

epheph: I see. owner remains on the old owner, but it doesn't matter because it is value-less? And the minute a report comes in, and there _IS_ value, then owner is set.

PR #744, opened 2018-09-14
Changes from running npm audit fix

tinybike:
I ran `npm audit` on augur-core, which generated some [alarming-looking output](https://gist.github.com/tinybike/e72182e3737f0e33b60575ef141a650e), followed by a suggestion to run `npm audit fix`.  I did so, and it changed a bunch of stuff in package-lock.json.  It appears to mostly be un-fixing package version numbers, which doesn't seem desirable per se.

@nuevoalex up to you what (if anything) to do with this PR, feel free to close it if it's not needed!


joeykrug:
Has this been run on the UI and augur node repos? If not would be good to run there too imo

PR #745, opened 2018-09-18
Achapman/ch15705/trading api improvements

nuevoalex:
- Simplify API by removing unused (in our code base) and redundant methods
 - Simplify / fix the fill order loop in trading
 - Add ability to ignore owned shares when creating / filling / trading
 - Add ability to trade in terms of total cost instead of shares
 - fxpAmount -> amount
 - displayPrice -> price


source/contracts/trading/Trade.sol
```diff
@@ -15,133 +15,105 @@ import 'libraries/CashAutoConverter.sol';
...
+        while (_orderId != 0 && _bestAmount > 0 && _tradeData.loopLimit > 0 && isMatch(_orderId, _type, _orderPrice, _tradeData.price)) {
+            bytes32 _nextOrderId = _orders.getWorseOrderId(_orderId);
+            _orders.setPrice(_tradeData.market, _tradeData.outcome, _orderPrice);
+            _bestAmount = IFillOrder(controller.lookup("FillOrder")).fillOrder(_tradeData.sender, _orderId, _bestAmount, _tradeData.tradeGroupId, _tradeData.ignoreShares);
+            _orderId = _nextOrderId;
+            _orderPrice = _orders.getPrice(_orderId);
+            _tradeData.loopLimit -= 1;
         }
-        if (_orderId != 0) {
+        if (_orderId != 0 && isMatch(_orderId, _type, _orderPrice, _tradeData.price)) {
```
adrake33: This seems a little redundant since `isMatch` already returns false if `_orderId` is 0. Was this intentional?

adrake33:
Looks like Travis is also complaining about several unused function params in Trade.sol.

source/contracts/trading/Trade.sol
nuevoalex: Good point, can just be `isMatch`!

adrake33:
There are some parameter mismatches in TestFixture.ts that still need to be resolved.

PR #750, opened 2018-10-03
Achapman/ch15793/buy back system

nuevoalex:
This includes both the Auction system and the buyback system for fees since they're somewhat intertwined.

Documentation on the Auction System can be found here: 

https://docs.google.com/document/d/1q5uMM0wjRqgBlVzc_B2J7RGocnkHKrETlGgVwzdLW3c/edit

The integration tests likely don't pass right now and I still want to sit on this a bit and review it but its large enough that I wanted to get some eyes on it.

One thing that is not included here is a shutdown for the auction such that it would actually stop running (and minting/burning REP). That may be something we want in addition to toggling if the auction is actually used to set reporting fees.

I wrote the Auction a bit differently than for example the FeeWindow in that it is a single instance per Universe and its state is implicit based on the current time. This simplifies interaction with the contract a lot, reduces (maybe eliminates?) the need for Augur Node processing, and removes the need to handle transfers between auctions.


source/contracts/reporting/Auction.sol
```diff
@@ -0,0 +1,239 @@
...
+        uint256 _attoRepSold = initialAttoRepBalance.sub(currentAttoRepBalance);
+        uint256 _newTotalAttoRepSold = _attoRepSold.add(_attoRepAmount);
+        currentUpperBoundRepPriceInAttoEth = currentUpperBoundRepPriceInAttoEth
+            .mul(_attoRepSold)
+            .add(_attoRepAmount.mul(_repPriceInAttoEth))
+            .div(_newTotalAttoRepSold);
+        currentAttoRepBalance = currentAttoRepBalance.sub(_attoRepAmount);
+        // We don't update the price while in bootstrap mode since its a one sided auction
+        if (!bootstrapMode) {
+            currentRepPrice = currentUpperBoundRepPriceInAttoEth.add(currentUpperBoundRepPriceInAttoEth) / 2;
```
epheph: currentUpperBoundRepPriceInAttoEth ?

source/contracts/reporting/Auction.sol
```diff
@@ -0,0 +1,239 @@
...
+        uint256 _timePassed = controller.getTimestamp().sub(initializationTime).sub(currentAuctionIndex * 1 days);
+        uint256 _priceDecrease = initialRepSalePrice.mul(_timePassed) / Reporting.getAuctionDuration();
+        return initialRepSalePrice.sub(_priceDecrease);
+    }
+
+    function getEthSalePriceInAttoRep() public initializeNewAuctionIfNeeded returns (uint256) {
+        require(!bootstrapMode);
+        uint256 _timePassed = controller.getTimestamp().sub(initializationTime).sub(currentAuctionIndex * 1 days);
+        uint256 _priceDecrease = initialEthSalePrice.mul(_timePassed) / Reporting.getAuctionDuration();
+        return initialEthSalePrice.sub(_priceDecrease);
```
epheph: _priceDecrease could go above initialEthSalePrice if timePassed is greater than duration. I don't know what behavior we want there, but I imagine it's common if dust is left and it isn't worth the gas to take, even for free?

nuevoalex: In that case the auction would be over and the modifier to initialize would fail since there is no active auction.

Also the sub would raise an exception when it resulted in overflow

epheph: Maybe that's fine, error means disabled

source/contracts/LegacyReputationToken.sol
```diff
@@ -37,4 +37,19 @@ contract LegacyReputationToken is OldLegacyReputationToken {
...
+
+    // Interface fulfilment
+    function mintForAuction(uint256) public returns (bool) {
+        return true;
+    }
+
+    // Interface fulfilment
+    function burnForAuction(uint256) public returns (bool) {
+        return true;
+    }
```
MicahZoltu: We should not be adding any code to `LegacyReputationToken`.  We cannot change that code (it is already deployed) so changing it here is only lying to ourselves and likely to cause us to think something is working in test/test network only to find out it doesn't work on mainnet.

source/contracts/reporting/Auction.sol
```diff
@@ -0,0 +1,239 @@
...
+contract Auction is DelegationTarget, Initializable, IAuction {
+    using SafeMathUint256 for uint256;
+
+    enum AuctionType {
+        DORMANT_1,
+        DORMANT_2,
+        UNRECORDED,
+        DORMANT_3,
+        DORMANT_4,
+        DORMANT_5,
```
MicahZoltu: Would love some better names for these than `DORMANT_X`.

source/contracts/reporting/Auction.sol
```diff
@@ -0,0 +1,239 @@
...
+        currentLowerBoundRepPriceInAttoEth = 0;
+
+        currentAuctionIndex = _currentAuctionIndex;
+
+        initialRepSalePrice = lastRepPrice.mul(Reporting.getAuctionInitialPriceMultiplier());
+        initialEthSalePrice = Reporting.getAuctionInitialPriceMultiplier().mul(10**36).div(lastRepPrice);
+        return true;
+    }
+
+    function tradeRepForEth(uint256 _attoEthAmount) public initializeNewAuctionIfNeeded returns (bool) {
```
MicahZoltu: Style: I'm not a huge fan of using modifiers for this sort of thing.  I like to think of modifiers as preconditions, while this modifier is an action (potentially).

source/contracts/reporting/Auction.sol
```diff
@@ -0,0 +1,239 @@
...
+        require(!bootstrapMode || currentAuctionType == AuctionType.UNRECORDED);
+        bootstrapped = true;
+
+        uint256 _auctionRepBalanceTarget = reputationToken.totalSupply() / Reporting.getAuctionTargetSupplyDivisor();
+        uint256 _repBalance = reputationToken.balanceOf(this);
+
+        if (_repBalance < _auctionRepBalanceTarget) {
+            reputationToken.mintForAuction(_auctionRepBalanceTarget.sub(_repBalance));
+        } else {
+            reputationToken.burnForAuction(_repBalance.sub(_auctionRepBalanceTarget));
```
MicahZoltu: We should only burn REP that was sourced from fee payouts.  Excess REP from an attack against the auction should be used to make the following auction larger, thus increasing the cost of a sustained attack against the system with each iteration.

source/contracts/reporting/Auction.sol
```diff
@@ -0,0 +1,239 @@
...
+        uint256 _newTotalAttoRepSold = _attoRepSold.add(_attoRepAmount);
+        currentUpperBoundRepPriceInAttoEth = currentUpperBoundRepPriceInAttoEth
+            .mul(_attoRepSold)
+            .add(_attoRepAmount.mul(_repPriceInAttoEth))
+            .div(_newTotalAttoRepSold);
+        currentAttoRepBalance = currentAttoRepBalance.sub(_attoRepAmount);
+        // We don't update the price while in bootstrap mode since its a one sided auction
+        if (!bootstrapMode) {
+            currentRepPrice = currentLowerBoundRepPriceInAttoEth.add(currentUpperBoundRepPriceInAttoEth) / 2;
+        }
```
MicahZoltu: How much code could we simplify if we skipped bootstrap mode and just seeded the system with some ETH?  (lets ignore for a moment where we get that ETH)

source/contracts/reporting/Auction.sol
```diff
@@ -0,0 +1,239 @@
...
+
+    function tradeEthForRep(uint256 _attoRepAmount) public initializeNewAuctionIfNeeded payable returns (bool) {
+        require(currentAttoRepBalance > 0);
+        require(_attoRepAmount > 0);
+        _attoRepAmount = _attoRepAmount.min(currentAttoRepBalance);
+        uint256 _repPriceInAttoEth = getRepSalePriceInAttoEth();
+        uint256 _attoEthCost = _attoRepAmount.mul(_repPriceInAttoEth) / 10**18;
+        // This will raise an exception if insufficient ETH was sent
+        msg.sender.transfer(msg.value.sub(_attoEthCost));
+        reputationToken.transfer(msg.sender, _attoRepAmount);
```
MicahZoltu: I think the auction would be "better" if it was a uniform price auction instead (everyone gets the same price of the last bidder).  I think we should weigh the benefits of a uniform price auction (encourages people to bid early rather than playing a game of chicken) against the costs (code complexity).

As far as implementation, my initial thinking is that people trade ETH/REP for tokens that represent how much ETH/REP they contributed, then when the auction ends (due to amount of `ETH * current_ETH_price >= REP_available` or `REP * current_REP_price >= ETH_available` everyone gets a share of the ETH/REP available proportionate to their share of the auction tokens they hold.  I don't think code complexity will go up that much, though would need a prototype to verify that.

source/contracts/reporting/Auction.sol
```diff
@@ -0,0 +1,239 @@
...
+        manualRepPriceInAttoEth = _repPriceInAttoEth;
+        return true;
+    }
+
+    function addToAuthorizedPriceFeeders(address _priceFeeder) public afterInitialized onlyKeyHolder returns (bool) {
+        authorizedPriceFeeders[_priceFeeder] = true;
+        return true;
+    }
+
+    function removeFromAuthorizedPriceFeeders(address _priceFeeder) public afterInitialized onlyKeyHolder returns (bool) {
```
MicahZoltu: Naming: This function name shouldn't be pluralized, it only removes a single `priceFeeder`.  Same with previous function.

source/contracts/reporting/Market.sol
```diff
@@ -47,7 +47,7 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
     uint256 private endTime;
     uint256 private numOutcomes;
     bytes32 private winningPayoutDistributionHash;
-    uint256 private validityBondAttoeth;
+    uint256 private validityBondAttoEth;
```
MicahZoltu: Naming: `millimeters` or `milliMeters`?

source/contracts/LegacyReputationToken.sol
nuevoalex: This is the ReputationTokenV1 since in the context of v2 the LegacyReputationToken is the REP token of v1.

I guess I can just create a second IReputationToken interface instead of sharing the real one. Will do that.

source/contracts/reporting/Auction.sol
nuevoalex: Suggest!

source/contracts/reporting/Auction.sol
```diff
@@ -0,0 +1,239 @@
...
+        // If this auction is over and it is a recorded auction use the price it found
+        if (getAuctionIndexForCurrentTime() != currentAuctionIndex && currentAuctionType == AuctionType.RECORDED) {
+            return currentRepPrice;
+        }
+
+        return repPrice;
+    }
+
+    function getAuctionIndexForCurrentTime() public view returns (uint256) {
+        return controller.getTimestamp().sub(initializationTime) / Reporting.getAuctionDuration();
```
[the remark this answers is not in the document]
epheph: This is what the dormant's are for. Only some of the AuctionIndex's are valid to run an auction on (UNRECORDED, RECORDED)

source/contracts/reporting/Auction.sol
nuevoalex: I wanted to really call out where this was happening and how the new auction initialization works.

source/contracts/reporting/Auction.sol
nuevoalex: Yup, makes sense and will address

source/contracts/reporting/Auction.sol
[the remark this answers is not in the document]
nuevoalex: The dormant auctions are skipped

source/contracts/reporting/Auction.sol
nuevoalex: It wasn't too complicated but that would be nicer of course.

source/contracts/reporting/Auction.sol
epheph: I think the name makes sense since it isn't `removeAuthorizedPriceFeed` it's remove from a larger set?

nuevoalex: The name is is remove _from_ authorized price feeders. Could rename it to just `removeAuthorizedPriceFeeder`

source/contracts/reporting/Auction.sol
epheph: Then we will need some sort of a threshold for when we consider the auction ended, and what is considered "dust" . Very commonly, systems are left with tiny amounts that no longer make sense to purchase. If i were using this system and wanted to buy a lot, I would buy up to what is considered dust + 1, forcing time to go lower before someone decides that it is worth it to buy that last batch

source/contracts/LegacyReputationToken.sol
nuevoalex: Oh actually this contract is not the _real_ Legacy REP token contract.

This is the one used in testing and development which includes a faucet function so it already is accepted to not be 1:1 with the real thing. _That_ contract is in the legacy_reputation folder as `OldLegacyReputationToken.sol`

source/contracts/reporting/Auction.sol
nuevoalex: Will just make it a function

source/contracts/LegacyReputationToken.sol
MicahZoltu: The test contract for LegacyReputationToken should not contain new REP interface.  I recognize the need for the faucet to make it work in test networks, but we don't want to accidentally code against the new token interface and then have it fail in production because the real thing doesn't have those methods.

source/contracts/reporting/Auction.sol
MicahZoltu: I didn't actually know what they were for, so I didn't have a better name (and why I thought they should get a better name).

Perhaps `DAY_3_NO_AUCTION`, `DAY_4_NO_AUCTION`?  Alternatively, perhaps we should just have a `day` integer and have two constants that we compare against for `UNRECORDED_AUCTION_DAY` and `RECORDED_AUCTION_DAY` rather than having an enum with a bunch of duplicate-like entries?

source/contracts/reporting/Auction.sol
MicahZoltu: One option would be for FF to just fund the initial auction.  Probably too expensive though.  Another option would be to have a separate mechanism for turning initial REP into ETH rather than trying to bundle it into the existing auction system.

nuevoalex: That seems more complicated to me to create a whole other system. There's really not much additional code to do the bootstrap

source/contracts/reporting/Auction.sol
MicahZoltu: In a uniform price multiunit auction the auction ends when the capital in escrow is greater than `lot * price`.  Since a price of `0` means the lot sells for `0`, this means the auction will _always_ fully clears the whole lot (exception for if _no one_ bids).

source/contracts/reporting/Auction.sol
MicahZoltu: Ah, I didn't catch that.  I reasonably strongly thing this should be renamed as @nuevoalex suggested since right now the function name is referencing an internal (private) datastructure as its name, rather than the publicly visible concept of price feeders.

source/contracts/LegacyReputationToken.sol
nuevoalex: I'll do this because it's technically true, but in protest because the failure case would never happen for this code

source/contracts/reporting/Auction.sol
nuevoalex: I like having a specific term for the days where the auction is dormant.

source/contracts/reporting/Auction.sol
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    function isActive() public view returns (bool) {
+        RoundType _roundType = getRoundType();
+        return _roundType == RoundType.UNRECORDED || _roundType == RoundType.RECORDED;
```
epheph: What if it's in bootstrapMode?

source/contracts/reporting/AuctionToken.sol
```diff
@@ -0,0 +1,85 @@
...
+        require(auction.getAuctionIndexForCurrentTime() > auctionIndex);
+        uint256 _ownerBalance = balances[msg.sender];
+        uint256 _tokenBalance = redemptionToken.balanceOf(this);
+        uint256 _redemptionAmount = _ownerBalance.mul(_tokenBalance).div(totalSupply());
+        if (redemptionToken == ERC20(cash)) {
+            cash.withdrawEtherTo(msg.sender, _redemptionAmount);
+        } else {
+            redemptionToken.transfer(msg.sender, _redemptionAmount);
+        }
+        burn(msg.sender, _ownerBalance);
```
epheph: I know withdrawEtherTo is not reentrant, but it feels safer to burn first since we are performing a send?

source/contracts/reporting/Auction.sol
nuevoalex: Hm so for the case where people don't participate and it ends up occuring on the RECORDED round perhaps we could ignore it explicitly.

Thing is we wont be using the results anyway for some time so I'm not sure its worth the code.

source/contracts/reporting/Auction.sol
```diff
@@ -0,0 +1,289 @@
...
+
+    function getAuctionStartTime() public view returns (uint256) {
+        uint256 _auctionIndex = getAuctionIndexForCurrentTime();
+        uint256 _auctionDay = _auctionIndex % 7;
+        uint256 _weekStart = initializationTime.add(_auctionIndex).sub(_auctionDay);
+        uint256 _addedTime = _auctionDay > uint256(RoundType.UNRECORDED) ? uint256(RoundType.RECORDED) : uint256(RoundType.UNRECORDED);
+        return _weekStart.add(_addedTime.mul(Reporting.getAuctionDuration()));
+    }
+
+    function getAuctionEndTime() public view returns (uint256) {
```
adrake33: Might want to consider making the code for this and `getAuctionStartTime` more DRY since they're almost the same. (Then again, it's probably a bit more readable this way.)

source/contracts/reporting/Reporting.sol
```diff
@@ -38,4 +43,8 @@ library Reporting {
     function getMinimumReportingFeeDivisor() internal pure returns (uint256) { return MINIMUM_REPORTING_FEE_DIVISOR; }
     function getDefaultReportingFeeDivisor() internal pure returns (uint256) { return DEFAULT_REPORTING_FEE_DIVISOR; }
     function getInitialREPSupply() internal pure returns (uint256) { return INITIAL_REP_SUPPLY; }
+    function getAuctionInitialRepprice() internal pure returns (uint256) { return AUCTION_INITIAL_REP_PRICE; }
```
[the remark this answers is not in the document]
nuevoalex: Yup!

PR #751, opened 2018-10-12
bond and fee changes. also metadata removal

nuevoalex:


source/contracts/reporting/Reporting.sol
```diff
@@ -19,8 +19,8 @@ library Reporting {
     uint256 private constant TARGET_INVALID_MARKETS_DIVISOR = 100; // 1% of markets are expected to be invalid
     uint256 private constant TARGET_INCORRECT_DESIGNATED_REPORT_MARKETS_DIVISOR = 100; // 1% of markets are expected to have an incorrect designate report
     uint256 private constant TARGET_DESIGNATED_REPORT_NO_SHOWS_DIVISOR = 100; // 1% of markets are expected to have an incorrect designate report
-    uint256 private constant TARGET_REP_MARKET_CAP_MULTIPLIER = 15; // We multiply and divide by constants since we want to multiply by a fractional amount (7.5)
-    uint256 private constant TARGET_REP_MARKET_CAP_DIVISOR = 2;
+    uint256 private constant TARGET_REP_MARKET_CAP_MULTIPLIER = 3; // We multiply and divide by constants since we may want to multiply by a fractional amount
+    uint256 private constant TARGET_REP_MARKET_CAP_DIVISOR = 1;
```
adrake33: How were these numbers arrived at?

nuevoalex: Game Theory discussion / will be explained in updates to the whitepaper. This is the minimum target OI with the new Use or Lose system implemented. We want the lowest target OI possible to minimize fees

source/contracts/reporting/Market.sol
```diff
@@ -30,7 +30,7 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
     using SafeMathInt256 for int256;
 
     // Constants
-    uint256 private constant MAX_FEE_PER_ETH_IN_ATTOETH = 1 ether / 2;
+    uint256 private constant MAX_FEE_PER_ETH_IN_ATTOETH = 15 * 10**16; // 15%
```
MicahZoltu: Did you have an argument for why this is a reasonable number?  This feels way too low to put a hard-cap in for.  Credit cards frequently charge more than this (significantly in some cases), as an example of a situation where people are willing to pay greater than 15% in fees for something.

source/contracts/reporting/Reporting.sol
MicahZoltu: I don't remember actually coming to consensus on 3?  I know it was thrown around as a straw man number, but I do think we should make sure to think through what the right number is before v2.

source/contracts/reporting/Market.sol
nuevoalex: Yea. When considering the maximum reporting fee this is near the highest we can allow the market creator fee before the combined fee makes participation irrational (combined maximum is now 48% where 50% mean someone would always be losing even if they win).

source/contracts/reporting/Reporting.sol
nuevoalex: Of course. Will re-review after the whitepaper is updated

PR #753, opened 2018-10-22
Rename FeeWindow to DisputeWindow

nuevoalex:
Just a rename of FeeWindow -> DisputeWindow


MicahZoltu:
Same argument as before on why DisputeWindow is an inaccurate name that evokes the incorrect mental model in readers.

nuevoalex:
These are used very specifically to space out the time between dispute rounds and to give a window to initially dispute.

Seems like calling them dispute windows is very apt to me.

MicahZoltu:
Hmm, looking at the contents of this contract, it doesn't appear to contain what I thought it contained.  It appears to just be a place for tracking number of invalid/no-show markets?

nuevoalex:
That's all that really gets recorded on it, correct. If anything that is the coincidental data though since we have a non-arbitrary reason for spacing the dispute windows as we do. We could adjust bonds at a different cadence if we wanted but this is a nice way to do it.

tests/reporting/test_bond_redemption.py
```diff
@@ -69,12 +69,12 @@ def test_failed_crowdsourcer(finalize, localFixture, universe, market, cash, rep
 
     if finalize:
         # Fast forward time until the fee window is over and we can redeem to recieve the REP back
```
adrake33: Looks like there are still some references to "fee window" as 2 words. Recommend doing a find/replace on those as well.

PR #754, opened 2018-10-24
change ROI of disputing to 40% and burn the rest of the losing REP

nuevoalex:


tests/solidity_test_helpers/MockUniverse.sol
```diff
@@ -5,7 +5,7 @@ import 'reporting/IUniverse.sol';
 import 'reporting/IDisputeWindow.sol';
 import 'reporting/Reporting.sol';
 import 'libraries/ITyped.sol';
-import 'reporting/IReputationToken.sol';
+import 'reporting/IV2ReputationToken.sol';
```
MicahZoltu: Curious why you are calling this `IV2ReputationToken` rather than renaming previous token to `IV1ReputationToken`?

epheph: Seems confusing to rename old things and having to remember context of version of code you're reading

PR #755, opened 2018-10-25
Explicit Invalid outcome

nuevoalex:
Looks large but its mostly adjusting for the new function signature and additional outcome

### DIFF
--- a/source/contracts/reporting/Market.sol
+++ b/source/contracts/reporting/Market.sol
@@ -53,7 +53,6 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
     uint256 private noShowBond;
     bool private disputePacingOn;
     address private noShowBondOwner;
-    uint256 public invalidOutcomeIndex;
 
     // Collections
     IReportingParticipant[] public participants;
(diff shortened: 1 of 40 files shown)


source/contracts/reporting/Market.sol
```diff
@@ -61,6 +62,8 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
 
     function initialize(IUniverse _universe, uint256 _endTime, uint256 _feePerEthInAttoEth, ICash _cash, address _designatedReporterAddress, address _creator, uint256 _numOutcomes, uint256 _numTicks) public payable beforeInitialized returns (bool _success) {
         endInitialization();
+        invalidOutcomeIndex = _numOutcomes;
+        _numOutcomes += 1; // The INVALID outcome is always last
```
MicahZoltu: Out of curiosity, why always last instead of first?  It feels like things would be simpler if it was first because we could just have a constant for `INVALID_OUTCOME = 0` rather than having to do `_numOutcomes` (less clear).  Also, `0` _feels_ like a nice number for INVALID.

MicahZoltu: Would also save on storage, since it appears you are storing `invalidOutcomeIndex` explicitly, rather than just using the already stored `numOutcomes` (which incidentally is a waste of gas).

source/contracts/reporting/Market.sol
```diff
@@ -477,22 +480,18 @@ contract Market is DelegationTarget, ITyped, Initializable, Ownable, IMarket {
...
+    function derivePayoutDistributionHash(uint256[] _payoutNumerators) public view returns (bytes32) {
         uint256 _sum = 0;
         uint256 _previousValue = _payoutNumerators[0];
+        require(_payoutNumerators[0] == 0 || _payoutNumerators[0] == numTicks);
         require(_payoutNumerators.length == numOutcomes);
         for (uint256 i = 0; i < _payoutNumerators.length; i++) {
             uint256 _value = _payoutNumerators[i];
             _sum = _sum.add(_value);
-            require(!_invalid || _value == _previousValue);
             _previousValue = _value;
```
epheph: _previousValue unused

nuevoalex: nice catch, will fix in monorepo

PR #756, opened 2018-11-07
switch to using the optimized proxy contract instead of our delegator

nuevoalex:
Using the minimal proxy contract instead of our Delegator: https://github.com/optionality/clone-factory/issues/10

Gas savings of roughly 25-40% on everything!


source/contracts/Controlled.sol
```diff
@@ -20,7 +20,7 @@ contract Controlled is IControlled {
     }
 
     modifier onlyControllerCaller {
-        require(IController(msg.sender) == controller);
+        require(IController(msg.sender) == controller || controller == address(0));
```
MicahZoltu: Can you go into detail on how this is a good thing?  It means that if the controller address is ever set to 0, any `onlyControllerCaller` calls will be callable by anyone (essentially public with no modifier).  At the least, the modifier name should be renamed to be more clear, though I would like to see some strong arguments for why this should be this way.

source/contracts/libraries/CloneFactory.sol
```diff
@@ -0,0 +1,59 @@
...
+MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
+//solhint-disable max-line-length
+//solhint-disable no-inline-assembly
+
+contract CloneFactory {
```
MicahZoltu: Is this a straight copy/paste from https://github.com/optionality/clone-factory/blob/master/contracts/CloneFactory.sol?  Probably worth referencing the EIP in a comment in this file.

source/contracts/libraries/CloneFactory.sol
```diff
@@ -0,0 +1,59 @@
...
+in all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
```
MicahZoltu: I'm reasonably strongly against putting licenses into source files.  We have https://github.com/AugurProject/augur-core/blob/master/CRIBBED_CODE.txt for giving attribution.

source/contracts/libraries/CloneFactory.sol
```diff
@@ -0,0 +1,59 @@
...
+//solhint-disable max-line-length
+//solhint-disable no-inline-assembly
```
MicahZoltu: I don't think these are necessary given our repo wide solhint settings.

source/contracts/libraries/CloneFactory.sol
```diff
@@ -0,0 +1,59 @@
...
+    assembly {
+      let clone := mload(0x40)
+      mstore(clone, 0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000000000000000000000)
+      mstore(add(clone, 0x14), targetBytes)
+      mstore(add(clone, 0x28), 0x5af43d82803e903d91602b57fd5bf30000000000000000000000000000000000)
+      result := create(0, clone, 0x37)
+    }
+  }
+
+  function isClone(address target, address query) internal view returns (bool result) {
```
MicahZoltu: A ctrl-f on this PR suggests that we do not use `isClone` anywhere.  Recommend deleting it as dead code (reduces audit surface area).

source/contracts/reporting/IMarket.sol
```diff
@@ -10,7 +10,8 @@ import 'reporting/IDisputeWindow.sol';
 import 'trading/IShareToken.sol';
 import 'reporting/IReportingParticipant.sol';
 import 'reporting/IReputationToken.sol';
-//import 'reporting/IInitialReporter.sol';
+import 'IController.sol';
+import 'IControlled.sol';
```
MicahZoltu: This file added these two imports, but there are no other changes to the file, suggesting that these imports are not necessary?

source/contracts/Controlled.sol
nuevoalex: Same could be said for anyone calling the initialize function on any of our factory generated contracts. Its yet another thing we need to do as part of initialization (setting the controller). We also can't overload the initialization function with this as some contracts (Market in particular) cannot have any additional variables added to their initialize method.

That said we probably don't really need the `Controlled` class as much anyway as lookups should only really be relevant at construction time and only during the initial deploy.

There will be a separate issue for evaluating / removing the `Controlled` class entirely.

source/contracts/factories/AuctionFactory.sol
```diff
@@ -1,17 +1,18 @@
...
 
 
-contract AuctionFactory {
+contract AuctionFactory is CloneFactory {
     function createAuction(IController _controller, IUniverse _universe) public returns (IAuction) {
-        Delegator _delegator = new Delegator(_controller, "Auction");
-        IAuction _Auction = IAuction(_delegator);
-        _Auction.initialize(_universe);
-        return _Auction;
+        IAuction _auction = IAuction(createClone(_controller.lookup("Auction")));
```
epheph: Gas saving here! Constructing the delegator with address instead of forcing lookup on each invocation. 👍

source/contracts/factories/AuctionFactory.sol
```diff
@@ -1,17 +1,15 @@
...
+import 'reporting/IReputationToken.sol';
 
 
-contract AuctionFactory {
-    function createAuction(IController _controller, IUniverse _universe) public returns (IAuction) {
-        Delegator _delegator = new Delegator(_controller, "Auction");
-        IAuction _Auction = IAuction(_delegator);
-        _Auction.initialize(_universe);
-        return _Auction;
+contract AuctionFactory is CloneFactory {
```
epheph: Does this need to be CloneFactory?

source/contracts/factories/ReputationTokenFactory.sol
```diff
@@ -1,17 +1,14 @@
...
+import 'reporting/ReputationToken.sol';
 
 
-contract ReputationTokenFactory {
-    function createReputationToken(IController _controller, IUniverse _universe) public returns (IReputationToken) {
-        Delegator _delegator = new Delegator(_controller, "ReputationToken");
-        IReputationToken _reputationToken = IReputationToken(_delegator);
-        _reputationToken.initialize(_universe);
-        return _reputationToken;
+contract ReputationTokenFactory is CloneFactory {
```
epheph: Same here

source/contracts/factories/UniverseFactory.sol
```diff
@@ -1,16 +1,13 @@
...
 
-import 'libraries/Delegator.sol';
+import 'libraries/CloneFactory.sol';
 import 'IController.sol';
-import 'reporting/IUniverse.sol';
+import 'reporting/Universe.sol';
 
 
-contract UniverseFactory {
+contract UniverseFactory is CloneFactory {
```
epheph: Here

source/contracts/factories/AuctionFactory.sol
nuevoalex: Oh right it definitely doesn't now.

epheph:
So this look great, and it will save a ton of gas, both in a more optimal delegator AND in no more controller lookups.

Like @MicahZoltu , I am just a tiny bit spooked about the setController defaulting to 0, but see why this is required now with the new clone method. It helps that that modifier is ONLY called when setting controller, and a controller left at 0 should be very noticeable (since lookups stop working) it shouldn't be able to pass tests easily.

source/libraries/ContractDeployer.ts
```diff
@@ -172,29 +172,23 @@ Deploying to: ${networkConfiguration.networkName}
...
     private async upload(contract: Contract): Promise<void> {
-        const contractsToDelegate: {[key:string]: boolean} = {"Orders": true, "TradingEscapeHatch": true, "Cash": true};
         const contractName = contract.contractName
         if (contractName === 'Controller') return;
         if (contractName === 'Delegator') return;
         if (contractName === 'TimeControlled') return;
-        if (contractName === 'TestNetReputationToken') return;
+        if (contractName === 'TestNetReputationTokenFactory') return;
         if (contractName === 'Augur') return;
+        if (contractName === 'Universe') return;
```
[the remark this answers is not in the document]
MicahZoltu: This warning should be turned off.  It is terrible advice from a horrible era in software engineering (before guard clauses).

PR #760, opened 2018-12-07
Update contributing info to redirect to monorepo

adrake33:
If this looks ok, I'd like to add a similar CONTRIBUTING.md file to our other repos.  I believe having a CONTRIBUTING.md file should automatically add a banner like this to the top of the page:

screen shot 2018-12-07 at 3 01 47 pm

(For some reason, though, I'm not seeing this banner on the augur-core Issues page. I thought it might be because I had already dismissed the banner, but I still don't see it even if I view the page using an incognito window.)


adrake33:
Actually, I don't think we need to redirect people to the monorepo after all. Since ZenHub can handle issues from multiple repos and we don't need to use the GH Projects tab, it makes sense to continue to have people files issues in the v1 repos as they were previously doing. I'm not sure if anyone actually reads this file if the banner isn't showing, but we might as well update the file with the other minor changes I made.

